PR #5, opened 2017-08-16
Change Project to App and sort out the fields, add basic tests to cov…

jmoz:
```
pip install -r requirements.txt
./manage migrate
```


control_panel_api/migrations/0002_auto_20170816_1525.py
```diff
@@ -0,0 +1,32 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='App',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created', django_extensions.db.fields.CreationDateTimeField(auto_now_add=True)),
+                ('name', models.CharField(max_length=100)),
+                ('slug', django_extensions.db.fields.AutoSlugField(blank=True, editable=False, populate_from='name')),
+                ('repo_url', models.CharField(blank=True, default='', max_length=255)),
```
xoen: It may be more than plenty but a part of me wonder if limit this URL to `255` chars may be a problem for some funny named repos.

control_panel_api/settings.py
```diff
@@ -110,7 +111,7 @@
 
 REST_FRAMEWORK = {
     'DEFAULT_PERMISSION_CLASSES': [
-        'rest_framework.permissions.IsAdminUser',
+        'rest_framework.permissions.IsAuthenticated',
```
xoen: I'd probably stick with `IsAdminUser` as `IsAuthenticated` is quite permissive (I know this will change anyway).

control_panel_api/tests/test_views.py
```diff
@@ -4,19 +4,25 @@
...
 
 
-class UserViewTest(APITestCase):
+class AuthenticatedClientMixin(object):
     def setUp(self):
-        mommy.make('control_panel_api.User')
-        self.user = mommy.make('control_panel_api.User', is_staff=True)
-        self.client.force_authenticate(self.user)
+        super(AuthenticatedClientMixin, self).setUp()
+        self.client.force_login(mommy.make('control_panel_api.User'))
```
xoen: If we go back to the `IsAdminUser` permission class, this will obviously have to create a user with the `is_staff` flag set.

control_panel_api/tests/test_views.py
```diff
@@ -4,19 +4,25 @@
...
+
+
+class UserViewTest(AuthenticatedClientMixin, APITestCase):
+    def setUp(self):
+        super(UserViewTest, self).setUp()
+        self.fixture = mommy.make('control_panel_api.User')
+        self.client.force_authenticate(self.fixture)
```
xoen: Why do we use `force_authenticate` again here? I thought the point of the `AuthenticatedClientMixin` was to avoid to log the user in explicitly.

control_panel_api/tests/test_views.py
```diff
@@ -4,19 +4,25 @@
 from rest_framework.test import APITestCase
 
 
-class UserViewTest(APITestCase):
+class AuthenticatedClientMixin(object):
     def setUp(self):
-        mommy.make('control_panel_api.User')
-        self.user = mommy.make('control_panel_api.User', is_staff=True)
-        self.client.force_authenticate(self.user)
+        super(AuthenticatedClientMixin, self).setUp()
```
xoen: Why do we need to call `super().setUp()` here?

control_panel_api/serializers.py
```diff
@@ -16,7 +16,7 @@ class Meta:
         fields = ('id', 'url', 'name')
 
 
-class ProjectSerializer(serializers.HyperlinkedModelSerializer):
+class AppSerializer(serializers.HyperlinkedModelSerializer):
     class Meta:
-        model = Project
-        fields = ('id', 'url', 'name', 'slug', 'repository')
+        model = App
+        fields = ('id', 'name', 'slug', 'repo_url')
```
xoen: We should keep the `url` as well here as in the other views. (The `url` in this case is the "API hyperlink" which is kind of nice).

control_panel_api/tests/test_views.py
```diff
@@ -4,19 +4,25 @@
...
+        self.client.force_login(mommy.make('control_panel_api.User'))
+
+
+class UserViewTest(AuthenticatedClientMixin, APITestCase):
+    def setUp(self):
+        super(UserViewTest, self).setUp()
```
xoen: Is this `super().setUp()` necessary to call the `setUp()` in `AuthenticatedClientMixin`?

control_panel_api/tests/test_views.py
jmoz: So that when mixed in, it will still execute the parent TestCase setUp

control_panel_api/tests/test_views.py
jmoz: Yep so this executes the mixin setUp, which in turn executes the APITestCase setUp

control_panel_api/settings.py
jmoz: Sure.

control_panel_api/tests/test_views.py
jmoz: Well spotted - I think I was tired yday.

control_panel_api/serializers.py
jmoz: Yes good shout.

xoen:
Looks good to me, good stuff 👍

PR #6, opened 2017-08-16
Fix login button in Swagger UI

xoen:
The Swagger UI relies on the `LOGIN_URL` and `LOGOUT_URL` in the
settings to work properly.


jmoz:
LGTM

PR #7, opened 2017-08-18
Add command to load users from kubernetes

jmoz:


control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
...
+            try:
+                users.append(User.objects.create(
+                    username=username,
+                    email=email
+                ))
+                logger.debug("Imported {} {}".format(username, email))
+            except IntegrityError:
+                pass
+
+        self.stdout.write("Imported {} users".format(len(users)))
```
xoen: Minor point, if we keep the `users` in an array just to display the number of imported users at the end, it may be better to just have a counter (e.g.`users_count` or `n_added_user`).

control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
+import base64
+import logging
+
+from django.core.management.base import BaseCommand
+from django.db.utils import IntegrityError
+from kubernetes import config, client
```
xoen: We should probably add `kubernetes` in the `requirements.txt`.

control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
...
+
+        user_secrets = [item.data for item in results.items if item.metadata.name == Command.USER_SECRETS_NAME]
+
+        users = []
+        for item in user_secrets:
+            username = base64.b64decode(item['username'])
+            email = base64.b64decode(item['email'])
+
+            try:
+                users.append(User.objects.create(
```
xoen: What would happen if create raise an exception to `users`? I imagine that would return `None`, would that be added to `users`?

control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
...
+    def handle(self, *args, **options):
+        config.load_kube_config()
+        v1 = client.CoreV1Api()
+        results = v1.list_secret_for_all_namespaces()
+
+        user_secrets = [item.data for item in results.items if item.metadata.name == Command.USER_SECRETS_NAME]
+
+        users = []
+        for item in user_secrets:
+            username = base64.b64decode(item['username'])
```
xoen: We should also have the full name in `user-secrets` (see [helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/init-user/templates/user-secrets.yml#L12)), I feel like we should also import this information :)

xoen: Right, just checked the documentation of [`django.contrib.auth.models.User`](https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#user-model) and it seems it has separate fields for first name and last name (which makes sense), while we have the full name in the secret, not sure how to handle this (I mean, we may put the string before the first space in the `first_name` and the rest in `last_name` as a good guess)

control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Import Kubernetes user-secrets data into the User.  Uses KUBECONFIG envvar or defaults to ~/.kube/config"
+
+    USER_SECRETS_NAME = 'user-secrets'
+
+    def add_arguments(self, parser):
```
xoen: Do we need to override the method in the parent class if we don't have arguments? (apologies for the silly question, I'm not familiar with any of this)

xoen:
This look good to me :) - see my comments.

control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
...
+        pass
+
+    def handle(self, *args, **options):
+        config.load_kube_config()
+        v1 = client.CoreV1Api()
+        results = v1.list_secret_for_all_namespaces()
+
+        user_secrets = [item.data for item in results.items if item.metadata.name == Command.USER_SECRETS_NAME]
```
xoen: Couldn't find the documentation for this module but I was wondering if there is a way to get the list of secrets matching this directly from k8s.

Far from a k8s expert but I guess `list_secret_for_all_namespaces()` will call [this k8s endpoint](https://kubernetes.io/docs/api-reference/v1.6/#list-all-namespaces-211).

Nevermind, [doesn't look like fieldSelector is supported very well](https://github.com/kubernetes/kubernetes/issues/15128) a part from pods at the moment.

control_panel_api/management/commands/load_kube_users.py
xoen: Actually, if the `create` raises, it would not even return, so `append()` should be executed at all :D.

control_panel_api/management/commands/load_kube_users.py
```diff
@@ -0,0 +1,42 @@
...
+from django.core.management.base import BaseCommand
+from django.db.utils import IntegrityError
+from kubernetes import config, client
+
+from control_panel_api.models import User
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
xoen: So this would be available via `python manage.py load_kube_users`, nice :)!

I wonder if we should change `load_` with `import_` to be more explicit, maybe call it `import_k8s_users`? Just an idea.

control_panel_api/management/commands/load_kube_users.py
jmoz: Good point

control_panel_api/management/commands/load_kube_users.py
jmoz: Thanks, forgot

control_panel_api/management/commands/load_kube_users.py
jmoz: Yep exception is thrown and caught, all ok

control_panel_api/management/commands/load_kube_users.py
jmoz: No it's just a stub method that gets called. I just left it here so it's obvious if we need to add arguments. Actually I'll remove it, now you know about it.

control_panel_api/management/commands/load_kube_users.py
jmoz: Yeh I did look for something that does like this:

kubectl get secret user-secrets -n user-andyhd -o json

it gets secret by the name 'user-secrets'

But I couldn't find in library.

control_panel_api/management/commands/load_kube_users.py
jmoz: Yes import is fine, is k8s the standard naming? (I have no idea)

control_panel_api/management/commands/load_kube_users.py
jmoz: I just checked response and all names are "first last" so can split them. Adding a default to '' if error.

control_panel_api/management/commands/load_kube_users.py
xoen: Yeah, don't worry, I was "thinking aloud" and filtering in memory is more than OK.

control_panel_api/management/commands/load_kube_users.py
xoen: 👍

control_panel_api/management/commands/load_kube_users.py
xoen: Sounds good to me 👍

control_panel_api/management/commands/load_kube_users.py
xoen: Don't know if there is a standard naming convention or anything like that but kubernetes is usually shortened to `k8s`.

xoen:
Good stuff @jmoz  👍

control_panel_api/management/commands/load_kube_users.py
kerin: I think we'd be better off providing a custom `User` object with a single `name` field, rather than splitting existing values into two strings. As this User model will become the canonical user representation I'd rather we [handle names sensibly](http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) from the start.

xoen:
Another thing occurred me this weekend. I don't know django's commands but I imagine this will run locally but the DB will be somewhere else, so when running this locally it will have to connect to the remote DB and/or run on the same machine on which the app is deployed. Or something like that.

We'll see how it goes once we have a DB, etc...

PR #9, opened 2017-08-22
Prepare docker image for deployment using CircleCI

andyhd:
## What

* Added Dockerfile for API webapp
* Updated CircleCI config to build docker image and run tests using it, then push to quay.io, so that it can be deployed to the cluster

## How to review

1. See [CircleCI build logs](https://circleci.com/gh/ministryofjustice/analytics-platform-control-panel/75)
2. See [quay.io images](https://quay.io/repository/mojanalytics/control-panel?tab=tags)

See [#342 API and frontend deployment](https://trello.com/c/jxufKhNR)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,8 @@ ADD requirements.txt requirements.txt
 
 RUN pip3 install -r requirements.txt
 
+ADD run_tests run_tests
+ADD wait_for_db wait_for_db
 ADD manage.py manage.py
 ADD control_panel_api control_panel_api
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,4 +23,4 @@ ADD control_panel_api control_panel_api
 
 EXPOSE 8000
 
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM alpine:3.6
+
+MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
+
+RUN apk update && \
+    apk add \
+        --virtual build-deps \
+        gcc \
+        musl-dev \
+        postgresql-dev \
```
xoen: Is this needed to install some of the requirements? Postgres should run in a separate container right? (using docker compose to run the tests).

xoen:
As far as I can tell looks good

Dockerfile
andyhd: Yes, these are dependencies of `psycopg2` the postgres client library

xoen: 👍

PR #10, opened 2017-08-24
Added skeleton of permissions and filtering for users/apps

xoen:
This is currently very simple and assumes superusers (users with
`is_superuser == True`) can do/see everything, others can't do/see
anything.

I also updated the views tests to make requests both as superuser and
as normal user.

This cover one of the use cases:

    As a superuser, I want to be able to create a user

**NOTE**: This is just a start, apps permissions/filters will need to take
in considerations teams/roles/permissions.

### DIFF
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -1,0 +1,36 @@
+"""
+Custom filters
+
+See: http://www.django-rest-framework.org/api-guide/filtering/#custom-generic-filtering
+"""
+from rest_framework.filters import BaseFilterBackend
+
+from control_panel_api.permissions import is_superuser
+
+
+class AppFilter(BaseFilterBackend):
+    """
+    Filter to get visible apps.
+
+    Currently superusers see everything, others see nothing.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        if is_superuser(request.user):
+            return queryset
+        else:
+            return queryset.none()
+
+
+class UserFilter(BaseFilterBackend):
+    """
+    Filter to get visible users.
+
+    Currently superusers see everything, others see nothing.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        if is_superuser(request.user):
+            return queryset
+        else:
+            return queryset.none()
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -8,29 +8,33 @@ from rest_framework.filters import BaseFilterBackend
 from control_panel_api.permissions import is_superuser
 
 
-class AppFilter(BaseFilterBackend):
+class SuperusersOnlyFilter(BaseFilterBackend):
+    """
+    Superusers can see everything. Other users can't see anything
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        if is_superuser(request.user):
+            return queryset
+        else:
+            return queryset.none()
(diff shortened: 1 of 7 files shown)


control_panel_api/permissions.py
```diff
@@ -0,0 +1,28 @@
...
+class IsSuperuser(BasePermission):
+    """
+    Only superusers are authorised
+    """
+
+    def has_permission(self, request, view):
+        return is_superuser(request.user)
+
+
+class AppPermissions(IsSuperuser):
```
jmoz: Rather than name the permission after the view, it would be better named to show it's intention e.g. IsAdmin or IsSUperuser - in the viewset then it would be more obvious what the class did, rather than a catch-all class where you have to access the class and look up it's hierarchy.  The rest framework has a permissions file like this.

Also I would not use empty classes for the sake of it, just use the IsSuperuser at the moment.

And at runtime can the permissions be composed of multiple classes, rather than just use one?

control_panel_api/filters.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        if is_superuser(request.user):
+            return queryset
+        else:
+            return queryset.none()
+
+
+class AppFilter(SuperusersOnlyFilter):
```
jmoz: Same applies here as what I said for permissions.

control_panel_api/views.py
```diff
@@ -1,17 +1,27 @@
...
-    UserSerializer
+    UserSerializer,
 )
 
 
 class UserViewSet(viewsets.ModelViewSet):
     queryset = User.objects.all()
     serializer_class = UserSerializer
+    filter_backends = (UserFilter,)
+    permission_classes = (UserPermissions, )
```
jmoz: Here the permissions would be better named with their intention.

control_panel_api/tests/test_views.py
```diff
@@ -1,26 +1,48 @@
...
 
 
 class UserViewTest(AuthenticatedClientMixin, APITestCase):
+
     def setUp(self):
         super(UserViewTest, self).setUp()
-        self.fixture = mommy.make('control_panel_api.User')
+        self.fixture = self.normal_user
+
+    def test_list_as_superuser(self):
```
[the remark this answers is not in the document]
xoen: Good point, I'll add some tests for permissions and filtering (even if at this stage they're very basic and they'll change)

control_panel_api/permissions.py
xoen: I was playing with multiple permission classes and I found that if you have more than one permission class, they'll all have to return `True`.

This limits the usefulness of this mechanism, e.g. let's say we have the following viewset using these two permission classes:

```python
class SomeViewSet(viewsets.ModelViewSet):
    # ...
    permission_classes = (IsSuperuser, IsAppMaintainer,)
```

It reads better but it's not behaving as expected: both classes will need to return `True`, so you'll have to be an app maintainer and a super user at the same time, which is probably not what we'd want.

Also some of the logic may be more complicated than this and also return different things depending on the methods and/or depending on the object involved.

xoen:
@jmoz made the suggested changes to the tests. It's pretty much what I've shown you offline but feel free to have another look ;)

PR #11, opened 2017-08-25
Added Team, Role and TeamMembership models

xoen:
The `TeamMembership` model keep the relationship between teams, users and
roles.
A user belongs to a team with a given role (e.g. "maintainer" or "member").

A user can be in a team only once and only with a single role:
  * add a user to a team more than once doesn't make any sense
  * add a user to a team with more than one role is confusing and doesn't
    make any sense (e.g. has the user the "member" role or the "maintainer"
    role?)

I also added these membership utility methods:
  * `User#teams` returns the teams (queryset) the user belongs to
  * `Team#users` returns the users (queryset) in the team
  * `Team#users_with_role` returns the users (queryset) in the team with
    the given `role_code`


andyhd:
Why Role "code" and not "name"?

xoen:
Thanks for the review @andyhd , I'll add a `name` and use [`django.db.TimeStampedModel`](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/models.py#L14) as parent class of our models (and see if breaks).

xoen:
@andyhd FYI: I added some tests, the `name` field to roles and changed models to inherit from `django_extensions.db.models.TimeStampedModel`, feel free to have another look.

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,79 @@
...
+
+from control_panel_api.models import (
+    Role,
+    Team,
+    TeamMembership,
+    User,
+)
+
+
+class FixtureTestCase(TestCase):
```
andyhd: This might be better named something like `TeamMembershipTestCase`

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @classmethod
+    def setUpTestData(cls):
```
andyhd: Should this be called `setUpClass`? https://docs.python.org/3.6/library/unittest.html#setupclass-and-teardownclass

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,79 @@
...
+            team=cls.team_justice,
+            user=cls.user_bob,
+            role=cls.role_member,
+        )
+        # Add other user to other team as member
+        TeamMembership.objects.create(
+            team=cls.team_other,
+            user=cls.user_other,
+            role=cls.role_member,
+        )
```
andyhd: Should these fixtures be set up with `model_mommy`?

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class UserTest(FixtureTestCase):
+
+    def test_teams(self):
+        alices_teams = self.user_alice.teams()
+
+        assert self.team_justice in alices_teams
+        assert self.team_other not in alices_teams
```
andyhd: When using `unittest`, these assertions should really be `self.assertIn` and `self.assertNotIn` (https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn)

control_panel_api/tests/test_models.py
andyhd: These tests should all be methods on a single class, rather than subclasses of `FixtureTestCase`.

control_panel_api/tests/test_models.py
andyhd: These `TeamMembership` objects should not be in a global setup - these are setting up the context for individual tests and should be created in the tests methods that test them. Eg:
```python
def test_alice_is_in_team_justice(self):
    TeamMembership.objects.create(user=alice, team=team_justice, role=maintainer)
    self.assertIn(team_justice, alice.teams())
    self.assertNotIn(team_other, alice.teams())
```

The first set of fixtures (alice, bob, other, and the teams) are like the "Given some users and teams exist" context.
The second set are more like "When Alice is maintainer of Team Justice".

xoen:
Rebased off `api` branch and removed unnecessary `("team", "user", "role")` unique constraint, as we already have a unique constraint on `("team", "user")`.

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,98 @@
...
+        assert self.user_bob not in justice_maintainers
+        assert self.user_bob in justice_members
+        assert self.user_other not in justice_maintainers
+        assert self.user_other not in justice_members
+
+
+class Membership(FixtureTestCase):
+
+    def test_user_can_be_added_to_team_only_once(self):
+        raised_integrity_error = False
```
andyhd: This is reimplementing what `unittest` already provides in `self.assertRaises`: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

control_panel_api/tests/test_models.py
xoen: Bear in mid that `TestCase` here is an instance of `django.test.TestCase` which provides the`setUpTestData` method for this specific use case: https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.TestCase.setUpTestData

control_panel_api/tests/test_models.py
xoen: To be honest I don't see what's the benefit of using it in this case when using the django ORM methods are called in the same way and they're standard.

control_panel_api/tests/test_models.py
xoen: I'll give it a go 👍

control_panel_api/tests/test_models.py
andyhd: Fair enough - I just thought you were using it already

control_panel_api/tests/test_models.py
andyhd: 👍

control_panel_api/tests/test_models.py
xoen: As I said, these tests are using the `django.test.TestCase` class which is using atomic transactions (to clean the DB, etc...) which means exceptions would be caught and then `assertRaises` would not work: https://stackoverflow.com/a/24589930/455642

andyhd: That SO answer suggests you use `TransactionTestCase` - that seems like a good fit here

control_panel_api/tests/test_models.py
xoen: Alright, moving everything in a `TestCase` as these tests are all related to the membership 👍

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,89 @@
...
+            code="member",
+            name="Member",
+        )
+
+        # Add Alice to Justice team as maintainer
+        TeamMembership.objects.create(
+            team=cls.team_justice,
+            user=cls.user_alice,
+            role=cls.role_maintainer,
+        )
```
andyhd: This is a bit verbose - maybe in this test module there should be a function `add_user_to_team(user, team, role="Member")` or maybe a method `User.add_to_team(team, role)` or `Team.add_user(user, role)` or similar?

control_panel_api/tests/test_models.py
```diff
@@ -0,0 +1,89 @@
...
+            # (trying to) Add Alice to team justice again
+            TeamMembership.objects.create(
+                team=self.team_justice,
+                user=self.user_alice,
+                role=self.role_member,
+            )
+        except IntegrityError:
+            raised_integrity_error = True
+
+        assert raised_integrity_error
```
andyhd: This last test could be separated to allow using `assertRaises`, eg:
```python
class UniqueMembershipTest(TransactionTestCase):
    ...
    def test_user_can_be_added_to_team_only_once(self):
        with self.assertRaises(IntegrityError):
            team_justice.add_user(alice)
```
I realise this complicates the fixtures setup - you could have module-level variables `users = {}` and use `setUpModule` to create them, then in your tests you can get them with `users['Alice']` or similar.

andyhd: This last `assert` should be `self.assertTrue(raised_integrity_error)`

control_panel_api/tests/test_models.py
xoen: Yes, we may want to add some methods like that in the future.

control_panel_api/tests/test_models.py
xoen: Done 👍

PR #12, opened 2017-09-01
Add custom slugify for App.slug which will replace _ underscore, but …

jmoz:
## What

Add custom slugify for App.slug which will replace _ underscore, but still rely on standard implementation. Test replacement and also collisions where an id is incremented postfix

## How to review

1. Run tests

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -1,5 +1,8 @@
+import re
+
 from django.contrib.auth.models import AbstractUser
 from django.db import models
+from django.template.defaultfilters import slugify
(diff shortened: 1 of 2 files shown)


control_panel_api/models.py
```diff
@@ -25,8 +28,11 @@ def teams(self):
 
 
 class App(TimeStampedModel):
+    def _slugify(value):
+        return re.sub(r'_+', '-', slugify(value))
```
xoen: nice

xoen:
Looks good 👍

PR #13, opened 2017-09-01
Add S3Bucket model

xoen:
### What
This PR adds the `S3Bucket` model.

### How
It's mostly a plan model. The only thing to bear in mind here is that
S3 Bucket names have several restrictions:

http://docs.aws.amazon.com/en_gb/AmazonS3/latest/dev/BucketRestrictions.html

The restriction on the lenght is validated using the following regular
expression: `^.{3,63}$` which means any sequence of characters long between
3 and 63.

The other restrictions are validated by this more complex RegEx:

`^([a-z][a-z0-9-]*[a-z0-9])(.[a-z][a-z0-9-]*[a-z0-9])*$`

The first capture group `([a-z][a-z0-9-]*[a-z0-9])` is used to match
roughtly what AWS calls a "label" which is a string starting with a letter
and ending with a letter or a digit, containing other letters, digits or
hyphens.

**NOTE**: To prevent S3 bucket names like `127.0.0.1` (which are invalid)
here I'm imposing the limitation that each group needs to start with a
letter. This is not a limitation of S3 but it makes things simpler.

The following capture group `(.[a-z][a-z0-9-]*[a-z0-9])*` is almost the
same, it means "after the first label, the bucket name can contains other
labels separated by a dot".

You can see this regular expression in action here: http://rubular.com/r/D36J05yWp0

### Ticket

https://trello.com/c/Revdks5V/362-add-s3-bucket-model-to-control-panel-api


control_panel_api/tests/test_models.py
```diff
@@ -104,3 +106,17 @@ def test_slug_collisions_increments(self):
...
         app2 = App.objects.create(name=name)
         self.assertEqual('foo-2', app2.slug)
+
+
+class S3BucketTestCase(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        # Create an S3 bucket
+        cls.s3_bucket_1, _ = S3Bucket.objects.get_or_create(
```
jmoz: Here I think you just need `S3Bucket.objects.create()`

xoen:
Cheers @jmoz

PR #14, opened 2017-09-04
Add S3 bucket endpoints

xoen:
## What

Added S3 buckets endpoints.

For now only superusers have access to them. This will change once we'll introduce Apps DAGs (Data Access Groups).

I also added a `unique` constraint to `S3Bucket`'s `name` field.

This PR includes tests for permissions, filtering and also for the `S3Bucket` model validations.

## Ticket
https://trello.com/c/IDy34ERH/365-add-api-endpoint-and-crud-methods-for-s3-bucket


xoen:
Cheers @jmoz

PR #15, opened 2017-09-05
Add AppS3Bucket model

xoen:
## What

This is the model for many-to-many relationship between apps and S3 buckets.
The model will also keep the level of access the app will have to the S3 bucket (e.g. "readonly" or "readwrite").

These records will be used to determine which app can access what.

## Ticket
https://trello.com/c/aWYJKS0B/373-cp-api-add-apps-dag-data-access-group-model


xoen:
@jmoz as discussed offline renaming this model once again to `AppS3Bucket`: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/15/commits/4ebcce4742e91ae4d0ccaf571053948eca193b3f

I also slightly changed the test to use django's association `_set` as discussed earlier offline.

PR #16, opened 2017-09-05
Feature aws service create bucket

jmoz:
## What

Port over the lambda code to create bucket and policies. Add services module and aws module with settings.

- AWS API code in the `aws` module.
- Domain logic in `services`.
- Environment and settings variables in `settings`.

https://trello.com/c/kJ2GfbK5/372-cp-api-service-object-to-create-an-s3-bucket


control_panel_api/settings.py
```diff
@@ -122,3 +122,8 @@
 
 LOGIN_URL = 'rest_framework:login'
 LOGOUT_URL = 'rest_framework:logout'
+
+# AWS variables
+BUCKET_REGION = os.environ.get('BUCKET_REGION', 'eu-west-1')
+ENV = os.environ.get('ENV', 'dev')
```
xoen: This is necessarily an AWS-related setting.
Also, we'll need to find a way to set these environment variables when deploying (didn't look at that part)

control_panel_api/services.py
```diff
@@ -0,0 +1,29 @@
...
+
+from . import aws
+
+READ_WRITE = 'readwrite'
+READ_ONLY = 'readonly'
+
+
+def _bucket_name(name):
+    """Prefix the bucket name with environment e.g. dev-james"""
+    return "{}-{}".format(settings.ENV, name)
```
xoen: I'm wondering how this will work with the model. I feel like we should probably do this in the S3Bucket model somehow? Maybe while saving the record check if it starts with `dev-` already and if not add it behind the scenes?

control_panel_api/services.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def _policy_name(name, readwrite=False):
+    """Prefix the policy name with bucket name, postfix with access level e.g. dev-james-readwrite"""
+    return "{}-{}".format(_bucket_name(name), READ_WRITE if readwrite else READ_ONLY)
+
+
+def create_team_bucket(name):
```
xoen: This method name should be more generic now that we can also create S3 buckets for apps (or for any reason really). Maybe just call is `create_bucket`, what do you reckon?

control_panel_api/aws.py
```diff
@@ -0,0 +1,85 @@
+import json
+
+import boto3
+
+
+def create_bucket(name, **kwargs):
+    boto3.client('s3').create_bucket(
+        Bucket=name,
+        ACL=kwargs.get('acl', ''),
```
xoen: I would default to `private` here for the `acl` argument.

control_panel_api/services.py
```diff
@@ -0,0 +1,29 @@
...
+    """Creates an s3 bucket and adds logging"""
+    name = _bucket_name(name)
+    aws.create_bucket(name, acl='private', region=settings.BUCKET_REGION)
+    aws.put_bucket_logging(name, target_bucket=settings.LOGS_BUCKET_NAME, target_prefix="{}/".format(name))
+
+
+def create_team_bucket_policies(name):
```
xoen: I would also rename this as well: `create_team_bucket_policies()` => `create_bucket_policies()`

control_panel_api/aws.py
```diff
@@ -0,0 +1,85 @@
...
+        CreateBucketConfiguration={'LocationConstraint': kwargs.get('region', '')},
```
xoen: If we switch `kwargs.get()` with `kwargs["ket"]` it would blow up right?

I think we should use (in general) a meaningful default for these options - one there is one - and just let it raise an Exception when a meaningful default doesn't exist and we want to make the argument mandatory.

control_panel_api/aws.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+
+def put_bucket_logging(name, **kwargs):
+    boto3.client('s3').put_bucket_logging(
+        Bucket=name,
+        BucketLoggingStatus={
+            'LoggingEnabled': {
+                'TargetBucket': kwargs.get('target_bucket', ''),
```
xoen: See comment on using `kwargs["key"]`.

control_panel_api/aws.py
```diff
@@ -0,0 +1,85 @@
...
+                'TargetPrefix': kwargs.get('target_prefix', '')
+            }
+        }
+    )
+
+
+def create_policy(name, readwrite=False):
+    boto3.client("iam").create_policy(
+        PolicyName=name,
+        PolicyDocument=json.dumps(get_policy_document(name, readwrite)),
```
xoen: There is a problem here: The `name` here is the name of the policy (coming from the `services` module - however the `get_policy_document()` function use it as `bucket_name` (e.g. to build the `bucket_arn` and determine on which S3 bucket you can perform these operations).

This mean that as it stands the IAM policy document would not work because would contain things like `"Resource": ["arn:aws:s3:::dev-mybucket-readonly"]`. The resource/bucket arn should just be `arn:aws:s3:::dev-mybucket`

xoen: As a related comment, I think we should try to use variable names which are a bit more explicit, e.g. `bucket_name` or `policy_name` whenever possible to avoid this kind of confusions.

Although to be fair, I think if the function (`create_policy`) is clearly operating on a certain thing (`policy`) it would be OK to omit the `_policy` suffix, as in this case.

control_panel_api/services.py
jmoz: Remember the App has a slug which is regex validated, so as the input comes in it should be all good by the time we pass it into aws

control_panel_api/services.py
jmoz: 👍

control_panel_api/services.py
```diff
@@ -0,0 +1,29 @@
+from django.conf import settings
+
+from . import aws
```
xoen: Stupid question: Is this equivalent to `import aws`? Actually probably more like `from control_panel_api import aws`?

control_panel_api/aws.py
jmoz: 👍

control_panel_api/services.py
jmoz: 👍

control_panel_api/services.py
xoen: You mean the `S3Bucket` model right? Yes, I'm kind of wondering if we should add this prefix there and tweak the regex.

control_panel_api/aws.py
jmoz: No should still work, but using `get` is the idiom and gives you a default.  Yep for stuff that can take a default and we have an option (like 'private') can use that.

control_panel_api/aws.py
jmoz: `get` > dict access by key

xoen:
I like the overall split between "higher" level `services` and the lower lever `aws` module. We should probably test this (I know...).

There are some minor comments and what I think is a bug, so have a look at the comments :)

(Otherwise good stuff!!)

control_panel_api/services.py
jmoz: Done like this so that if there is a library named `aws` it won't confilct

control_panel_api/aws.py
```diff
@@ -0,0 +1,74 @@
+import json
+
+import boto3
+
+
+def create_bucket(name, region, **kwargs):
```
xoen: We may well (in this case with 3 args) to also add `acl='private'` to the list of arguments.

control_panel_api/aws.py
```diff
@@ -0,0 +1,74 @@
...
+        PolicyName=name,
+        PolicyDocument=json.dumps(policy_document)
+    )
+
+
+def delete_policy(policy_arn):
+    boto3.client("iam").delete_policy(PolicyArn=policy_arn)
+
+
+def detach_policy_from_entities(policy_arn):
```
xoen: As quickly discussed offline, I think until we really need to perform this operation individually we may just make this function "private" (rename to `_detach....`) and call always in `delete_policy()` before the actual deletion of the policy.

(I would imagine from the service level perspective this is an implementation detail, the client code doesn't really concern with whether you need to detach the policy from stuff or not in order to delete it)

control_panel_api/aws.py
```diff
@@ -0,0 +1,74 @@
...
+        detach_policy_from_role(policy_arn, role["RoleName"])
+
+    for group in entities["PolicyGroups"]:
+        detach_policy_from_group(policy_arn, group["GroupName"])
+
+    for user in entities["PolicyUsers"]:
+        detach_policy_from_user(policy_arn, user["UserName"])
+
+
+def detach_policy_from_role(policy_arn, role_name):
```
xoen: All these `detach_policy_from_*` functions could probably also be "private" as they're only used internally at this stage.

control_panel_api/services.py
```diff
@@ -0,0 +1,107 @@
...
+
+def _policy_arn(bucket_name, readwrite=False):
+    return "{}:policy/{}".format(settings.IAM_ARN_BASE, _policy_name(bucket_name, readwrite))
+
+
+def _bucket_arn(bucket_name):
+    return "arn:aws:s3:::{}".format(bucket_name)
+
+
+def get_policy_document(bucket_name, readwrite):
```
xoen: It may be the case of make this function "private" as well (=> `_get_policy_document()`) as I don't think the client code would really use it directly.

control_panel_api/services.py
```diff
@@ -0,0 +1,107 @@
...
+
+def create_bucket(name):
+    """Creates an s3 bucket and adds logging"""
+    bucket_name = _bucket_name(name)
+    aws.create_bucket(bucket_name, region=settings.BUCKET_REGION, acl='private')
+    aws.put_bucket_logging(bucket_name, target_bucket=settings.LOGS_BUCKET_NAME,
+                           target_prefix="{}/".format(bucket_name))
+
+
+def create_bucket_policies(name):
```
xoen: I'm wondering if this function should be embedded within the `create_bucket()` function. E.g. when I create an S3 bucket using this service it would create the policies behind the scenes.

I can't imagine any normal case where we want to create the S3 bucket but not the IAM policies to access it.

control_panel_api/services.py
```diff
@@ -0,0 +1,107 @@
...
+    aws.create_policy(_policy_name(bucket_name, readwrite=True), get_policy_document(bucket_name, True))
+    aws.create_policy(_policy_name(bucket_name, readwrite=False), get_policy_document(bucket_name, False))
+
+
+def delete_bucket_policies(name):
+    """Delete policy from attached entities first then delete policy, for both policy types"""
+    bucket_name = _bucket_name(name)
+
+    policy_arn_readwrite = _policy_arn(bucket_name, readwrite=True)
+    aws.detach_policy_from_entities(policy_arn_readwrite)
```
xoen: As per comment on the `aws.py` file, we can probably hide these `detach_policy_from_entities()` calls within `delete_bucket()` making this code a bit simpler/easier.

control_panel_api/tests/test_services.py
```diff
@@ -0,0 +1,16 @@
+from unittest.case import TestCase
+
+from control_panel_api import services
+
+
+class ServicesTestCase(TestCase):
+    def test_policy_document_readwrite(self):
```
xoen: (I know the tests are still work in progress so apologies in advance for commenting on this)

I this the tests should be a bit more in depth as the one in the lambda functions code: https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/data_access/teams/tests/conftest.py

Test the policies are correct is not the nicest thing in the world as these JSON documents are not exactly pretty :D - but at the same time we can protect ourselves from accidental future changes that break them (as that would have the effect of giving the wrong level of access to these S3 buckets to the users with this policies).

xoen:
I'm a bit wary of deleting S3 buckets.

For the time being I'd keep the same behaviour of the lambda functions and just delete the IAM policies to access an S3 bucket rather than actually delete the S3 bucket in AWS.

control_panel_api/settings.py
jmoz: That will all be devops side e.g. Andy

control_panel_api/services.py
```diff
@@ -0,0 +1,115 @@
...
+    aws.put_bucket_logging(bucket_name, target_bucket=settings.LOGS_BUCKET_NAME,
+                           target_prefix="{}/".format(bucket_name))
+
+
+def delete_bucket(name):
+    """Delete s3 bucket"""
+    aws.delete_bucket(_bucket_name(name))
+
+
+def create_bucket_policies(name):
```
xoen: This should be "private" and called within `create_bucket()` as from the caller code it doesn't make much sense to create an S3 bucket without the policies.

control_panel_api/settings.py
```diff
@@ -122,3 +122,9 @@
...
```
xoen: FYI: I added these environment variables to the helm chart used to deploy:
 - helm-chart config: https://github.com/ministryofjustice/analytics-platform-config/pull/8
 - helm-chart: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/16

control_panel_api/views.py
```diff
@@ -48,3 +49,14 @@ class S3BucketViewSet(viewsets.ModelViewSet):
...
+    def perform_create(self, serializer):
+        instance = serializer.save()
+        services.create_bucket(instance.name)
+        services.create_bucket_policies(instance.name)
+
+    def perform_destroy(self, instance):
+        name = instance.name
+        instance.delete()
+        services.delete_bucket_policies(name)
+        services.delete_bucket(name)
```
jmoz: @xoen here did you say NOT delete bucket? Could also add a settings flag to enable or disable

control_panel_api/services.py
jmoz: Actually I think the explicit usage of these calls is better for readability - I want to keep the aws library as simple as possible with no magic functionality or coupled calls - the service layer should orchestrate these calls and domain logic.

control_panel_api/aws.py
jmoz: Yes they are used just internally but they are still fundamentally aws api calls so I would keep them open for any possible future usage, and not make it confusing why they are designated private or public

control_panel_api/tests/test_services.py
```diff
@@ -0,0 +1,84 @@
...
+from control_panel_api import services
+
+
+@override_settings(ENV='test', BUCKET_REGION='eu-test-2', LOGS_BUCKET_NAME='moj-test-logs')
+class ServicesTestCase(SimpleTestCase):
+    def test_policy_document_readwrite(self):
+        document = services.get_policy_document('foo', readwrite=True)
+
+        sids = [s['Sid'] for s in document['Statement']]
+        self.assertIn('UpdateRenameAndDeleteObjects', sids)
```
xoen: IAM policies' Sid have no effects on what level of access the policy gives.
We shouldn't test for their name, we may accidentally break the policy document and tests will still pass.

I advice we do what we did in the lambda function test and we keep a copy of the expected policy documents in the tests: https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/data_access/teams/tests/conftest.py#L22

This duplication is good because if we change the policy document for whatever reason tests will fail unless we explicitly update the test policy documents.

xoen: Also, do we want to test this method? I wonder if this method should be private, unless we want some parts of the app to get a policy document directly (I don't think we do it at the moment).

control_panel_api/tests/test_services.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    @patch('boto3.client')
+    def test_create_bucket_policies(self, mock_client):
+        services.create_bucket_policies('bucketname')
+
+        expected_calls = [
+            call(
+                PolicyName='test-bucketname-readwrite',
+                PolicyDocument=json.dumps(services.get_policy_document('test-bucketname', True))
```
xoen: See other comment: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/16/files#r138314735

Use `services.get_policy_document` (the actual code) means it will always pass even if, let's say we break it big time, we add a `"s3:DeleteObject",` to the `ReadObjects` actions.

We should keep a copy of the expected policies in the tests (even if it's a duplication it's a good one in this case).

control_panel_api/tests/test_services.py
```diff
@@ -0,0 +1,84 @@
...
+            call(
+                PolicyName='test-bucketname-readonly',
+                PolicyDocument=json.dumps(services.get_policy_document('test-bucketname', False))
+            )
+        ]
+        mock_client.return_value.create_policy.assert_has_calls(expected_calls)
+
+
+@override_settings(ENV='test', IAM_ARN_BASE='arn:aws:iam::1337')
+class NamingTestCase(SimpleTestCase):
```
xoen: Good to have these tests :)

control_panel_api/tests/test_views.py
```diff
@@ -128,15 +125,24 @@ def test_detail(self):
...
         response = self.client.get(reverse('s3bucket-detail', (self.fixture.id,)))
         self.assertEqual(HTTP_404_NOT_FOUND, response.status_code)
 
-    def test_create_when_valid_data(self):
+    @patch('control_panel_api.services.delete_bucket')
+    @patch('control_panel_api.services.delete_bucket_policies')
+    def test_delete_calls_apis(self, mock_delete_bucket_policies, mock_delete_bucket):
+        self.client.delete(reverse('s3bucket-detail', (self.fixture.id,)))
+        mock_delete_bucket_policies.assert_called()
+        mock_delete_bucket.assert_called()
```
xoen: As I said, let's just delete the IAM policies and not the S3 bucket itself for the time being (we can change this later if really requested)

control_panel_api/tests/test_views.py
```diff
@@ -156,6 +162,14 @@ def test_create_when_name_invalid(self):
...
+    @patch('control_panel_api.services.create_bucket')
+    @patch('control_panel_api.services.create_bucket_policies')
+    def test_create_calls_apis(self, mock_create_bucket, mock_create_bucket_policies):
+        data = {'name': 'test-bucket-123'}
+        self.client.post(reverse('s3bucket-list'), data)
+        mock_create_bucket.assert_called()
+        mock_create_bucket_policies.assert_called()
+
     def test_update_when_valid_data(self):
         data = {'name': 'test-bucket-updated'}
```
xoen: I don't know how to enforce this in django but we probably want to prevent S3 buckets renaming as it's non-trivial and we should treat S3 buckets are permanent things as apps, IAM policies, etc...refer to them so it's easy to break things if renaming them.

Also see: https://stackoverflow.com/questions/41511034/how-to-rename-aws-s3-bucket-name

control_panel_api/views.py
```diff
@@ -48,3 +49,14 @@ class S3BucketViewSet(viewsets.ModelViewSet):
     serializer_class = S3BucketSerializer
     filter_backends = (S3BucketFilter,)
     permission_classes = (S3BucketPermissions, )
+
+    def perform_create(self, serializer):
+        instance = serializer.save()
+        services.create_bucket(instance.name)
+        services.create_bucket_policies(instance.name)
```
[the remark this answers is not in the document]
jmoz: I think being explicit is better than implicit.  If it was coupled inside create_bucket then we would have to do the same with the delete_bucket - but as just demonstrated we don't want to delete the bucket at the same time as the policy.

control_panel_api/tests/test_views.py
jmoz: Yep no renames for the timebeing

control_panel_api/tests/test_views.py
jmoz: Cool, just removed it. Another example of a good test - removing the delete call in the view caused the tests fail.

control_panel_api/settings.py
xoen: Yeah, I was adding the role to the helm chart and as I was in there I added these environment variables you need ;)

control_panel_api/services.py
xoen: I agree, the `aws` layer show be very explicit, but the `services` layer here can abstract/hide the creation of the IAM policies inside the `create_bucket()` function.

control_panel_api/aws.py
xoen: Mmmm, we can keep them public if you think it's better 👍 

In my head this is still something the outside code doesn't care because it's only necessary to delete an IAM role (at this stage).

control_panel_api/views.py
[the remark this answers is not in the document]
xoen: I don't know, I thought the idea behind having the `services` module was that it would orchestrate these things for the client code. e.g. you ask `services` "create an S3 bucket" and it would take care of it.

It's similar to ask a waiter a dish, the waiter just knows who to ask the dish, not how to cook it, what kind of ingredients needs,  etc...I see `services` as an interface between our app code and AWS (in this case) => the app wants an S3 bucket, what happens behind the scenes is not important from the view perspective.

xoen:
Hey @jmoz I wonder what we should approach the bucket name thing. As I said, user may post a bucket name which doesn't include the prefix (e.g. doesn't start with `dev_`) but we want to enforce that somehow.

Shouldn't we add that prefix if the S3 bucket name in the request body doesn't start with `dev_`? (or whatever `settings.ENV` is)? But this means the `services` module shouldn't add it again/handle it. What do you think?

This is the only bit which makes me slightly nervous as we want to continue to use this naming convention to namespace S3 buckets in different environments.

jmoz:
@xoen the dev- naming thing is from the environment, not the user, so the services takes care of adding the env from the environment config - this is enforced in tests.

xoen:
@jmoz let's chat about this in Slack (I already had a chat with Kerin to have a second opinion and to be sure I wasn't completely crazy).

control_panel_api/services.py
```diff
@@ -121,3 +121,15 @@ def delete_bucket_policies(name):
     policy_arn_readonly = _policy_arn(bucket_name, readwrite=False)
     aws.detach_policy_from_entities(policy_arn_readonly)
     aws.delete_policy(policy_arn_readonly)
+
+
+def bucket_create(name):
```
xoen: One general observation about the names of these functions in this module (and probably in general).

I would rename these to make very explicit what's "private", e.g. internal utility function (at least for now) and what's the public interface of the module.

So for example we may have `services.create_bucket()` which calls internal things (e.g. `_create_bucket()` and `_create_bucket_policies()`).

I like the fact that someone can get in this module without having much knowledge of it and have a feeling of what can move around and what's a contract with the client code (which you can still change if needed, but you'll need to be more careful).

control_panel_api/tests/test_views.py
```diff
@@ -160,13 +160,11 @@ def test_create_when_name_invalid(self):
...
-    @patch('control_panel_api.services.create_bucket')
-    @patch('control_panel_api.services.create_bucket_policies')
-    def test_create_calls_apis(self, mock_create_bucket, mock_create_bucket_policies):
+    @patch('control_panel_api.services.bucket_create')
+    def test_create_calls_apis(self, mock_bucket_create):
         data = {'name': 'test-bucket-123'}
         self.client.post(reverse('s3bucket-list'), data)
-        mock_create_bucket.assert_called()
-        mock_create_bucket_policies.assert_called()
+        mock_bucket_create.assert_called()
```
xoen: I think I prefer this more, the `services` layer is now abstracting out the "complexity" of creating an S3 bucket and related access policies.

control_panel_api/tests/test_services.py
```diff
@@ -19,50 +20,47 @@ def test_policy_document_readwrite(self):
...
         expected_calls = [
-            call(
-                PolicyName='test-bucketname-readwrite',
-                PolicyDocument=json.dumps(services.get_policy_document('test-bucketname', True))
-            ),
-            call(
-                PolicyName='test-bucketname-readonly',
-                PolicyDocument=json.dumps(services.get_policy_document('test-bucketname', False))
-            )
+            call('test-bucketname-readwrite', services.get_policy_document('test-bucketname', True)),
```
xoen: I still think using `services.get_policy_document()` (the real thing), in this tests make them very easy to break. 

Ping me on slack if this sound confusing, I may be able to explain better there.

control_panel_api/models.py
```diff
@@ -62,7 +57,7 @@ class Meta:
 
     @property
     def arn(self):
-        return "arn:aws:s3:::{}".format(self.name)
+        return services.bucket_arn(self.name)
```
xoen: It makes sense to factor this out 👍

control_panel_api/services.py
xoen: As for yesterday's (Slack) discussion, we're keeping it simple and "trust" the user (guided via the UI) to pass an S3 bucket which starts with the right env name.

So basically here we don't prefix the bucket name with `settings.ENV` anymore as the name coming from the request/in the `S3Bucket` model already starts with `settings.ENV` (but we should probably tweak the `S3Bucket` model to validate this.

control_panel_api/services.py
```diff
@@ -0,0 +1,135 @@
...
+                "s3:ListBucket"
+            ],
+            "Effect": "Allow",
+            "Resource": [bucket_name_arn],
+        },
+        {
+            "Sid": "ReadObjects",
+            "Action": [
+                "s3:GetObject",
+                "s3:GetObjectAcl",
```
xoen: Try to add `"s3:DeleteObject",` here and run the tests ;)

control_panel_api/services.py
```diff
@@ -0,0 +1,135 @@
...
+    """Return bucket arn e.g. arn:aws:s3:::bucketname"""
+    return "arn:aws:s3:::{}".format(bucket_name)
+
+
+def app_slug(name):
+    """Create a valid slug for s3 using standard django slugify but we add some custom"""
+    return re.sub(r'_+', '-', slugify(name))
+
+
+def get_policy_document(bucket_name, readwrite):
```
xoen: This is internal only at the moment, prefix with `_`?

control_panel_api/services.py
```diff
@@ -0,0 +1,135 @@
...
+
+def create_bucket(name):
+    """Create an s3 bucket and add logging"""
+    bucket_name = _bucket_name(name)
+    aws.create_bucket(bucket_name, region=settings.BUCKET_REGION, acl='private')
+    aws.put_bucket_logging(bucket_name, target_bucket=settings.LOGS_BUCKET_NAME,
+                           target_prefix="{}/".format(bucket_name))
+
+
+def delete_bucket(name):
```
xoen: Is this used anywhere? We may well delete this for now? (no pun intended :P )

control_panel_api/services.py
xoen: This is internal only at the moment, prefix with `_`?

control_panel_api/services.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def create_bucket_policies(name):
+    """Create readwrite and readonly policies for s3 bucket"""
+    bucket_name = _bucket_name(name)
+    aws.create_policy(_policy_name(bucket_name, readwrite=True), get_policy_document(bucket_name, True))
+    aws.create_policy(_policy_name(bucket_name, readwrite=False), get_policy_document(bucket_name, False))
+
+
+def delete_bucket_policies(name):
```
xoen: This is internal only at the moment, prefix with `_`?

control_panel_api/tests/test_aws.py
```diff
@@ -0,0 +1,65 @@
...
+from unittest.mock import patch
+
+from control_panel_api import aws
+
+
+@patch('boto3.client')
+class AwsTestCase(TestCase):
+    def test_create_bucket(self, mock_client):
+        aws.create_bucket('bucketname', 'eu-west-1')
+        mock_client.return_value.create_bucket.assert_called()
```
xoen: Can't remember how to assert a method was called with certain params. 
Do you think it may be worth to be a bit more specific in some of these tests?

control_panel_api/tests/test_aws.py
```diff
@@ -0,0 +1,65 @@
...
+        aws.delete_bucket('bucketname')
+        mock_client.return_value.delete_bucket.assert_called()
+
+    def test_put_bucket_logging(self, mock_client):
+        aws.put_bucket_logging('bucketname', 'target_bucket', 'target_prefix')
+        mock_client.return_value.put_bucket_logging.assert_called()
+
+    def test_create_policy_json_encoded(self, mock_client):
+        aws.create_policy('bucketname', {'foo': 'bar'})
+        mock_client.return_value.create_policy.assert_called_with(
```
xoen: Oh, there we are, this is the way you test a method is called with certain params :)

control_panel_api/tests/test_aws.py
```diff
@@ -0,0 +1,65 @@
...
+            PolicyName='bucketname',
+            PolicyDocument='{"foo": "bar"}'
+        )
+
+    def test_delete_policy(self, mock_client):
```
xoen: One thing I try to do when doesn't make things incredibly annoying to test is to only tests the public interface of a module/class (which is your "contract" with the client code).

In this specific module maybe should we test only "public" functions not internal ones? This is more a generic thing so I'm not saying to change it.

The reason why usually makes sense to test only the public functions is that then you don't "break" the tests every time you refactor/rename internal/utility functions. (don't know if this make sense). E.g. you may decide to ditch the `delete_policy()` function or rename it to `_delete_policy()` or change some params and the tests would only cover the public functions and the expected side effects.

**BUT** as I said, I'm not saying to change this right now, it's just a general comment.

control_panel_api/tests/test_aws.py
jmoz: I've tried to not make the tests too granular and brittle if not necessary.  These are basic parameter passing no real domain logic in them.

control_panel_api/services.py
jmoz: Can delete.

control_panel_api/services.py
jmoz: Yep this makes sense as is real internal utility function.  However some of your other _ prefix recommendations I think are being a bit too strict and confusing a public function which is just not used with internal private functions.

control_panel_api/tests/test_services.py
jmoz: Ok I can test the entire doc - I was just testing for the addition of that extra part.

control_panel_api/tests/test_services.py
```diff
@@ -0,0 +1,96 @@
...
+    @patch('control_panel_api.services.create_bucket')
+    @patch('control_panel_api.services.create_bucket_policies')
+    def test_bucket_create(self, mock_create_bucket_policies, mock_create_bucket):
+        services.bucket_create('bucketname')
+
+        mock_create_bucket_policies.assert_called()
+        mock_create_bucket.assert_called()
+
+    @patch('control_panel_api.services.delete_bucket_policies')
+    def test_bucket_create(self, mock_delete_bucket_policies):
```
xoen: I think the name of this method is wrong and as far as I know (with my limited Python knowledge), is overriding the method just above?

I think this is actually testing the "deletion" of an S3 bucket (which as per tests, only delete the policies 👍 )

control_panel_api/tests/test_services.py
```diff
@@ -0,0 +1,96 @@
...
+        services.bucket_delete('bucketname')
+
+        mock_delete_bucket_policies.assert_called()
+
+
+@override_settings(ENV='test', IAM_ARN_BASE='arn:aws:iam::1337')
+class NamingTestCase(SimpleTestCase):
+    def test_bucket_name_has_env(self):
```
xoen: As discussed in Slack yesterday and also as per other comment, the user now provides the actual S3 bucket name including the env prefix, so we don't need to prefix this anymore.

Basically UI will make this request:

```
POST /s3buckets

{
  "name": "dev-bucketname"
}
```

So we don't need to alter it (yay :) )

control_panel_api/tests/test_aws.py
xoen: I see, so depending on the thing you're testing and it's probability to write it wrong you write more or less detailed tests?

OK, I would probably be more specific when there are possible security implications at the cost of making tests more brittle (e.g. policy document).

control_panel_api/tests/test_services.py
jmoz: Ok but I would rather do that in another branch of work I want to get this merged first

control_panel_api/tests/test_services.py
jmoz: Well spotted.

control_panel_api/tests/test_services.py
xoen: Fair enough, see if remove the env is too involved otherwise don't worry 👍

xoen:
FINAL WORDS (Apologies for all the noise James!):

I think overall looks good. Tests on policy documents may be a bit more specific and there may few bits to tweak but you (and me!) we can do that as a separate step. So feel free to merge this once you're reasonably happy with it 👍

control_panel_api/tests/test_services.py
jmoz: @aldo

xoen: I've seen the new tests which check the policy documents 👍

xoen:
🎆 🎆 🎆

PR #18, opened 2017-09-13
Enable continuous deployment of Control Panel API branches to Kubernetes

andyhd:
## What

* Add Circle CI config, to build a docker image for a Github PR, run tests, push to quay.io and trigger a Jenkins build
* Add `Jenkinsfile` to deploy docker image to Kubernetes cluster using helm

## How to review

1. Create a PR for `analytics-platform-control-panel`
2. See Circle CI build of PR branch (eg: https://circleci.com/gh/ministryofjustice/analytics-platform-control-panel/155)
3. See Jenkins build triggered by Circle CI (eg: https://jenkins.services.dev.mojanalytics.xyz/job/control-panel/job/continuous-deployment/9/)
4. See control panel API is deployed (eg: https://control-panel.services.dev.mojanalytics.xyz/)

See [Trello ticket #342](https://trello.com/c/jxufKhNR)

Also see https://github.com/ministryofjustice/analytics-platform-jenkins-lib/pull/19

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
       - run:
           name: Install dependencies
           command: |
-            apk update && apk add py2-pip
+            apk update && apk add curl py2-pip
             pip install docker-compose==1.15.0
 
       - restore_cache:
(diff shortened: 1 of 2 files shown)


andyhd:
I've just realised that all branches will deploy to the same hostname, so still need a bit of work to the helm chart to allow deploying at branch-specific URLs.

xoen:
As far as I can tell looks good to me @andyhd 👍

xoen:
@andyhd maybe we should only deploy if the branch is `master` for the time being?
I mean, sometimes a branch could be there but would not be ready yet so probably we should prevent that branch to take over a deploy from `master` (a stable version).

andyhd:
It will only build PR branches, so if you are working on a branch it will not build it until you make a PR.

xoen:
If someone creates a PR (so code which is not ready to be merged yet as it's being reviewed) would that cause a deployment?

I need to have another look at the Circle config...

xoen:
...yes as far as I can tell when it runs the tests and all is happy it would push the docker image and trigger the jenkins job to deploy. So I'm thinking maybe this last step should happen only for master for the time being? What do you reckon @andyhd ?

andyhd:
@xoen that's the point - it deploys PRs so we can test features before merging to master.

PR #19, opened 2017-09-15
Create/Delete AWS IAM roles for apps

xoen:
### What

- When an app is created we create its IAM role in AWS.
- The app role name would follow this pattern: `$ENV_app_$APP_NAME`
- When an app is deleted we delete the corresponding IAM role
- IAM roles will be assumed by the app's k8s pods this means the role
  assume policy needs to contain the k8s worker role ARN as principal
  (See [`sts:AssumeRole` required by kube2iam](https://github.com/jtblin/kube2iam#iam-roles))
- This ARN is passed in the settings via the `K8S_WORKER_ROLE_ARN`
  environment variable
- I'll **need** to update the cpanel helm chart and the cpanel helm chart config
  to pass this setting in order for this to work

### Related code

This is somewhat similar to what we already to in the users lamba functions:

https://github.com/ministryofjustice/analytics-platform-ops/blob/19046f0c10da4cde551bdc6f0847e332b0d92a7d/infra/terraform/modules/data_access/users/users.py#L18

Differences are:
* Here we create an IAM role for the app, not the user, so the IAM role name includes `_app_`
* We do **not** attach the `sts:AssumeRoleWithSAML` to the assume role policy as this role will not be assumed via SAML.

### Notes regarding the tests/design decisions

The code is mainly split in 3 layers:

`views` <=> `services` <=> `aws`

Each layer's tests only test that the right calls to the adiacent layer
are perfomed correctly.

The tests only test the "public" functions treating them as "black boxes"
and testing that given a certain input certain side effects happen.

As a concious design decision the `aws._detach_role_policies()` is "private".
This is (at the moment!) only used internally by `aws.delete_role()` when
(and if!) there will be the need to call it from the outside we'll make it
"public", until that moment this is an internal implementation detail
(which is of course tested) but not leaked into the client code.

**IMPORTANT**: We should try to find a way of mock every boto3 call while
running the tests as a precautionary measure: as it stands it's possible
to accidentaly write tests that could potentially interact with the real AWS
(which is bad). At the moment I renamed my AWS credentials locally to be sure
in this event boto3 would simply raise a "missing credentials" exception.
This is not ideal and we should discuss a possible solution for this.

### Ticket

https://trello.com/c/AULNJYns/359-add-iam-role-for-control-panel-api-app-to-terraform

Title the PR to complete the sentence: "Merging this PR will ..."

## Other resources

* [`users` lambda functions](https://github.com/ministryofjustice/analytics-platform-ops/blob/19046f0c10da4cde551bdc6f0847e332b0d92a7d/infra/terraform/modules/data_access/users/users.py#L18)
* [`sts:AssumeRole` required by kube2iam (k8s daemon used to attach IAM roles to pods)](https://github.com/jtblin/kube2iam#iam-roles)
* [SO question about IAM role not working with more details on AWS' `sts:AssumeRole`](https://stackoverflow.com/questions/21956794/aws-assumerole-authorization-not-working/33850060#33850060)

### DIFF
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -40,7 +40,7 @@ def delete_policy(policy_arn):
 def create_role(role_name, assume_role_policy):
     """Creates IAM role with the given name"""
 
-    boto3.client("iam").create_role(
+    aws_api_client("iam").create_role(
         RoleName=role_name,
         AssumeRolePolicyDocument=json.dumps(assume_role_policy)
     )
@@ -50,13 +50,13 @@ def delete_role(role_name):
     """Delete the given IAM role."""
 
     _detach_role_policies(role_name)
-    boto3.client("iam").delete_role(RoleName=role_name)
+    aws_api_client("iam").delete_role(RoleName=role_name)
 
 
 def _detach_role_policies(role_name):
     """Detaches all the policies from the given role"""
 
-    client = boto3.client("iam")
+    client = aws_api_client("iam")
 
     policies = client.list_attached_role_policies(RoleName=role_name)
     for policy in policies["AttachedPolicies"]:
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -37,6 +37,35 @@ def delete_policy(policy_arn):
     aws_api_client("iam").delete_policy(PolicyArn=policy_arn)
 
 
+def create_role(role_name, assume_role_policy):
+    """Creates IAM role with the given name"""
+
+    boto3.client("iam").create_role(
+        RoleName=role_name,
+        AssumeRolePolicyDocument=json.dumps(assume_role_policy)
+    )
+
+
+def delete_role(role_name):
+    """Delete the given IAM role."""
+
+    _detach_role_policies(role_name)
+    boto3.client("iam").delete_role(RoleName=role_name)
+
+
+def _detach_role_policies(role_name):
+    """Detaches all the policies from the given role"""
+
+    client = boto3.client("iam")
+
+    policies = client.list_attached_role_policies(RoleName=role_name)
+    for policy in policies["AttachedPolicies"]:
+        detach_policy_from_role(
+            role_name=role_name,
+            policy_arn=policy["PolicyArn"],
+        )
+
+
 def detach_policy_from_entities(policy_arn):
     """Get all entities to which policy is attached first then call separate detach operations"""
     entities = aws_api_client("iam").list_entities_for_policy(PolicyArn=policy_arn)
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -56,9 +56,7 @@ def delete_role(role_name):
 def _detach_role_policies(role_name):
     """Detaches all the policies from the given role"""
 
-    client = aws_api_client("iam")
-
-    policies = client.list_attached_role_policies(RoleName=role_name)
+    policies = aws_api_client("iam").list_attached_role_policies(RoleName=role_name)
     for policy in policies["AttachedPolicies"]:
         detach_policy_from_role(
             role_name=role_name,
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -57,7 +57,7 @@ def _create_app_role(role_name):
             {
                 "Effect": "Allow",
                 "Principal": {
-                    "AWS": settings.K8S_WORKER_ROLE_ARN,
+                    "AWS": f"{settings.IAM_ARN_BASE}:role/{settings.K8S_WORKER_ROLE_NAME}",
                 },
                 "Action": "sts:AssumeRole",
             }
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -130,7 +130,6 @@ def get_policy_document(bucket_name, readwrite):
 
 def create_bucket(bucket_name):
     """Create an s3 bucket and add logging"""
-    bucket_name = _bucket_name(name)
     aws.create_bucket(
         bucket_name, region=settings.BUCKET_REGION, acl='private')
     aws.put_bucket_logging(bucket_name, target_bucket=settings.LOGS_BUCKET_NAME,
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -14,6 +14,11 @@ def _policy_name(bucket_name, readwrite=False):
     return "{}-{}".format(bucket_name, READ_WRITE if readwrite else READ_ONLY)
 
 
+def _app_role_name(app_slug):
+    """Return the IAM role name for the given `app_slug`"""
+    return "{}_app_{}".format(settings.ENV, app_slug)
+
+
 def _policy_arn(bucket_name, readwrite=False):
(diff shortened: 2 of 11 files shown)


control_panel_api/settings.py
```diff
@@ -127,4 +127,5 @@
 BUCKET_REGION = os.environ.get('BUCKET_REGION', 'eu-west-1')
 ENV = os.environ.get('ENV', 'dev')
 LOGS_BUCKET_NAME = os.environ.get('LOGS_BUCKET_NAME', 'moj-analytics-s3-logs')
-IAM_ARN_BASE = os.environ.get('IAM_ARN_BASE', 'arn:aws:iam::593291632749')
\ No newline at end of file
+IAM_ARN_BASE = os.environ.get('IAM_ARN_BASE', 'arn:aws:iam::593291632749')
```
xoen: This is not considered sensitive (from discussion in `#security`) but I still think we shouldn't have it as default setting.

xoen:
FYI: I renamed the `K8S_WORKER_ROLE_ARN` setting to `K8S_WORKER_ROLE_NAME` as suggested by @kerin [here in the other PR](https://github.com/ministryofjustice/analytics-platform-config/pull/10#pullrequestreview-63053919): 

https://github.com/ministryofjustice/analytics-platform-control-panel/pull/19/commits/d951f41ba175ac9fce661626777c08f44db5dcde

control_panel_api/tests/test_services.py
```diff
@@ -66,21 +81,45 @@ def test_bucket_delete(self, mock_delete_bucket_policies):
 
         mock_delete_bucket_policies.assert_called()
 
+    @patch('control_panel_api.aws.create_role')
+    def test_app_create(self, mock_create_role):
+        app_slug = 'appname'
+        services.app_create(app_slug)
+
+        expected_role_name = "test_app_{}".format(app_slug)
```
jmoz: Is there a conflict here between naming conventions?  We hyphenate s3 buckets `dev-test-foo` so would it be a better idea to maintain that convention here?

control_panel_api/tests/test_services.py
```diff
@@ -41,14 +56,14 @@ def test_delete_bucket_policies(self, mock_detach_policy_from_entities, mock_del
         services.delete_bucket_policies('bucketname')
 
         expected_calls = [
-            call('arn:aws:iam::1337:policy/test-bucketname-readwrite'),
-            call('arn:aws:iam::1337:policy/test-bucketname-readonly')
+            call(f'{IAM_ARN_BASE}:policy/test-bucketname-readwrite'),
```
jmoz: Nice f-strings 👍

control_panel_api/views.py
```diff
@@ -45,6 +45,14 @@ class AppViewSet(viewsets.ModelViewSet):
     filter_backends = (AppFilter,)
     permission_classes = (AppPermissions,)
 
+    def perform_create(self, serializer):
+        app = serializer.save()
+        services.app_create(app.slug)
+
+    def perform_destroy(self, instance):
+        instance.delete()
+        services.app_delete(instance.slug)
```
jmoz: Is the instance and it's slug still valid here after deletion?

control_panel_api/services.py
```diff
@@ -34,6 +39,44 @@ def app_slug(name):
...
+    """Creates the IAM role for the given app"""
+    _create_app_role(_app_role_name(app_slug))
+
+
+def _create_app_role(role_name):
+    """Creates the IAM role for the given app"""
+
+    # See: `sts:AssumeRole` required by kube2iam
+    # https://github.com/jtblin/kube2iam#iam-roles
+    assume_role_policy = {
```
jmoz: I think the policy document could be extracted out of the function. Maybe stored somehwere else or put in a function?  Some consistency how this is done with the bucket `get_policy_document()` would be good.

xoen:
FYI @jmoz : I rebased but now tests fail (some of them are trivial, some, I need to check) so I need to fix them.

control_panel_api/services.py
```diff
@@ -19,6 +19,11 @@ def _policy_name(bucket_name, readwrite=False):
     return "{}-{}".format(bucket_name, READ_WRITE if readwrite else READ_ONLY)
 
 
+def _app_role_name(app_slug):
+    """Return the IAM role name for the given `app_slug`"""
+    return "{}_app_{}".format(settings.ENV, app_slug)
```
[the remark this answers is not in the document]
xoen: In short no. The bucket name is something visible to the end user, the user enter the S3 bucket name directly and also needs to know the exact S3 bucket name (as it may write some code to read/write from it).

The IAM role here is something needed behind by the app to access the S3 bucket but its name is generated off the S3 bucket - not entered by the user. (but we still want to keep the same convention of prefixing it with the env name).

Let me know if this not clear and I can try to explain better 👍

control_panel_api/services.py
xoen: This is a bit confusing because IAM policy documents and role assume policy document have the same syntax but different meaning.

I first thought of moving it but then I realised that we'll do it later (IF needed - YAGNI) when we'll add the logic to create the IAM role for the users to remove this future duplication.

For now, let's leave it inline and we'll move it out later when we'll need to have slightly different assume role policies for the users (because they'll need the additional statement for the SAML federated login).

control_panel_api/tests/test_services.py
xoen: As discussed this morning via Slack, I can't remember why I didn't use hyphens (`-`) for the IAM role names. I looked again this morning and it seems like [in theory they should be legal](http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html) (alongside "weirder" symbols like `=` or `@` also appear to be legal...)

So...even if breaking with past (lambda functions) convention I think you're right in saying it would be nice to be consistent here...I'll change (we'll find out if it will break and there was actually a reason for not using hyphens 😱 )

control_panel_api/views.py
xoen: I tried to inspect using `pdb` and the app's slug was there after the `delete` (but predictably not the `id` :D) - Also, tests pass so looks like get the app slug like this is valid.

Do you have any strong preference to get it into a variable before the `delete` and want me to change it?

control_panel_api/aws.py
```diff
@@ -37,6 +37,35 @@ def delete_policy(policy_arn):
...
+    """Delete the given IAM role."""
+
+    _detach_role_policies(role_name)
+    aws_api_client("iam").delete_role(RoleName=role_name)
+
+
+def _detach_role_policies(role_name):
+    """Detaches all the policies from the given role"""
+
+    client = aws_api_client("iam")
```
jmoz: As per the other functions in this file you can chain the call instead of assigning to `client`. For consistency.

xoen: This was coming from a time where inside the loop we also needed the client. 
Changing it now 👍

xoen: Done: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/19/commits/890936b97c71b5c9eb7f370e370cf651fc60617c

xoen:
BAAAAAM

PR #20, opened 2017-09-15
Remove magic bucket naming and validate on bucket name create

jmoz:
## What

Remove magic bucket naming and validate on bucket name create

## How to review

1. Run tests

https://trello.com/c/WWG3HLi2/386-bucket-naming-should-be-validated-on-creation-as-such-dev-bucketname-and-server-check-for-correct-env

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -34,7 +34,7 @@ class S3Bucket(TimeStampedModel):
     name = models.CharField(unique=True, max_length=63, validators=[
         validators.validate_env_prefix,
         validators.validate_s3_bucket_length,
-        validators.validate_s3_bucket_label,
+        validators.validate_s3_bucket_labels,
     ])
 
     class Meta:
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -1,10 +1,9 @@
 from django.contrib.auth.models import AbstractUser
-from django.core.validators import RegexValidator
 from django.db import models
 from django_extensions.db.fields import AutoSlugField
 from django_extensions.db.models import TimeStampedModel
 
-from control_panel_api import services
+from control_panel_api import services, validators
 
 
 class User(AbstractUser):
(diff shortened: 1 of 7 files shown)


control_panel_api/models.py
```diff
@@ -32,24 +31,10 @@ class Meta:
 
 
 class S3Bucket(TimeStampedModel):
-    # See: AWS' Bucket Restrictions and Limitations
```
xoen: I think I like it more with the validations in a separate module 👍

control_panel_api/models.py
```diff
@@ -32,24 +31,10 @@ class Meta:
...
-    LENGTH_REGEX = '^.{3,63}$'
-
     name = models.CharField(unique=True, max_length=63, validators=[
-        RegexValidator(regex=LENGTH_REGEX,
-                       message="must be between 3 and 63 characters"),
-        RegexValidator(regex=LABELS_REGEX,
-                       message="is invalid, check AWS S3 bucket names restrictions (for example, can only contains letters, digits, dots and hyphens)"),
+        validators.validate_env_prefix,
+        validators.validate_s3_bucket_length,
+        validators.validate_s3_bucket_label,
```
xoen: I know the name of this wasn't probably great in the first place but I'd probably call this `validators.validate_s3_bucket_labels` (plural) as the idea was for this regex to validate that the S3 bucket name was 1+ labels (as defined by AWS S3 bucket naming restrictions document).

control_panel_api/tests/test_views.py
```diff
@@ -239,11 +241,27 @@ def test_create_when_name_invalid(self):
...
+        data = {'name': 'bucketname'}
+        response = self.client.post(reverse('s3bucket-list'), data)
+        self.assertEqual(HTTP_400_BAD_REQUEST, response.status_code)
+
     @patch('control_panel_api.services.bucket_create')
     def test_create_calls_apis(self, mock_bucket_create):
         data = {'name': 'test-bucket-123'}
         self.client.post(reverse('s3bucket-list'), data)
-        mock_bucket_create.assert_called()
+        mock_bucket_create.assert_called_with('test-bucket-123')
```
xoen: better 👍

xoen:
Looks good to me 👍

control_panel_api/models.py
jmoz: Done

PR #21, opened 2017-09-18
Refactor settings structure and use mock aws client 

jmoz:
## What

Allow the switching of the boto3 client to a magicmock during test so no live calls are made (which has happened whilst developing).

- This refactors into a conventional multi settings file layout.
- Regular running of ./manage.py tasks will continue to work as settings package imports the base.py.
- No longer need to patch boto3 in view tests etc to prevent a live call.
- No longer need to override settings in tests to add test values.
- **Now, tests must be run as `./manage.py test --settings=control_panel_api.settings.test`**

### DIFF
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -2,6 +2,7 @@ import json
 
 from django.conf import settings
 
+
 aws_api_client = settings.AWS_API_CLIENT_HANDLER
 
 
@@ -37,8 +38,7 @@ def delete_policy(policy_arn):
 
 
 def detach_policy_from_entities(policy_arn):
-    """Get all entities to which policy is attached
-    See: http://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.list_entities_for_policy"""
+    """Get all entities to which policy is attached first then call separate detach operations"""
     entities = aws_api_client("iam").list_entities_for_policy(PolicyArn=policy_arn)
 
     for role in entities["PolicyRoles"]:
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -1,10 +1,12 @@
 import json
 
-import boto3
+from django.conf import settings
+
+aws_api_client = settings.AWS_API_CLIENT_HANDLER
 
 
 def create_bucket(name, region, acl='private'):
-    boto3.client('s3').create_bucket(
+    aws_api_client('s3').create_bucket(
         Bucket=name,
         ACL=acl,
         CreateBucketConfiguration={'LocationConstraint': region},
@@ -12,7 +14,7 @@ def create_bucket(name, region, acl='private'):
 
 
 def put_bucket_logging(name, target_bucket, target_prefix):
-    boto3.client('s3').put_bucket_logging(
+    aws_api_client('s3').put_bucket_logging(
         Bucket=name,
         BucketLoggingStatus={
             'LoggingEnabled': {
@@ -24,20 +26,20 @@ def put_bucket_logging(name, target_bucket, target_prefix):
 
 
 def create_policy(name, policy_document):
-    boto3.client("iam").create_policy(
+    aws_api_client("iam").create_policy(
         PolicyName=name,
         PolicyDocument=json.dumps(policy_document)
     )
 
 
 def delete_policy(policy_arn):
-    boto3.client("iam").delete_policy(PolicyArn=policy_arn)
+    aws_api_client("iam").delete_policy(PolicyArn=policy_arn)
 
 
 def detach_policy_from_entities(policy_arn):
     """Get all entities to which policy is attached
-    See: http://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.list_entities_for_policy"""
-    entities = boto3.client("iam").list_entities_for_policy(PolicyArn=policy_arn)
+    See: http://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.list_entities_for_policy"""
+    entities = aws_api_client("iam").list_entities_for_policy(PolicyArn=policy_arn)
 
     for role in entities["PolicyRoles"]:
         detach_policy_from_role(policy_arn, role["RoleName"])
@@ -50,21 +52,21 @@ def detach_policy_from_entities(policy_arn):
 
 
 def detach_policy_from_role(policy_arn, role_name):
-    boto3.client("iam").detach_role_policy(
+    aws_api_client("iam").detach_role_policy(
         RoleName=role_name,
         PolicyArn=policy_arn,
     )
 
 
 def detach_policy_from_group(policy_arn, group_name):
-    boto3.client("iam").detach_group_policy(
+    aws_api_client("iam").detach_group_policy(
         GroupName=group_name,
         PolicyArn=policy_arn,
     )
 
 
 def detach_policy_from_user(policy_arn, user_name):
-    boto3.client("iam").detach_user_policy(
+    aws_api_client("iam").detach_user_policy(
         UserName=user_name,
         PolicyArn=policy_arn,
     )
--- a/control_panel_api/settings/base.py
+++ b/control_panel_api/settings/base.py
@@ -1,7 +1,9 @@
 import os
 
+import boto3
 
 # SECURITY WARNING: keep the secret key used in production secret!
+
 SECRET_KEY = os.environ.get(
     'SECRET_KEY',
     '(2gbfi1uc1llww251t00s7$^luuzvivf7l+(snj=sbt#s8h!wu')
@@ -15,10 +17,8 @@ DEBUG = os.environ.get('DEBUG', 'False').lower() == 'true'
 ALLOWED_HOSTS = list(
     filter(None, os.environ.get('ALLOWED_HOSTS', '').split(' ')))
 
-
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-
 
 # Application definition
 
@@ -68,7 +68,6 @@ TEMPLATES = [
 
 WSGI_APPLICATION = 'control_panel_api.wsgi.application'
 
-
 # Database
 # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
 
@@ -82,7 +81,6 @@ DATABASES = {
         'PORT': os.environ.get('DB_PORT', '5432'),
     }
 }
-
 
 # Password validation
 # https://docs.djangoproject.com/en/1.11/ref/settings/#auth-password-validators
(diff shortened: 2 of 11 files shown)


jmoz:
If good I will merge after previous PR.

control_panel_api/aws.py
```diff
@@ -1,18 +1,20 @@
 import json
 
-import boto3
+from django.conf import settings
+
+aws_api_client = settings.AWS_API_CLIENT_HANDLER
```
andyhd: PEP8 - 2 lines after imports

control_panel_api/aws.py
```diff
@@ -24,20 +26,20 @@ def put_bucket_logging(name, target_bucket, target_prefix):
...
 def delete_policy(policy_arn):
-    boto3.client("iam").delete_policy(PolicyArn=policy_arn)
+    aws_api_client("iam").delete_policy(PolicyArn=policy_arn)
 
 
 def detach_policy_from_entities(policy_arn):
     """Get all entities to which policy is attached
-    See: http://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.list_entities_for_policy"""
-    entities = boto3.client("iam").list_entities_for_policy(PolicyArn=policy_arn)
+    See: http://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.list_entities_for_policy"""
```
andyhd: This URL is was correct.  Not sure this doc comment is very useful anyway.

control_panel_api/settings/base.py
```diff
@@ -1,7 +1,9 @@
 import os
 
+import boto3
 
 # SECURITY WARNING: keep the secret key used in production secret!
+
```
andyhd: Probably better to keep the comment next to the setting it comments on, but it's not a very useful comment anyway, so just remove it.

control_panel_api/tests/test_aws.py
```diff
@@ -4,7 +4,7 @@
 from control_panel_api import aws
 
 
-@patch('boto3.client')
+@patch('control_panel_api.aws.aws_api_client')
```
andyhd: Do you need to patch the `aws_api_client` if it's set to `MagicMock()`?

control_panel_api/tests/test_services.py
```diff
@@ -67,11 +65,12 @@ def test_bucket_delete(self, mock_delete_bucket_policies):
...
 
-@override_settings(ENV='test', IAM_ARN_BASE='arn:aws:iam::1337')
 class NamingTestCase(SimpleTestCase):
     def test_policy_name_has_readwrite(self):
-        self.assertEqual('bucketname-readonly', services._policy_name('bucketname', readwrite=False))
-        self.assertEqual('bucketname-readwrite', services._policy_name('bucketname', readwrite=True))
+        self.assertEqual('bucketname-readonly',
+                         services._policy_name('bucketname', readwrite=False))
+        self.assertEqual('bucketname-readwrite',
+                         services._policy_name('bucketname', readwrite=True))
```
andyhd: Even though this indentation style is ok by PEP8, it's better to indent at regular levels. Break the line after the open parenthesis and indent 4 spaces instead.

xoen:
Just read the description of this (not the code yet) - Is there a way to use the test settings without us having to pass that explicitly when running the tests?

Basically my concern is that someone may run the tests hitting the real AWS by accident by running `$./manage.py test` (e.g. new developer, someone being tired, editor integration, etc...)

control_panel_api/aws.py
```diff
@@ -1,18 +1,20 @@
...
 
-import boto3
+from django.conf import settings
+
+aws_api_client = settings.AWS_API_CLIENT_HANDLER
 
 
 def create_bucket(name, region, acl='private'):
-    boto3.client('s3').create_bucket(
+    aws_api_client('s3').create_bucket(
```
xoen: You know that interestingly while working on the app IAM policy creation I had a similar idea?

I still wonder if there is a way to globally mock boto3 client to return a mock while testing (again, as a way to protect ourself by "accidentally" use `boto.client()` instead of this).

control_panel_api/aws.py
jmoz: bad pycharm

control_panel_api/aws.py
jmoz: removed

control_panel_api/settings/base.py
jmoz: removed

control_panel_api/settings/test.py
```diff
@@ -0,0 +1,9 @@
+from .base import *
```
xoen: Interesting, so maybe we may have an if in the main `settings.py` to check if the `ENV` is `test` and mock automatically in that case.

So if someone forget to override the settings it would pick that we're in test environment and swap the boto3 client transparently.

xoen: Also, in theory it's possible to stub `boto3.client` calls to return the mock right?

What I'm thinking is, maybe we can do this so that when the tests are run `boto3.client()` would return a mock.

control_panel_api/tests/test_aws.py
jmoz: Yes good shout, can just `import` it

control_panel_api/settings/test.py
xoen: Another thing I'm realising is that we'd ideally want a way to set `ENV="test"` when running `python manage.py test` (don't know if there is a way to plug into it).

control_panel_api/tests/test_services.py
jmoz: This is auto pycharm consistent pep8 formatting

control_panel_api/tests/test_aws.py
xoen: Is `@patch()` basically stubbing the argument to return a `MagicMock` instance behind the scenes?

As said in other comment, ideally if we can do that transparently when running tests, we may well remove this line (or forget by accident...) and any `boto3.client()` call would return the mock instead of a genuine client.

control_panel_api/tests/test_services.py
jmoz: Ps. what do we @xoen @andyhd  think about increasing line length? I find 80 to be quite annoyingly unreadable the way it breaks most lines and restrictive considering modern retina screens/IDEs with auto line wrapping. I prefer 100

https://www.python.org/dev/peps/pep-0008/#maximum-line-length

"Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters"

control_panel_api/aws.py
jmoz: Yeh this is what it's doing!

xoen: Kind of. Here I'm explicitly calling `aws_api_client()` - what happens if a new dev comes in and by accident/distraction calls `boto3.client()`? 

What I'm trying to say is that ideally (if possible!) we'd want our code to not change and still call `boto3.client()` but it would return a mock if running the tests/`ENV` is `test`...it's a subtle difference.

control_panel_api/tests/test_services.py
xoen: Regarding the long lines, I prefer 80 chars as I often use vertical split screen to work on different/related files side-by-side.

I mean, my editor still word-wraps them anyway, so maybe not a big deal, but I think with more than 80 columns the word-wrapping would happen quite often. Also, don't know how that would work on @andyhd 's end...just my 2 cents. See what Andy thinks but we can give it a try if you really want.

control_panel_api/tests/test_aws.py
jmoz: This is what it is doing now. The patch was just there as that was how you inject the ability to assert against the mock, but as Andy pointed out this is now redundant as the client is a mock by default so I can just import it (i'm updating code now).

control_panel_api/aws.py
andyhd: `boto3` isn't imported, so calling `boto3.client` would fail. And we'd probably catch that in a code review?

control_panel_api/settings/test.py
jmoz: I hear what you're saying about the manage test thing.  The usual practice is like this though where a developer can see the settings files are separated and named as test.py this then needs to be specified.  Between us here it's not an issue as I've just notified you all, and a new dev can read a line in the README about running them.

Another thing is we haven't really configured environments yet e.g. prod or local. We could default local to using mock so to explcitly hit aws you have to specify it via `--settings`, we could even create a settings file like `local_aws.py` that connects to it.

So this PR fixes our test issues, but now we have to think about local dev.  Explicitly have to use settings with actual aws?  Also what about a staging/dev aws environment to connect to? @andyhd

andyhd: Perhaps just this: https://stackoverflow.com/a/7651002

control_panel_api/tests/test_services.py
andyhd: 80 chars ☑️

andyhd: This isn't really a line-length thing, though - it's indenting to an arbitrary number of spaces, not necessarily a multiple of 4.

control_panel_api/aws.py
jmoz: Yeh I think reading the code it's clear this is the paradigm, and this is the aws file where it's all located. You're talking about monkey patching it, which then adds a confusing layer of magic, how do they know what they're doing etc.

control_panel_api/settings/test.py
andyhd: The second answer to that stackoverflow question shows how easy it would be to override the TestRunner too.

manage.py
```diff
@@ -3,7 +3,9 @@
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "control_panel_api.settings")
+    settings = 'control_panel_api.settings.test' if 'test' in sys.argv else 'control_panel_api.settings'
```
jmoz: I've seen this kind of manage.py stuff before so fairly happy with it.

andyhd: It's more readable to split this into a regular `if ... else` block

jmoz: Yep

control_panel_api/aws.py
xoen: @andyhd I was talking in general, not specifically in this file. Anyway, if you think this is enough I'm happy with the explicit import of this specific client.

control_panel_api/settings/test.py
xoen: @andyhd good one. I think that's what I had in mind as it greatly reduce the chances of error (which is the point of these changes in the first place).

I think just add a note in the `README` wouldn't cut. A new dev comes in, see it's a django app then would run the test with the standard `python manage.py test`, let's say he already had AWS credentials in place it may _potentially_ hit the real service.

xoen:
Hang on, was this merged? Right...that would explain the conflicts on the other branch 👍

manage.py
xoen: Uh, so we actually don't need to change the way we run the tests. I didn't see this.

PR #23, opened 2017-09-19
Grant app access to S3 bucket

xoen:
## What

When a `AppS3Bucket` record is created for an app and an S3 bucket,
the right IAM policy to access the S3 bucket (depending on this record
`access_level`) is attached to the app IAM role.

### Ticket
Ticket: https://trello.com/c/oR2oS4Af/382-cp-api-when-s3-bucket-access-is-granted-to-an-app-attach-iam-policy

### Related code
This is similar to what already done by one of the lambda functions:

https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/data_access/memberships/memberships.py#L38

### DIFF
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -5,13 +5,13 @@ from django.template.defaultfilters import slugify
 
 from . import aws
 
-READ_WRITE = 'readwrite'
-READ_ONLY = 'readonly'
+READWRITE = 'readwrite'
+READONLY = 'readonly'
 
 
 def _policy_name(bucket_name, readwrite=False):
     """Prefix the policy name with bucket name, postfix with access level e.g. dev-james-readwrite"""
-    return "{}-{}".format(bucket_name, READ_WRITE if readwrite else READ_ONLY)
+    return "{}-{}".format(bucket_name, READWRITE if readwrite else READONLY)
 
 
 def _app_role_name(app_slug):
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -153,6 +153,13 @@ def delete_bucket_policies(bucket_name):
     aws.delete_policy(policy_arn_readonly)
 
 
+def app_granted_access_to_bucket(app_slug, bucket_name, readwrite):
+    aws.attach_policy_to_role(
+        policy_arn=_policy_arn(bucket_name, readwrite),
+        role_name=_app_role_name(app_slug),
+    )
+
+
 def bucket_create(name):
     """Simple facade function for view to call on s3bucket create"""
     create_bucket(name)
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -153,7 +153,11 @@ def delete_bucket_policies(bucket_name):
     aws.delete_policy(policy_arn_readonly)
 
 
-def app_granted_access_to_bucket(app_slug, bucket_name, readwrite):
+def apps3bucket_create(apps3bucket):
+    app_slug = apps3bucket.app.slug
+    bucket_name = apps3bucket.s3bucket.name
+    readwrite = (apps3bucket.access_level == READWRITE)
+
     aws.attach_policy_to_role(
         policy_arn=_policy_arn(bucket_name, readwrite),
         role_name=_app_role_name(app_slug),
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -154,13 +154,14 @@ def delete_bucket_policies(bucket_name):
 
 
 def apps3bucket_create(apps3bucket):
-    app_slug = apps3bucket.app.slug
-    bucket_name = apps3bucket.s3bucket.name
-    readwrite = (apps3bucket.access_level == READWRITE)
+    policy_arn = _policy_arn(
+        apps3bucket.s3bucket.name,
+        apps3bucket.access_level == READWRITE,
+    )
 
     aws.attach_policy_to_role(
-        policy_arn=_policy_arn(bucket_name, readwrite),
-        role_name=_app_role_name(app_slug),
+        policy_arn=policy_arn,
+        role_name=_app_role_name(apps3bucket.app.slug),
     )
 
 
--- a/control_panel_api/tests/test_services.py
+++ b/control_panel_api/tests/test_services.py
@@ -1,9 +1,14 @@
 from unittest.mock import patch, call
 
 from django.conf import settings
-from django.test.testcases import SimpleTestCase
+from django.test.testcases import SimpleTestCase, TransactionTestCase
 
 from control_panel_api import services
+from control_panel_api.models import (
+    App,
+    AppS3Bucket,
+    S3Bucket,
+)
 from control_panel_api.tests import (
     APP_IAM_ROLE_ASSUME_POLICY,
     POLICY_DOCUMENT_READONLY,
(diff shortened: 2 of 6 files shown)


control_panel_api/services.py
```diff
@@ -153,6 +153,13 @@ def delete_bucket_policies(bucket_name):
     aws.delete_policy(policy_arn_readonly)
 
 
+def app_granted_access_to_bucket(app_slug, bucket_name, readwrite):
```
jmoz: Following the convention from other service calls that are called by the view, this should be named `service.apps3bucket_create`

control_panel_api/views.py
```diff
@@ -58,6 +58,15 @@ class AppS3BucketViewSet(viewsets.ModelViewSet):
     queryset = AppS3Bucket.objects.all()
     serializer_class = AppS3BucketSerializer
 
+    def perform_create(self, serializer):
+        app_s3bucket = serializer.save()
+
+        services.app_granted_access_to_bucket(
```
jmoz: I think we should call the service method here and pass an instance, not many parameters, and encapsulate the readwrite logic inside the service layer - view stays clean - could look like this:

```
def perform_create(self, serializer):
    app_s3bucket = serializer.save()
    services.app_s3bucket_create(app_s3bucket)
```

Could also refactor existing view/service and use OO properly and just pass the instance and keep the views clean.

control_panel_api/services.py
xoen: I'm not really sure about this naming convention as the service layer names should be driven by the domain problem in my mind - Not by whatever we called the models (especially in the case of this one which it's not exactly great).

This `services` module is starting to look more and more like `callbacks` rather than a service layer with a set of high level operations you can call.

Rename this function is obviously trivial but we should probably have a quick chat about it.

control_panel_api/views.py
xoen: I tried to do exactly that in another of the services function but had troubles with testing (it was on deletion of something so the `assert_called_with()` was confused by the missing id on the argument or something like that)...

...in the end I decided to keep that function consistent with the rest. Again this is more general question, why this function receive a model instance but most of the others receive `app_slug`, `bucket_name`, etc...? Should we change all of them? Is 3 an high number of arguments?

control_panel_api/services.py
xoen: Even if I think long term we should try to keep the service layer more high-level, I'm changing it to `app_s3bucket` (as it's the relation between apps and S3 buckets) 👍

control_panel_api/views.py
xoen: I'll change it 👍

xoen: FYI: Another thing to bear in mind here is that we're coupling the `services` layer with the models which is not strictly necessary.

e.g. see the tests of this function now, I'm just calling it passing some strings, now that I'm changing it to receive a model instance I'll have to create a `AppS3Bucket` record just to test the function to grant access to an app. This makes me wonder about the design of it.

xoen:
@jmoz I renamed the function as requested: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/23/commits/532962bf3eb38b9636990ebbcd86a9cdcab86eeb

Also rename constants in `services` module to keep them consistent with what we already have in the  `AppS3Bucket` model: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/23/commits/02315e51812ea74d4b7def9a9130247dd18a61a7

Let me know if this is not good to go 👍

control_panel_api/services.py
```diff
@@ -153,6 +153,17 @@ def delete_bucket_policies(bucket_name):
     aws.delete_policy(policy_arn_readonly)
 
 
+def apps3bucket_create(apps3bucket):
+    app_slug = apps3bucket.app.slug
+    bucket_name = apps3bucket.s3bucket.name
+    readwrite = (apps3bucket.access_level == READWRITE)
```
andyhd: This is a code smell - the function requires a lot of knowledge of the internals of the apps3bucket object - this should probably be a method on the class.

control_panel_api/tests/test_services.py
```diff
@@ -92,6 +97,28 @@ def test_app_delete(self, mock_delete_role):
 
         mock_delete_role.assert_called_with(expected_role_name)
 
+    @patch('control_panel_api.aws.attach_policy_to_role')
+    def test_apps3bucket_create(self, mock_attach_policy_to_role):
+        app = App.objects.create(slug = 'appslug')
```
andyhd: PEP8 - no spaces around `=` in keyword arguments

control_panel_api/services.py
xoen: Originally this function was receiving these arguments as a string.

I know this is usually a symptom of methods being in the wrong place.
At the same time, don't we risk to overload models with a lot of methods? In the past I've worked on some applications in which we had models for basic crud stuff, and service objects to add a layer of context-aware operations - As in methods used when performing certain operations in certain contexts, or maybe even validations only happening in certain context, e.g. admin interface could have different validations or something like that.

I feel like this comment is a more general one on the general architecture and structure of the app rather than something specific of this feature changes itself: Should we "ignore" this for the specific feature and have an offline discussion with @jmoz about of we want to structure things and refactor existing stuff?

control_panel_api/tests/test_services.py
xoen: 😱 🙀 😱 🙀 😱 🙀 😱 🙀 😱 🙀

xoen: Changing it 👍 (also interesting that my editor PEP8 thing didn't change it!)

xoen:
FYI: PEP8 change: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/23/commits/ebb1af29bf55b5931532a975e153258998cf2579

control_panel_api/services.py
xoen: @andyhd and @jmoz to recap what just discussed offline:
- We're going to refactor and move these functions in classes.
- We'll merge these PRs and we'll move them as a separate step/PRs
- We'll look at the functions and see if there is some natural grouping which would lead to ideas on where to put this code (potentially start by moving them in the models and maybe further down the line move them in "proper" service classes)

jmoz: Parentheses not needed

xoen: Mmmmm...but I like parenthesis.

Alright. I'll remove them then.

control_panel_api/services.py
```diff
@@ -153,6 +153,17 @@ def delete_bucket_policies(bucket_name):
     aws.delete_policy(policy_arn_readonly)
 
 
+def apps3bucket_create(apps3bucket):
+    app_slug = apps3bucket.app.slug
```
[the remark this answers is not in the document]
xoen: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/23/commits/a0d0bef0383bc7947237a10c6989482ee252f302

control_panel_api/services.py
xoen: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/23/commits/a0d0bef0383bc7947237a10c6989482ee252f302

xoen:
@jmoz BAAAAAM

xoen:
PS: @jmoz let me know if you have any troubles rebasing after the merge. Also, let me know if you want to have a look at moving these functions from `services` to the models later 👍

PR #24, opened 2017-09-20
Delete role and policy on apps3bucket deletion

jmoz:
## What

When you delete the apps3bucket relation it should also detach policy and role.

(Note: apologies on line length reformatting - I updated to 80 char after team decided on that length)

## How to review

1. Run tests.
2. Call it live and check.

### DIFF
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -30,7 +30,7 @@ def bucket_arn(bucket_name):
     return "arn:aws:s3:::{}".format(bucket_name)
 
 
-def app_slug(name):
+def app_slugify(name):
     """Create a valid slug for s3 using standard django slugify but we add some custom"""
     return re.sub(r'_+', '-', slugify(name))
 
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -15,7 +15,6 @@ def _policy_name(bucket_name, readwrite=False):
 
 
 def _app_role_name(app_slug):
-    """Return the IAM role name for the given `app_slug`"""
     return "{}_app_{}".format(settings.ENV, app_slug)
 
 
@@ -26,7 +25,6 @@ def _policy_arn(bucket_name, readwrite=False):
 
 
 def bucket_arn(bucket_name):
-    """Return bucket arn e.g. arn:aws:s3:::bucketname"""
     return "arn:aws:s3:::{}".format(bucket_name)
 
 
@@ -36,15 +34,10 @@ def app_slugify(name):
 
 
 def app_create(app_slug):
-    """Creates the IAM role for the given app"""
     _create_app_role(_app_role_name(app_slug))
 
 
 def _create_app_role(role_name):
-    """Creates the IAM role for the given app"""
-
-    # See: `sts:AssumeRole` required by kube2iam
-    # https://github.com/jtblin/kube2iam#iam-roles
     assume_role_policy = {
         "Version": "2012-10-17",
         "Statement": [
@@ -69,12 +62,10 @@ def _create_app_role(role_name):
 
 
 def app_delete(app_slug):
-    """Deletes the IAM role for the given app"""
     aws.delete_role(_app_role_name(app_slug))
 
 
 def get_policy_document(bucket_name, readwrite):
-    """Return a standard policy document for a bucket, use a boolean flag `readwrite` to add write access"""
     bucket_name_arn = bucket_arn(bucket_name)
 
     statements = [
@@ -130,7 +121,6 @@ def get_policy_document(bucket_name, readwrite):
 
 
 def create_bucket(bucket_name):
-    """Create an s3 bucket and add logging"""
     aws.create_bucket(
         bucket_name, region=settings.BUCKET_REGION, acl='private')
     aws.put_bucket_logging(bucket_name, target_bucket=settings.LOGS_BUCKET_NAME,
@@ -167,14 +157,12 @@ def detach_bucket_access_from_app_role(app_slug, bucket_name, access_level):
 
 
 def bucket_create(name):
-    """Simple facade function for view to call on s3bucket create"""
     create_bucket(name)
     create_bucket_policies(name)
 
 
 def bucket_delete(name):
-    """Simple facade function for view to call on s3bucket delete
-    Note we do not destroy the actual data, just the policies"""
+    """Note we do not destroy the actual data, just the policies"""
     delete_bucket_policies(name)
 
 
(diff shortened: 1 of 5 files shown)


control_panel_api/views.py
```diff
@@ -58,6 +58,10 @@ class AppS3BucketViewSet(viewsets.ModelViewSet):
     queryset = AppS3Bucket.objects.all()
     serializer_class = AppS3BucketSerializer
 
+    def perform_destroy(self, instance):
```
andyhd: why do all these method names start with `perform_`? isn't that redundant?

control_panel_api/services.py
```diff
@@ -163,3 +166,16 @@ def bucket_delete(name):
     """Simple facade function for view to call on s3bucket delete
     Note we do not destroy the actual data, just the policies"""
     delete_bucket_policies(name)
+
+
+def apps3bucket_delete(apps3bucket):
+    """On delete we detach the corresponding s3 read/write policy from app role
```
andyhd: This comment only describes _what_ the function does, it doesn't explain _why_

control_panel_api/services.py
```diff
@@ -163,3 +166,16 @@ def bucket_delete(name):
     """Simple facade function for view to call on s3bucket delete
     Note we do not destroy the actual data, just the policies"""
     delete_bucket_policies(name)
+
+
+def apps3bucket_delete(apps3bucket):
```
andyhd: Shouldn't this function be a method on `AppS3Bucket`?

control_panel_api/services.py
```diff
@@ -138,8 +139,10 @@ def create_bucket(bucket_name):
 
 def create_bucket_policies(bucket_name):
     """Create readwrite and readonly policies for s3 bucket"""
-    aws.create_policy(_policy_name(bucket_name, readwrite=True), get_policy_document(bucket_name, True))
-    aws.create_policy(_policy_name(bucket_name, readwrite=False), get_policy_document(bucket_name, False))
+    aws.create_policy(_policy_name(bucket_name, readwrite=True),
+                      get_policy_document(bucket_name, True))
+    aws.create_policy(_policy_name(bucket_name, readwrite=False),
+                      get_policy_document(bucket_name, False))
```
andyhd: Please avoid arbitrary indentation levels. Break the line after the open parenthesis and indent 4 spaces.
```python
aws.create_policy(
    _policy_name(...
```

control_panel_api/tests/test_views.py
```diff
@@ -177,6 +177,15 @@ def test_delete(self):
...
         self.assertEqual(HTTP_404_NOT_FOUND, response.status_code)
 
+    @patch('control_panel_api.services.apps3bucket_delete')
+    def test_delete_calls_api(self, mock_apps3bucket_delete):
+        response = self.client.delete(
+            reverse('apps3bucket-detail', (self.apps3bucket_1.id,)))
+        self.assertEqual(HTTP_204_NO_CONTENT, response.status_code)
+
+        called_args, _ = mock_apps3bucket_delete.call_args
+        self.assertIsInstance(called_args[0], AppS3Bucket)
```
xoen: As discussed offline earlier I would personally be more specific here and check the function was called with the right instance, e.g. 

```python
called_args, _ = mock_apps3bucket_delete.call_args
self.assertEqual(called_args[0].app, self.apps3bucket_1.app)
self.assertEqual(called_args[0].s3bucket, self.apps3bucket_1.s3bucket)
```

...or something like that, you got the idea. But that's just me.

control_panel_api/views.py
xoen: This is not exactly apparent and I may be wrong but I think here we're overriding a method coming from Django Rest Framework (the beauty of inheritance 💃 )

control_panel_api/services.py
jmoz: This follows pep8, and is auto formatted, see first example

https://www.python.org/dev/peps/pep-0008/#indentation

Do we have a custom code style to follow I can reference?

control_panel_api/tests/test_views.py
jmoz: Yep done.

control_panel_api/services.py
jmoz: Yep, also removed other ones.

control_panel_api/views.py
jmoz: @andyhd what @xoen said!

control_panel_api/services.py
xoen: This is what PEP8 says about indentation: https://www.python.org/dev/peps/pep-0008/#indentation

I tend to write stuff like this as:

```
aws.create_policy(
    _policy_name(bucket_name, readwrite=True),
    get_policy_document(bucket_name, True),
)
aws.create_policy(
    _policy_name(bucket_name, readwrite=False),
    get_policy_document(bucket_name, False),
)
```

which is valid from a PEP8-perspective and it's visually clear what's an argument of the function, etc...BUT, you two are the python ninjas.

control_panel_api/tests/test_views.py
xoen: 🌟

control_panel_api/services.py
```diff
@@ -15,35 +15,29 @@ def _policy_name(bucket_name, readwrite=False):
...
 
 def app_create(app_slug):
-    """Creates the IAM role for the given app"""
     _create_app_role(_app_role_name(app_slug))
 
 
 def _create_app_role(role_name):
-    """Creates the IAM role for the given app"""
-
-    # See: `sts:AssumeRole` required by kube2iam
```
xoen: I think that comment was actually useful (it certainly saved me some research time in the lambda function version the other day!)

xoen:
You may need to rebase (let me know if you need any input from me to resolve possible conflicts) but 👍

control_panel_api/services.py
jmoz: Added back

control_panel_api/services.py
andyhd: It is acceptable to PEP8, but I think PEP8 is wrong on this one.

jmoz: Cool.  I would rather follow a strict set of rules than having to debate _personal opinion and preference_.

However I happen to agree on this preference which is also valid pep8 @xoen @andyhd so I'll try and hack my formatter to use hanging!

xoen: What if PEP8 was wrong

PR #25, opened 2017-09-25
Update attached app policies when `AppS3Bucket` access level changes

xoen:
## What

When the AppS3Bucket record is updated the policies attached to the
policies changes accordingly:
* If new `access_level` is `readwrite`, old one would be `readonly` (`readonly` => `readwrite`):
  * so the readonly policy would be detached from the app role and the readwrite policy would be attached 
* Conversely, when new `access_level` is `readonly` (`readwrite` => `readonly`):
  * the readonly policy would be attached and the readwrite policy detached 

**NOTE**: When talking about policy here I mean IAM policy to access the S3 bucket in the `AppS3Bucket` record.

## Ticket
https://trello.com/c/D3y3Mzre/388-cp-api-when-app-access-level-to-s3-bucket-changes-iam-policies-attached-to-its-role-should-change-too

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -112,3 +112,9 @@ class AppS3Bucket(TimeStampedModel):
     class Meta:
         # one record per app/s3bucket
         unique_together = ('app', 's3bucket')
+
+    def has_readwrite_access(self):
+        return self.access_level == self.READWRITE
+
+    def update_aws_permissions(self):
+        services.apps3bucket_update(self)
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -173,7 +173,7 @@ def bucket_delete(name):
 def apps3bucket_create(apps3bucket):
     policy_arn = _policy_arn(
         apps3bucket.s3bucket.name,
-        apps3bucket.access_level == READWRITE,
+        apps3bucket.has_readwrite_access(),
     )
 
     aws.attach_policy_to_role(
@@ -181,6 +181,27 @@ def apps3bucket_create(apps3bucket):
         role_name=_app_role_name(apps3bucket.app.slug),
     )
 
+
+def apps3bucket_update(apps3bucket):
+    app_role_name = _app_role_name(apps3bucket.app.slug)
+
+    new_policy_arn = _policy_arn(
+        apps3bucket.s3bucket.name,
+        apps3bucket.has_readwrite_access(),
+    )
+    old_policy_arn = _policy_arn(
+        apps3bucket.s3bucket.name,
+        not apps3bucket.has_readwrite_access(),
+    )
+
+    aws.attach_policy_to_role(
+        policy_arn=new_policy_arn,
+        role_name=app_role_name,
+    )
+    aws.detach_policy_from_role(
+        policy_arn=old_policy_arn,
+        role_name=app_role_name,
+    )
 
 def apps3bucket_delete(apps3bucket):
     """:type apps3bucket: control_panel_api.models.AppS3Bucket"""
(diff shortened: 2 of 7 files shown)


control_panel_api/services.py
```diff
@@ -182,6 +182,27 @@ def apps3bucket_create(apps3bucket):
...
+    new_policy_arn = _policy_arn(
+        apps3bucket.s3bucket.name,
+        apps3bucket.has_readwrite_access(),
+    )
+    old_policy_arn = _policy_arn(
+        apps3bucket.s3bucket.name,
+        not apps3bucket.has_readwrite_access(),
+    )
+
+    aws.attach_policy_to_role(
```
jmoz: I would have thought detach needs to be called first?

xoen: Order doesn't make much difference as long as in the end the "old" policy is detached and the "new" one is attached. The end result would be the same.

xoen:
(rebased off `api` branch)

control_panel_api/models.py
```diff
@@ -112,3 +112,9 @@ class AppS3Bucket(TimeStampedModel):
     class Meta:
         # one record per app/s3bucket
         unique_together = ('app', 's3bucket')
+
+    def has_readwrite_access(self):
+        return self.access_level == self.READWRITE
+
+    def update_aws_permissions(self):
+        services.apps3bucket_update(self)
```
jmoz: So here, this ties with the convo we had with Andy other day about the model coupling in service library.  The goal of my refactoring is to remove the model from the services and just let the service functions accept parameters (could then be used elsewhere in commands and reduces coupling).  So this method should not pass `self`, but pass its attributes as parameters to `apps3bucket_update`.  That method would require a slight rewrite to accept params.

And the function name now change from `apps3bucket_update ` to something like `aws_update`, but we still need to decide on them.

xoen:
@jmoz Yes, as I was aware of you're refactoring work I avoided moving too many things around to prevent more conflicts than the necessary. 

I'll support you with any issues rebasing and we can further rebase on your branch 👍

xoen:
@andyhd FYI: This should work, we'll further refactor once merged as part of James' refactoring.
Let me know if there is anything broken here, we can address style things as part of the refactor.

control_panel_api/tests/test_views.py
```diff
@@ -224,6 +225,37 @@ def test_update(self):
...
+
+    def test_update_when_s3bucket_changed(self):
+        data = {
+            'app': self.app_1.id,
+            's3bucket': self.s3_bucket_2.id,
+            'access_level': AppS3Bucket.READWRITE,
+        }
+        response = self.client.put(
+            reverse('apps3bucket-detail', (self.apps3bucket_1.id,)), data)
+        self.assertEqual(HTTP_400_BAD_REQUEST, response.status_code)
```
andyhd: These last two tests look very similar, can they be data driven? Like parametrized tests with pytest? Eg:
```python
def test_updates(self):
    cases = [
        {'app': self.app_2.id, 's3bucket': self.s3_bucket_1.id, ...},
        {'app': self.app_1.id, 's3bucket': self.s3_bucket_2.id, ...}
    ]
    for case in cases:
        self.assert_update_fails(self.apps3bucket_1, case)
```

andyhd: These tests have some common lines of code, which could be made clearer with a custom assert method, eg:
```python
def test_update_when_s3bucket_changed(self):
    data = {...}
    self.assert_update_fails(self.apps3bucket_1, data)
```

xoen:
Thanks @andyhd - I'll merge this and we'll have a look as part of the refactoring work on the other branch 👍

PR #26, opened 2017-09-26
Feature refactor model service

jmoz:
## What

Goal of this PR was to extract some of the ORM code from the service layer to reduce coupling.

Views are clean and call a method on model.

Model is clean and calls `service` functions passing its instance attributes as params.  Service does not need to know about django model.  Service function could be called by a command by just passing params.

Service layer does a bit of orchestrating and calls lower level aws wrapper functions.

Each layer is tested with each other.

## Issues

- Naming of model methods.
- Refactor out policy name and arn etc.
- Now we have django models with methods hitting aws, is this ok? Do we want to further pull out this to different classes to clean up and make clearer?

## How to review

1. Check design.
2. Run tests.

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -159,8 +159,26 @@ class AppS3Bucket(AccessToS3Bucket):
         # one record per app/s3bucket
         unique_together = ('app', 's3bucket')
 
+    def aws_create(self):
+        services.attach_bucket_access_to_app_role(
+            self.s3bucket.name,
+            self.has_readwrite_access(),
+            self.app.role_name
+        )
+
+    def aws_delete(self):
+        services.detach_bucket_access_from_app_role(
+            self.s3bucket.name,
+            self.has_readwrite_access(),
+            self.app.role_name
+        )
+
     def update_aws_permissions(self):
-        services.apps3bucket_update(self)
+        services.apps3bucket_update(
+            self.s3bucket.name,
+            self.has_readwrite_access(),
+            self.app.role_name
+        )
 
 
 class UserS3Bucket(AccessToS3Bucket):
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -50,9 +50,6 @@ class App(TimeStampedModel):
 
 
 class S3Bucket(TimeStampedModel):
-    READWRITE = 'readwrite'
-    READONLY = 'readonly'
-
     name = models.CharField(unique=True, max_length=63, validators=[
         validators.validate_env_prefix,
         validators.validate_s3_bucket_length,
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -1,5 +1,9 @@
+import re
+
+from django.conf import settings
 from django.contrib.auth.models import AbstractUser
 from django.db import models
+from django.template.defaultfilters import slugify
 from django_extensions.db.fields import AutoSlugField
 from django_extensions.db.models import TimeStampedModel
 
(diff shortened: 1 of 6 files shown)


control_panel_api/models.py
```diff
@@ -22,15 +26,33 @@ def get_short_name(self):
 
 
 class App(TimeStampedModel):
+    def _slugify(name):
```
xoen: Should this be an instance method? e.g. pass `self` and replace name with `self.name`:

```python
    def _slugify(self):
        return re.sub(r'_+', '-', slugify(self.name))
```

xoen: Oh hang on...it's because the signature of this function needs to be like that to work with `slugify_function`, sorry, didn't realise that.

Mmmm...wonder if we should move it up in the module (don't know if it's reusable in other models), but don't worry about that 👍

control_panel_api/models.py
```diff
@@ -22,15 +26,33 @@ def get_short_name(self):
...
+
+    def aws_create_role(self):
+        services.create_app_role(self.aws_role_name)
+
+    def aws_delete_role(self):
+        services.delete_app_role(self.aws_role_name)
+
 
 class S3Bucket(TimeStampedModel):
+    READWRITE = 'readwrite'
```
xoen: Two things:
1. This may be misleading here, S3 buckets are not RO/RW per se
2. It doesn't look like we actually use these constants anywhere?

control_panel_api/models.py
```diff
@@ -116,5 +146,23 @@ class Meta:
     def has_readwrite_access(self):
         return self.access_level == self.READWRITE
 
-    def update_aws_permissions(self):
-        services.apps3bucket_update(self)
+    def aws_create(self):
+        services.attach_bucket_access_to_app_role(
```
xoen: (thinking aloud, **ignore me**): Wouldn't be nicer if these functions were called:
- `attach_bucket_access_to_app_role` => `grant_bucket_access_to_app`
- `detach_bucket_access_from_app_role` => `revoke_bucket_access_from_app`

In the future it may be nice to have an higher level class (`AWSAppS3Bucket`?), something like:

```python
AWSAppS3Bucket.from(apps3bucket).grant_access()
AWSAppS3Bucket.from(apps3bucket).update_access()
AWSAppS3Bucket.from(apps3bucket).revoke_access()

AWSS3Bucket.from(s3bucket).create()
AWSS3Bucket.from(s3bucket).delete()

AWSApp.from(app).create()
AWSApp.from(app).delete()
```

control_panel_api/services.py
```diff
@@ -14,30 +11,13 @@ def _policy_name(bucket_name, readwrite=False):
     return "{}-{}".format(bucket_name, READWRITE if readwrite else READONLY)
 
 
-def _app_role_name(app_slug):
-    return "{}_app_{}".format(settings.ENV, app_slug)
-
-
 def _policy_arn(bucket_name, readwrite=False):
```
xoen: Potentially we may have `aws_policy_arn()` in the `S3Bucket` model.
This, IMHO, is very related to an S3Bucket as this is the ARN for policies who determine the access to it.

xoen: Mmmm, but with the limitation of not being able to pass models instances to functions in `service` I guess it's not possible.

Maybe when we'll have `AWS*` "models" which incapsulates models this would not be a problem.

control_panel_api/services.py
```diff
@@ -148,50 +133,39 @@ def delete_bucket_policies(bucket_name):
...
-    create_bucket_policies(name)
-
-
-def bucket_delete(name):
-    """Note we do not destroy the actual data, just the policies"""
-    delete_bucket_policies(name)
-
-
-def apps3bucket_create(apps3bucket):
+def attach_bucket_access_to_app_role(bucket_name, readwrite, app_role_name):
```
xoen: Again, feels like these functions' names could be higher level and contains the words `grant`/`revoke` in them. 

But for now don't worry if it's too annoying to rename. The more I think about it, the more I like the idea of having `AWS*` "models" which performs these operations and have access to the models' instances.

xoen:
Few comments but looks good, it feels better now 👍 

(still think it would be cleaner to move the AWS-related logic in `AWS*` "models", e.g. `AppAWS`, `S3BucketAWS`, `AppS3BucketAWS`, etc...but that's for another day :)

control_panel_api/models.py
xoen: I'm glad I have your complete approval on this @andyhd 🤣

xoen: Shame `from` is a reserved keyword in Python 🤓, as I said, I was pretty much just thinking aloud. And I'm not suggesting to change this **now**.

At the same time, I still think there we're missing a trick here, some kind of entity which:
 - knows db models
 - knows how to update AWS to reflect the state of things in the CP-API DB

Whether this makes sense, how to call these things, how to call their methods I don't know yet.

jmoz: So right now we are at least HONEST.

control_panel_api/services.py
jmoz: Yeh this area needs another look.

control_panel_api/models.py
jmoz: @andyhd I've added a new ticket for this https://trello.com/c/vV7GOAPw/414-fix-slugify-function

control_panel_api/models.py
jmoz: Well spotted - was old code from when I was trying something else.

PR #27, opened 2017-09-28
Add the created_by User to App and S3Bucket on create

jmoz:
## What

Adds a created_by field to S3Bucket and App which will save the user who created the record.

## How to review

1. ./manage.py migrate
2. ./manage.py test


control_panel_api/migrations/0011_auto_20170927_1449.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('control_panel_api', '0010_auto_20170906_1501'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='app',
+            name='created_by',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL),
```
xoen: Is it `CASCADE` the best strategy here?

It would feel a bit harsh/unpredictable to delete an app/s3bucket just because the user who created it was deleted?

Should we use the `SET_NULL` maybe? The annoyance would be that *potentially* `created_by` *may be* `None`/`NULL` so we'll have to guard against that when we'll start to use it.

control_panel_api/models.py
```diff
@@ -34,6 +34,7 @@ def _slugify(name):
     name = models.CharField(max_length=100, blank=False)
     slug = AutoSlugField(populate_from='name', slugify_function=_slugify)
     repo_url = models.URLField(max_length=512, blank=True, default='')
+    created_by = models.ForeignKey(User, null=True)
```
xoen: See comment on `on_delete.CASCADE` strategy.

control_panel_api/models.py
```diff
@@ -58,6 +59,7 @@ class S3Bucket(TimeStampedModel):
         validators.validate_s3_bucket_length,
         validators.validate_s3_bucket_labels,
     ])
+    created_by = models.ForeignKey(User, null=True)
```
xoen: See comment on `on_delete.CASCADE` strategy.

control_panel_api/migrations/0011_auto_20170927_1449.py
jmoz: Good idea.

xoen:
Looks good. Only concern is the `on_delete.CASCADE` strategy which would mean delete a user may propagate and delete the apps/S3 buckets he created.

xoen:
I'm happy with this now 👍

control_panel_api/migrations/0011_auto_20170927_1449.py
andyhd: I think this highlights the need to mark things (users) as deleted, rather than actually deleting them.

control_panel_api/serializers.py
```diff
@@ -24,10 +24,12 @@ class Meta:
 
 
 class AppSerializer(serializers.ModelSerializer):
-
     class Meta:
         model = App
-        fields = ('id', 'url', 'name', 'slug', 'repo_url', 'apps3buckets')
+        fields = (
+            'id', 'url', 'name', 'slug', 'repo_url', 'apps3buckets',
```
andyhd: Minor formatting point - it's preferable to have one item per line, so that when the list changes, diffs are clearer.

control_panel_api/migrations/0011_auto_20170927_1449.py
xoen: @andyhd yes, I think long term (or also medium-term :D) that's probably the best approach 👍

control_panel_api/serializers.py
xoen: I tend to prefer one per line for "long" lists as well. Should we have a "rule" 🚔  for if something has more than 3 items split it straight away? (with trailing comma on the last element :trollface: )

jmoz: Updated for the diff.

PR #30, opened 2017-09-29
User create role in aws

jmoz:
## What

When a User is created and deleted we create a role and delete it on aws.

Also refactored services.delete_app_role as functionality was the same.

## How to review

1. ./manage.py test
2. Run app and test manually (I did and it works).


xoen:
I'm having a look at this @jmoz - It would be probably worth start merging some of the other PRs and rebase off `api` before merging this, as the other PRs are probably more complicated/annoying to rebase.

control_panel_api/models.py
```diff
@@ -24,6 +24,16 @@ def get_full_name(self):
     def get_short_name(self):
         return self.name
 
+    @property
+    def aws_role_name(self):
+        return f"{settings.ENV}_user_{self.username.lower()}"
+
+    def aws_create_role(self):
+        services.create_user_role(self.aws_role_name)
```
xoen: Why this is called `create_user_role()` but not `create_role()`?
I would imagine because create the IAM role for a user and create the IAM role for the app involves slightly different things.

My point would be, I wonder if we should also keep `services.delete_user_role()`/`services.delete_app_role()`. I see why you renamed it to `services.delete_role()` to remove the duplication, still one part of me thinks they're conceptually two different things.

BUT, this is just food for thought, if you prefer to have `services.delete_role()` for the time being, I wouldn't mind. I'll have a look at the rest in case I see any reasons to not go this way.

control_panel_api/settings/base.py
```diff
@@ -124,5 +124,6 @@
 LOGS_BUCKET_NAME = os.environ.get('LOGS_BUCKET_NAME', 'moj-analytics-s3-logs')
 IAM_ARN_BASE = os.environ.get('IAM_ARN_BASE', '')
 K8S_WORKER_ROLE_NAME = os.environ.get('K8S_WORKER_ROLE_NAME', '')
-
+SAML_PROVIDER_ARN = os.environ.get('SAML_PROVIDER_ARN', '')
+K8S_WORKER_ROLE_ARN = os.environ.get('K8S_WORKER_ROLE_ARN', '')
```
xoen: This is not required as we have both `IAM_ARN_BASE` and `K8S_WORKER_ROLE_NAME` here already.

[See existing `services.create_app_role()` implementation](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/api/control_panel_api/services.py#L56).

control_panel_api/settings/base.py
```diff
@@ -124,5 +124,6 @@
 LOGS_BUCKET_NAME = os.environ.get('LOGS_BUCKET_NAME', 'moj-analytics-s3-logs')
 IAM_ARN_BASE = os.environ.get('IAM_ARN_BASE', '')
 K8S_WORKER_ROLE_NAME = os.environ.get('K8S_WORKER_ROLE_NAME', '')
-
+SAML_PROVIDER_ARN = os.environ.get('SAML_PROVIDER_ARN', '')
```
xoen: Don't forget to update [`cpanel` helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/tree/master/charts/cpanel) and `cpanel`'s [dev](https://github.com/ministryofjustice/analytics-platform-config/blob/master/chart-env-config/dev/cpanel.yml) and [alpha](https://github.com/ministryofjustice/analytics-platform-config/blob/master/chart-env-config/alpha/cpanel.yml) configs or this would break the app when deployed.

(Ask me or @andyhd, and probably Kerin when you get to that point)

control_panel_api/models.py
jmoz: Because as per the method you created `create_app_role` this method encapsulates the policy document.

control_panel_api/settings/test.py
```diff
@@ -8,3 +8,5 @@
 LOGS_BUCKET_NAME = 'moj-test-logs'
 IAM_ARN_BASE = 'arn:aws:iam::123'
 K8S_WORKER_ROLE_NAME = 'test-k8s-worker-role'
+SAML_PROVIDER_ARN = 'arn:aws:iam::12:saml-provider/test'
+K8S_WORKER_ROLE_ARN = 'arn:aws:iam::123:role/test'
```
xoen: As said in other comment, this setting is not necessary, we're generating it by using `IAM_ARN_BASE` and `K8S_WORKER_ROLE_NAME` (this will probably be clear when you'll refactor `create_app_role()` and `create_user_role()`.

control_panel_api/models.py
jmoz: @xoen yes I'm going to merge them in the correct order.

control_panel_api/tests/test_models.py
```diff
@@ -15,8 +15,33 @@
...
+class UserTestCase(TestCase):
+    @patch('control_panel_api.services.create_user_role')
+    def test_aws_create_role_calls_service(self, mock_create_user_role):
+        username = 'james'
+        user = User.objects.create(username=username)
+        user.aws_create_role()
+
+        expected_role_name = f"test_user_{username}"
+
+        mock_create_user_role.assert_called_with(
```
xoen: Minor annoying note, I may be wrong here but we can probably now squeeze these `assert_called_with` params in a single line without breaking the our holy 80-chars rule 🚓 :D

control_panel_api/services.py
```diff
@@ -206,7 +206,7 @@ def create_user_role(role_name):
             {
                 "Effect": "Allow",
                 "Principal": {
-                    "AWS": settings.K8S_WORKER_ROLE_ARN
+                    "AWS": f"{settings.IAM_ARN_BASE}:role/{settings.K8S_WORKER_ROLE_NAME}",
```
xoen: 🌟

control_panel_api/tests/test_models.py
jmoz: More than happy to oblige!

control_panel_api/services.py
```diff
@@ -176,3 +176,41 @@ def apps3bucket_update(bucket_name, readwrite, app_role_name):
         policy_arn=old_policy_arn,
         role_name=app_role_name,
     )
+
+
+def create_user_role(role_name):
```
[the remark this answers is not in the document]
jmoz: Here I'm happy that each one has its own function and the policy document is contained within.  I don't think it needs to be refactored into a boolean param.

Only reason I did with other function is that the implementation was literally the same.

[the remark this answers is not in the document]
xoen: I respectfully disagree on this :)

They're both creating a role with pretty much the same assume role policy document.
Only minor difference is that statement, rest is *exactly* the same.

Happy to discuss this and being convinced of the contrary 🦊 .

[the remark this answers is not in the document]
jmoz: Oh I just looked properly and can see the rest of it is the same.  

Will we be creating a role for anything else other than User and App?  If not I can refactor.  If so then multiple flags isn't good.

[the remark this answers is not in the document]
xoen: I don't predict adding more things' IAM roles at this stage, but we'll cross that bridge when we'll reach it :)

Factor this duplication is probably the best right now, but as you point out if in a year with end up adding more kind of roles, we may need to revisit this.

PR #31, opened 2017-10-04
UserS3Bucket aws calls

jmoz:
## What

Add aws create/delete/update role and policy for User and S3Bucket. 

Also refactored tests.

## How to review

1. ./manage.py test
2. Run api and check aws for user role and s3 policy.


control_panel_api/models.py
```diff
@@ -206,3 +207,25 @@ class UserS3Bucket(AccessToS3Bucket):
...
+            self.s3bucket.name,
+            self.has_readwrite_access(),
+            self.user.aws_role_name,
+        )
+
+    def aws_update(self):
+        services.update_bucket_access(
+            self.s3bucket.name,
+            self.has_readwrite_access(),
+            self.user.aws_role_name,
```
xoen: Just an observation: Now seems like the only difference between these methods in `AppS3Bucket` and `UserS3Bucket` is the role to which the S3 bucket policies are attached/detached.

It makes me wonder if we could use template method to push these three methods up into `AccessToS3Bucket`, e.g. subclasses implements a method which returns the IAM role name...something like this:

```python
class AccessToS3Bucket(TimeStampedModel):

    def aws_role_name():
          raise NotImplementedError

    def aws_create(self):
        services.attach_bucket_access_to_role(
            self.s3bucket.name,
            self.has_readwrite_access(),
            self.aws_role_name(), # <-- Note this
        )

    # similar thing for `aws_update()`/`aws_delete()`
   

class AppS3Bucket(AccessToS3Bucket):

    def aws_role_name():
          return self.app.aws_role_name


class UserS3Bucket(AccessToS3Bucket):

    def aws_role_name():
          return self.user.aws_role_name
```

I don't know if any of this make sense @jmoz - what do you reckon?

jmoz: Yep that's good OO I'll use your code

andyhd: nice one

control_panel_api/tests/test_views.py
```diff
@@ -225,36 +201,26 @@ def test_update(self, mock_aws_update):
...
-        data = {
-            'app': self.app_2.id,
-            's3bucket': self.s3_bucket_1.id,
-            'access_level': AppS3Bucket.READWRITE,
-        }
-        response = self.client.put(
-            reverse('apps3bucket-detail', (self.apps3bucket_1.id,)), data)
-        self.assertEqual(HTTP_400_BAD_REQUEST, response.status_code)
+    def test_update_bad_requests(self):
+        fixtures = (
```
xoen: nice one 👍

control_panel_api/tests/test_views.py
```diff
@@ -282,72 +248,131 @@ def test_detail(self):
...
-        response = self.client.post(reverse('s3bucket-list'), data)
-        self.assertEqual(HTTP_400_BAD_REQUEST, response.status_code)
+        mock_aws_create.assert_called()
 
-    def test_create_when_name_short(self):
-        data = {'name': 'ab'}
-        response = self.client.post(reverse('s3bucket-list'), data)
-        self.assertEqual(HTTP_400_BAD_REQUEST, response.status_code)
+    def test_create_bad_requests(self):
+        fixtures = (
```
xoen: even nicer :)

control_panel_api/serializers.py
```diff
@@ -46,11 +46,30 @@ class Meta:
...
+            )
+
+        return super().update(instance, validated_data)
+
+
+class UserS3BucketSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = UserS3Bucket
+        fields = ('id', 'url', 'user', 's3bucket', 'access_level')
```
xoen: @jmoz just realised, shouldn't we expose `is_admin` as well here?

jmoz: Yes, well spotted, thanks.

PR #33, opened 2017-10-06
Add COMMIT_HASH as docker image tag and parameter to jenkins job

andyhd:
## What

* Tag the docker image with the commit hash to avoid kubernetes caching the image by branch name and failing to redeploy the api branch on merge (among other cases)
* Pass the commit hash to Jenkins (which is updated to pass it to helm) to install the latest version of the image and not a cached version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,9 @@ jobs:
           name: Push application Docker image
           command: |
             docker login -u ${QUAY_USERNAME} -p ${QUAY_PASSWORD} quay.io
+            docker tag app "quay.io/mojanalytics/control-panel:${CIRCLE_SHA1}"
             docker tag app "quay.io/mojanalytics/control-panel:${CIRCLE_BRANCH}"
+            docker push "quay.io/mojanalytics/control-panel:${CIRCLE_SHA1}"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -56,13 +56,16 @@ jobs:
           name: Push application Docker image
           command: |
             docker login -u ${QUAY_USERNAME} -p ${QUAY_PASSWORD} quay.io
+            docker tag app "quay.io/mojanalytics/control-panel:${CIRCLE_SHA1}"
             docker tag app "quay.io/mojanalytics/control-panel:${CIRCLE_BRANCH}"
+            docker push "quay.io/mojanalytics/control-panel:${CIRCLE_SHA1}"
             docker push "quay.io/mojanalytics/control-panel:${CIRCLE_BRANCH}"
 
       - deploy:
```
xoen: Question: Doesn't this mean we always deploy after tests pass? So in theory the environment or branch shouldn't matter from the CD perspective? I may miss something

andyhd: Yes, continuous deployment basically means deploying when tests pass. What this essentially does is create an environment for every branch.

PR #35, opened 2017-10-09
AppUser implementation and tests with migrations

jmoz:
Run the tests.


control_panel_api/models.py
```diff
@@ -60,6 +62,18 @@ def aws_delete_role(self):
         services.delete_role(self.aws_role_name)
 
 
+class AppUser(TimeStampedModel):
```
xoen: Regarding the name of this model, I know in the design diagram we called this `AppUser` but, I always thought about these models in terms of `SubjectObject`.
E.g. `UserS3Bucket` or `AppS3Bucket` because it's the user or the app the subject accessing the S3 bucket.
Shouldn't we call this `UserApp` to be consistent with this?

control_panel_api/models.py
```diff
@@ -60,6 +62,18 @@ def aws_delete_role(self):
...
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
```
xoen: I think we should be consistent with the other "association models" and add `related_name='userapps'` here and to the `app` association below.

control_panel_api/models.py
```diff
@@ -45,6 +45,8 @@ def _slugify(name):
     slug = AutoSlugField(populate_from='name', slugify_function=_slugify)
     repo_url = models.URLField(max_length=512, blank=True, default='')
     created_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
+    users = models.ManyToManyField(
```
xoen: On the subject of relationships (see other comment), do we need this association now? An app would have many `userapps`, objects containing app they're in and if they're an admin for the app.

In the serialiser will expose `userapps` which means the client will get the apps there (as it gets the S3 buckets via `users3buckets`) and this will be consistent with the other associations.

control_panel_api/models.py
```diff
@@ -60,6 +62,18 @@ def aws_delete_role(self):
         services.delete_role(self.aws_role_name)
 
 
+class AppUser(TimeStampedModel):
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    app = models.ForeignKey(App, on_delete=models.CASCADE)
```
xoen: See comment above about `related_name='userapps'`.

control_panel_api/serializers.py
```diff
@@ -6,7 +6,8 @@
     AppS3Bucket,
     S3Bucket,
     User,
-    UserS3Bucket)
+    UserS3Bucket,
+    AppUser)
```
xoen: Minor style point, I tend to keep these sorted alphabetically and also leave a trailing comma/newline after last element, e.g.

```python
from control_panel_api.models import (
    AppS3Bucket,
    # ...
    UserApp,
    UserS3Bucket,
)
```

...but this is really me being a bit of an OCD person, sorry :P

control_panel_api/tests/test_views.py
```diff
@@ -15,7 +15,10 @@
     App,
     AppS3Bucket,
     S3Bucket,
-    User, UserS3Bucket)
+    User,
+    UserS3Bucket,
+    AppUser
```
xoen: See other comment on annoying style, ignore me if too annoying :)

control_panel_api/serializers.py
andyhd: +1

andyhd: Not that bothered about the trailing comma, though - it's a bad habit to get into if you then start writing JSON, where it is an error

xoen: Agreed, I like the trailing commas as then someone else can add an element at the end without modifying the existing lines of code. But yes, not in JSON as not valid 🐛

control_panel_api/urls.py
```diff
@@ -5,14 +5,14 @@
...
 
-
 router = routers.DefaultRouter()
 router.register(r'apps', views.AppViewSet)
 router.register(r'apps3buckets', views.AppS3BucketViewSet)
 router.register(r'users3buckets', views.UserS3BucketViewSet)
 router.register(r'groups', views.GroupViewSet)
 router.register(r's3buckets', views.S3BucketViewSet)
 router.register(r'users', views.UserViewSet)
+router.register(r'appusers', views.AppUserViewSet)
```
xoen: I think the general rule on trying to keep things alphabetically sorted applies - unless there is a specific reason why order is important. Don't think it's the case in this instance (but as I said, I'd prefer the model/entpoints to be `UserApp`/`userapps` as in my view it's the user that is the subject managing the apps in this case).

control_panel_api/views.py
```diff
@@ -1,8 +1,6 @@
 from django.contrib.auth.models import Group
 from rest_framework import viewsets
-from rest_framework.exceptions import ValidationError
```
xoen: Nice one 👍

control_panel_api/views.py
```diff
@@ -25,7 +25,9 @@
     GroupSerializer,
     S3BucketSerializer,
     UserSerializer,
-    UserS3BucketSerializer)
+    UserS3BucketSerializer,
+    AppUserSerializer
```
xoen: As per other imports (and lists where order is not important), I'd keep this alphabetically sorted and with a trailing comma.

control_panel_api/serializers.py
andyhd: Although it's a tiny thing, I apply YAGNI - code for what you need now, not what you might need in future.

xoen: I like the open-close principle although: https://en.wikipedia.org/wiki/Open/closed_principle

> software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification

e.g. the next person can add a line to the list without modifying existing lines - but as you said, it's a tiny application of it. Just to be clear, I'm not that bothered really :)

control_panel_api/serializers.py
```diff
@@ -80,3 +81,22 @@ class S3BucketSerializer(serializers.ModelSerializer):
     class Meta:
         model = S3Bucket
         fields = ('id', 'url', 'name', 'arn', 'apps3buckets', 'created_by')
```
xoen: Shouldn't we add/expose the `userapps` association here and in the `AppSerializer` for consistency?

xoen:
I think overall this is really good @jmoz 🌟 - The main things I think should change are:
 * the name of the model/associations
 * adding the `userapps` relationship (via `related_name='userapps'`)
 * expose this `userapps` association in the `UserSerializer` and `AppSerializer`

control_panel_api/serializers.py
andyhd: The open-closed principle is a good one, but it doesn't apply to lists in source code 😛

control_panel_api/serializers.py
xoen: Speaking of serialisers and associations, `S3BucketSerializer` should also expose `users3buckets` (currently only expose `apps3buckets`.
Basically, we should be able to get these "through" associations from all the serialisers for `App`, `S3Bucket` and `User`.

xoen:
🌟

PR #36, opened 2017-10-11
Generate `App.slug` from `repo_url` field

xoen:
### What

* Generate the app slug field as before but from the repository name (last part of `repo_url` field) rather than from the `name` field (entered by the user)
* The `repo_url` field now it mandatory and unique
* Expose app IAM role name in API response
* Renamed `aws_role_name` => `iam_role_name`

### Repository name definition
Part of `repo_url` between the last slash (`/`) and `.git` (if present)

Examples:

`repo_url` | `repo_name`
------------ | -------------
`https://www.example.com/org1/repo-one` | `repo-one`
`git@github.com:org/org1/repo-two.git` | `repo-two`

### Ticket
https://trello.com/c/4LKCxdYN/417-cp-api-generate-app-slug-from-repourl


control_panel_api/models.py
```diff
@@ -59,6 +59,23 @@ def aws_create_role(self):
...
+        The name is the part after the last slash in the URL, without
+        the '.git' (if present).
+
+        Examples:
+
+        "https://github.com/org/a_repo_name" => "a_repo_name"
+        "git@github.com:org/repo_2.git" => "repo_2"
+        '''
+
+        last_after_slash = self.repo_url.split('/')[-1]
```
jmoz: Do we need to include the user part as well else this can easily conflict.  The user/repo_name is what makes the url unique.  E.g. /github/james/foo /github/aldo/foo

control_panel_api/serializers.py
```diff
@@ -24,6 +24,9 @@ class Meta:
 
 
 class AppSerializer(serializers.ModelSerializer):
+
+    iam_role_name = serializers.ReadOnlyField(source='aws_role_name')
```
jmoz: I would not switch around field names as it causes confusion (I just had to trace code changes to figure out what this field was), just name it the same or edit one of them to match.

control_panel_api/models.py
xoen: As discussed offline, repos are on the same organisation and also, `slugify` takes care of collisions.

xoen:
(Thanks for the pycharm magical rename James 👍)

control_panel_api/tests/test_views.py
```diff
@@ -134,13 +137,15 @@ def test_update(self):
         self.assertEqual(data['name'], response.data['name'])
 
 
+from django.test import tag
+@tag('fluffy')
```
jmoz: 👎

xoen: Cheers @jmoz: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/36/commits/a1c2dfeb312b7c9a97b0add9fd8fd7087048cf47

jmoz:
LGTM

PR #37, opened 2017-10-16
Add filtering apps by repo_url field

xoen:
### What

Ability to filter apps by `repo_url` field.

This is part of the work to avoid asking the user for the app IAM role name when deploying.

I updated the `SuperusersOnlyFilter` class to inherit from `django_filters.rest_framework.DjangoFilterBackend`, this is the class currently used by all the views (will change once we give access to non-superuser users and we need more granular permissions).

### Relevant DRF / django_filters documentation
* [DRF's filtering](http://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend)
* [django-filter](https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html)

### Ticket
https://trello.com/c/vtPRBZfa/418-cp-api-investigate-implement-app-filtering-by-repourl

### DIFF
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -3,17 +3,20 @@ Custom filters
 
 See: http://www.django-rest-framework.org/api-guide/filtering/#custom-generic-filtering
 """
-from rest_framework.filters import BaseFilterBackend
+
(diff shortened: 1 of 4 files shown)


xoen:
Cheers @andyhd

PR #38, opened 2017-10-16
Change the user to use auth0_id as primary key

jmoz:
## What

We now use a string field auth0_id as primary key, which can be mapped from auth0 like `github|123`, we can map the github username to the `username` field.  This will be implemented in future work.

This requires a db rebuild as the django migrations do not work with a simple alter field.  The squashed initial migration fixes this and is also best practice to regularly squash.

The createsuperuser command works and prompts for auth0_id and username.

## How to review

1. Run the tests.
2. Drop the db for dev and run migration.

https://trello.com/c/5IBvOn39/419-cp-api-add-auth0id-to-user-model


xoen:
I’ll read through this and see if I can think of any way to avoid destroying the DB, which is particularly **annoying** in the case of the `alpha` environment.

control_panel_api/models.py
```diff
@@ -26,7 +29,7 @@ def get_short_name(self):
 
     @property
     def aws_role_name(self):
-        return f"{settings.ENV}_user_{self.username.lower()}"
+        return f"{settings.ENV}_user_{self.auth0_id}"
```
xoen: This shouldn't change, I'm not even sure you can call an IAM role something like `dev_user_github|123`, plus it would break compatibility with existing stuff.

control_panel_api/models.py
```diff
@@ -11,10 +11,13 @@
 
 
 class User(AbstractUser):
+    auth0_id = models.CharField(max_length=128, primary_key=True)
     name = models.CharField(max_length=256, blank=True)
 
     teams = models.ManyToManyField('Team', through='TeamMembership')
 
+    REQUIRED_FIELDS = ['email', 'auth0_id']
```
xoen: `username` should also be required.

control_panel_api/tests/test_models.py
```diff
@@ -19,11 +18,11 @@
...
     @patch('control_panel_api.services.create_role')
     def test_aws_create_role_calls_service(self, mock_create_role):
-        username = 'james'
-        user = User.objects.create(username=username)
+        auth0_id = 'github|1'
+        user = User.objects.create(auth0_id=auth0_id)
         user.aws_create_role()
 
-        expected_role_name = f"test_user_{username}"
+        expected_role_name = f"test_user_{auth0_id}"
```
xoen: As said, the IAM role name shouldn't change. It'll have to be generated from the (GitHub) username as before to maintain compatibility.

[Also, Auth0's user ID includes a pipe (`|`) which is [not a valid character](http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html) for an IAM name.]

control_panel_api/models.py
jmoz: Ok. Ticket mentions it. Will revert.

control_panel_api/models.py
jmoz: This is implicit.

When explicitly specified you get:

"ERRORS:
control_panel_api.User: (auth.E002) The field named as the 'USERNAME_FIELD' for a custom user model must not be included in 'REQUIRED_FIELDS'."

control_panel_api/tests/test_permissions.py
```diff
@@ -1,6 +1,5 @@
 from unittest.mock import patch
 
-from django.test.utils import override_settings
```
xoen: nice 👍

control_panel_api/models.py
xoen: Right...it's implicit because `username` in the `AbstractUser` class says `username` is `unique` which means can't be blank.

control_panel_api/migrations/0001_initial.py
```diff
@@ -42,16 +47,146 @@ class Migration(migrations.Migration):
...
+            name='App',
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('created', models.DateTimeField(auto_now_add=True)),
+                ('created', django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created')),
+                ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified')),
                 ('name', models.CharField(max_length=100)),
-                ('slug', models.CharField(max_length=100)),
-                ('repository', models.CharField(blank=True, default='', max_length=255)),
+                ('slug', django_extensions.db.fields.AutoSlugField(blank=True, editable=False, populate_from='name')),
```
xoen: I think [this is wrong now](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/api/control_panel_api/models.py#L45), I don't know if it's because the squash happened before [that change](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/36)?

This automated squashing is not giving me much confidence to be honest... :)

control_panel_api/models.py
andyhd: Explicit is better than implicit - can we put username in?

xoen:
As discussed offline with James, I'm not a great fun of having to destroy "live" databases *but* we're kind of forced to do in this case.

The following may work (as quickly tested locally):
1) Use django's `python manage.py dumpdata control_panel_api` to get a dump of dev/alpha data
2) Update the JSON file by changing the PK values (and the FK references!!)
3) Update the DB to have the new schema
4) Copy this updated file to one of the `dev`/`alpha` pod - using `kubectl cp ...`
5) Run `python manage.py loaddata file`

CC: @andyhd @kerin any thoughts on this?

kerin:
@xoen just drop and recreate the DB - nothing is in use yet, so it won't impact any users

PR #42, opened 2017-10-19
Auth0 authn

andyhd:
## What

* Rebased @jmoz's work
* Accept `Authorization: JWT <token>` header and verify JWT to authenticate user

## How to review

1. Run `docker-compose build` to build the docker image
2. Make sure you have the correct values in `.env`:

    | name | value |
    | ----- | ----- |
    | `OIDC_CLIENT_ID` | The client ID for the dev `API Client` in Auth0 |
    | `OIDC_CLIENT_SECRET` | The client secret for the dev `API Client` in Auth0 |

3. Use `docker-compose up` to run an instance of the API
4. `docker-compose exec app python3 manage.py createsuperuser` to create a user with an `auth0_id`
5. Run the [frontend](https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/tree/api-client-auth)
6. Login with Auth0
7. See your user data retrieved from the API

See [Trello #423](https://trello.com/c/kRVSY6XF)

### DIFF
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -1,0 +1,35 @@
+import jwt
+from django.conf import settings
+from jwt.exceptions import InvalidTokenError
+from rest_framework.authentication import BaseAuthentication
+from rest_framework.exceptions import AuthenticationFailed
+
+from control_panel_api.models import User
+
+
+class Auth0JWTAuthentication(BaseAuthentication):
+    def authenticate(self, request):
+        try:
+            prefix, token = request.META.get('HTTP_AUTHORIZATION', '').split()
+        except (ValueError, TypeError):
+            return None
+
+        if prefix != 'JWT':
+            raise AuthenticationFailed('JWT prefix missing')
+
+        try:
+            decoded = jwt.decode(token, key=settings.AUTH0_SECRET, audience=settings.AUTH0_AUDIENCE)
+        except InvalidTokenError:
+            raise AuthenticationFailed('JWT decode error')
+
+        try:
+            username = decoded['sub']
+        except KeyError:
+            raise AuthenticationFailed('JWT missing sub username field')
+
+        try:
+            user = User.objects.get(username=username)
+        except User.DoesNotExist:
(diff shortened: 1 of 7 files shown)


control_panel_api/tests/test_authentication.py
```diff
@@ -0,0 +1,52 @@
...
+from model_mommy import mommy
+from rest_framework.reverse import reverse
+from rest_framework.status import HTTP_403_FORBIDDEN, HTTP_200_OK
+from rest_framework.test import APITestCase
+
+from control_panel_api.authentication import Auth0JWTAuthentication
+
+
+# Decode token at https://jwt.io/
+GOOD_TOKEN = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJlbWFpbCI6ImphbWVzQGptb3ouY28udWsiLCJuYW1lIjoiSmFtZXMgTW9ycmlzIiwiYXVkIjoiYXVkaWVuY2UiLCJzdWIiOiJnaXRodWJ8MTIzNDUifQ.Vfvhtm_TXbtOBKcIWed0YzVH7gIKlSdzg36bbIK6UZ4'  # noqa
```
jmoz: We could improve the token tests and make the more readable by creating a function to encode a token

andyhd: Yep, I will update the test to build the token with `jwt.encode`, so we can see what is inside.

control_panel_api/authentication.py
```diff
@@ -0,0 +1,47 @@
...
+        except (ValueError, TypeError):
+            return None
+
+        if prefix != 'JWT'.encode():
+            raise AuthenticationFailed('JWT prefix missing')
+
+        try:
+            decoded = jwt.decode(
+                token,
+                key=settings.AUTH0_CLIENT_SECRET,
```
kerin: This might be quite an annoying suggestion, but would it be better to call these vars `OIDC_CLIENT_*` rather than referencing a specific vendor?

PR #44, opened 2017-10-20
Normalise App.repo_url to not have the trailing '.git'

xoen:
### What
To avoid ambiguity we're ensuring that the App repo URL will always have
the same format. This will enable us to simplify the Jenkins job to deploy
a webapp.

This gets the app details by filtering apps by repo_url but currently has
to perform two separate HTTP requests (in the worst-case scenario) to
try both versions of the repo URL.

This will mean we can just filter by the `repo_url` without the trailing
".git".

### Part of ticket
https://trello.com/c/yfUlKPdb/428-cp-api-jenkins-normalise-repourl-and-avoid-possible-second-request

### DIFF
--- a/control_panel_api/tests/test_views.py
+++ b/control_panel_api/tests/test_views.py
@@ -143,6 +143,13 @@ class AppViewTest(AuthenticatedClientMixin, APITestCase):
 
         self.assertEqual(self.superuser.auth0_id, response.data['created_by'])
 
+    @patch('control_panel_api.models.App.aws_create_role')
+    def test_create_normalises_repo_url(self, mock_aws_create_role):
+        data = {'name': 'foo', 'repo_url': 'https://example.com.git'}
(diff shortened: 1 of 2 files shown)


xoen:
Cheers Andy

control_panel_api/tests/test_views.py
```diff
@@ -143,13 +143,26 @@ def test_create(self, mock_aws_create_role):
 
         self.assertEqual(self.superuser.auth0_id, response.data['created_by'])
 
+    @patch('control_panel_api.models.App.aws_create_role')
```
jmoz: You are mocking but doing nothing with it.  The patch doesn't need to be there.

xoen: 👍

andyhd: ugh, I missed something :(

PR #45, opened 2017-10-20
Import script map auth0 id

jmoz:
## What

Uses the auth0 client to get a list of auth0 github users and map the correct user_id.

The auth0 client and code in moj_analytics/ is existing code and was checked out from master branch.

## How to review

1. Add auth0 env vars
2. Run `./manage.py import_k8s_users`
3. Profit!

### DIFF
--- a/control_panel_api/management/commands/import_k8s_users.py
+++ b/control_panel_api/management/commands/import_k8s_users.py
@@ -1,42 +1,76 @@
 import base64
 import logging
 
+from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db.utils import IntegrityError
-from kubernetes import config, client
+from kubernetes import client, config
 
 from control_panel_api.models import User
+from moj_analytics.auth0_client import Auth0, ManagementAPI, User as Auth0User
 
 logger = logging.getLogger(__name__)
 
 
+def get_auth0_users():
+    api = Auth0(
+        settings.OIDC_CLIENT_ID,
+        settings.OIDC_CLIENT_SECRET,
+        settings.OIDC_DOMAIN
+    )
+    api.access(ManagementAPI(settings.OIDC_DOMAIN))
+
+    results = api.management.get_all(Auth0User, 'connection=github')
+
+    return results
+
+
 class Command(BaseCommand):
-    help = "Import Kubernetes user-secrets data into the User.  Uses KUBECONFIG envvar or defaults to ~/.kube/config"
+    help = "Import Kubernetes user-secrets data into the User.  " \
+           "Uses KUBECONFIG envvar or defaults to ~/.kube/config"
 
     USER_SECRETS_NAME = 'user-secrets'
 
     def handle(self, *args, **options):
         config.load_kube_config()
         v1 = client.CoreV1Api()
+
         results = v1.list_secret_for_all_namespaces()
 
-        user_secrets = [item.data for item in results.items if item.metadata.name == Command.USER_SECRETS_NAME]
+        user_secrets = [
+            item.data for item in results.items
+            if item.metadata.name == Command.USER_SECRETS_NAME]
+
+        decoded_user_secrets = [
+            {k: base64.b64decode(v).decode() for k, v in item.items()}
+            for item in user_secrets]
+
+        auth0_users_by_nickname = {
+            item['nickname'].lower(): item for item in get_auth0_users()}
 
         users_added = 0
-        for item in user_secrets:
-            username = base64.b64decode(item['username'])
-            email = base64.b64decode(item['email'])
-            name = base64.b64decode(item['fullname'])
+        for item in decoded_user_secrets:
+            username = item['username']
+            email = item['email']
+            name = item['fullname']
+
+            try:
+                auth0_id = auth0_users_by_nickname[username.lower()]['user_id']
+            except KeyError:
+                logger.warning(f"User {username} not found in auth0")
+                continue
 
             try:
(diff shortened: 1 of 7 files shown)


xoen:
LGTM

control_panel_api/management/commands/import_k8s_users.py
```diff
@@ -1,42 +1,76 @@
...
+    api = Auth0(
+        settings.OIDC_CLIENT_ID,
+        settings.OIDC_CLIENT_SECRET,
+        settings.OIDC_DOMAIN
+    )
+    api.access(ManagementAPI(settings.OIDC_DOMAIN))
+
+    results = api.management.get_all(Auth0User, 'connection=github')
+
+    return results
```
andyhd: No need for a variable, just inline `return api.manage.get_all(...`

control_panel_api/management/commands/import_k8s_users.py
```diff
@@ -1,42 +1,76 @@
...
+    help = "Import Kubernetes user-secrets data into the User.  " \
+           "Uses KUBECONFIG envvar or defaults to ~/.kube/config"
 
     USER_SECRETS_NAME = 'user-secrets'
 
     def handle(self, *args, **options):
         config.load_kube_config()
         v1 = client.CoreV1Api()
+
         results = v1.list_secret_for_all_namespaces()
```
andyhd: I know this is @kerin's code, but `v1` and `results` are not well named. I'd suggest `kubernetes_api` and `secrets`.

control_panel_api/management/commands/import_k8s_users.py
```diff
@@ -1,42 +1,76 @@
...
 
-        user_secrets = [item.data for item in results.items if item.metadata.name == Command.USER_SECRETS_NAME]
+        user_secrets = [
+            item.data for item in results.items
+            if item.metadata.name == Command.USER_SECRETS_NAME]
```
andyhd: I'd pull this out into a function `get_user_secrets`, to keep the code in `handle` at a single level of abstraction.

control_panel_api/management/commands/import_k8s_users.py
```diff
@@ -1,42 +1,76 @@
...
+
+        decoded_user_secrets = [
+            {k: base64.b64decode(v).decode() for k, v in item.items()}
+            for item in user_secrets]
```
andyhd: I'd probably include this in `get_user_secrets` - the `handle` method shouldn't care about needing to decode the secrets.

control_panel_api/management/commands/import_k8s_users.py
```diff
@@ -1,42 +1,76 @@
...
+
+        auth0_users_by_nickname = {
+            item['nickname'].lower(): item for item in get_auth0_users()}
```
andyhd: You're only using this dict for mapping username to auth0_id, so it should just be:
```python
auth0_id_by_nickname = {item['nickname'].lower(): item['user_id'] for item in get_auth0_users()}
```

control_panel_api/management/commands/import_k8s_users.py
```diff
@@ -1,42 +1,76 @@
...
+        for item in decoded_user_secrets:
+            username = item['username']
+            email = item['email']
+            name = item['fullname']
+
+            try:
+                auth0_id = auth0_users_by_nickname[username.lower()]['user_id']
+            except KeyError:
+                logger.warning(f"User {username} not found in auth0")
+                continue
```
andyhd: I know `try...except` is more Pythonic, but I think this would be more clear written as:
```python
auth0_id = auth0_id_by_nickname.get(username.lower())
if auth0_id is None:
    logger.warning(...)
    continue
```

moj_analytics/tests/conftest.py
```diff
@@ -0,0 +1,228 @@
+import json
+from unittest import mock
+
+import pytest
+from responses import RequestsMock
```
andyhd: These requirements are missing from `requirements.txt` - but maybe the `mojanalytics` module should be published to Pypi and added to `requirements.txt` instead

control_panel_api/management/commands/import_k8s_users.py
kerin: @andyhd nothing to do with me!

control_panel_api/management/commands/import_k8s_users.py
jmoz: Yeh, I was using it for debug purposes.

control_panel_api/management/commands/import_k8s_users.py
jmoz: Well I'm filtering and reducing it which is usually an inline operation.

control_panel_api/management/commands/import_k8s_users.py
jmoz: I did have that originally but then you will see below I make use of exceptions so I decided to keep it consistent. So consistency and pythonic, or clearer what do you prefer?

moj_analytics/tests/conftest.py
jmoz: For the time being I'd just say add the requirements.

jmoz: Also @andyhd I looked at the moj_analytics stuff and noticed it has **a lot** of pytest code.  I'd prefer if this was rewritten to standard django unit test and using mock/patch.  

A lot of it could be simplified or it indicates the client should be.

PR #48, opened 2017-10-24
Document running with docker-compose or on host

andyhd:
## What

* Added documentation on running with docker-compose
* Added list of environment variables


andyhd:
This could do with some explanation of the environment variables, but I don't know what some of them are.

PR #49, opened 2017-10-25
App and S3Bucket return nested data

jmoz:
## What

This returns some of the nested data when on app or s3bucket to prevent multiple calls for data we already have available.

Use different serializers which exclude the parent data.

Create simple serializer for app and s3bucket which does not return relations.

## App output

```
{
  "count": 1,
  "next": null,
  "previous": null,
  "results": [
    {
      "id": 1,
      "url": "http://127.0.0.1:8003/apps/1/",
      "name": "jamesapp",
      "slug": "foocom",
      "repo_url": "http://foo.com",
      "iam_role_name": "dev_app_foocom",
      "created_by": "foo",
      "apps3buckets": [
        {
          "id": 1,
          "url": "http://127.0.0.1:8003/apps3buckets/1/",
          "s3bucket": {
            "id": 2,
            "url": "http://127.0.0.1:8003/s3buckets/2/",
            "name": "dev-james2",
            "arn": "arn:aws:s3:::dev-james2",
            "created_by": "foo"
          },
          "access_level": "readwrite"
        }
      ],
      "userapps": []
    }
  ]
}
```

## S3Bucket output

```
{
  "count": 1,
  "next": null,
  "previous": null,
  "results": [
    {
      "id": 2,
      "url": "http://127.0.0.1:8003/s3buckets/2/",
      "name": "dev-james2",
      "arn": "arn:aws:s3:::dev-james2",
      "apps3buckets": [
        {
          "id": 1,
          "url": "http://127.0.0.1:8003/apps3buckets/1/",
          "app": {
            "id": 1,
            "url": "http://127.0.0.1:8003/apps/1/",
            "name": "jamesapp",
            "slug": "foocom",
            "repo_url": "http://foo.com",
            "iam_role_name": "dev_app_foocom",
            "created_by": "foo"
          },
          "access_level": "readwrite"
        }
      ],
      "created_by": "foo"
    }
  ]
}
```

## How to review

1. ./manage.py test
2. Check the endpoints for app and s3bucket

### DIFF
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -35,27 +35,6 @@ class GroupSerializer(serializers.ModelSerializer):
         fields = ('id', 'url', 'name')
 
 
-class AppSerializer(serializers.ModelSerializer):
-
-    class Meta:
-        model = App
-        fields = (
-            'id',
-            'url',
-            'name',
-            'slug',
-            'repo_url',
-            'iam_role_name',
-            'created_by',
-            'apps3buckets',
-            'userapps',
-        )
-
-    def validate_repo_url(self, value):
-        """Normalise repo URLs by removing trailing .git"""
-        return value.rsplit(".git", 1)[0]
-
-
 class AppS3BucketSerializer(serializers.ModelSerializer):
 
     class Meta:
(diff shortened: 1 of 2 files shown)


xoen:
I've noticed that we don't do the same for `userapps`/`users3buckets` in the `UserSerializer` class.
I guess we could do in a separate iteration but it may be a good idea to be consistent.

control_panel_api/serializers.py
```diff
@@ -94,11 +73,76 @@ def update(self, instance, validated_data):
...
+        fields = (
+            'id',
+            'url',
+            'name',
+            'slug',
+            'repo_url',
+            'iam_role_name',
+            'created_by',
+            'apps3buckets',
+            'userapps',
```
xoen: At some point we'll want the `userapps` to behave similarly for consistency as well.

xoen:
Interesting approach. As said, at some point we'll probably want the remaining associations to be exposed in a consistent way.

Although for now this will be enough I'm temped to say, let's change all the other associations for consistency now. Otherwise looks good to me 👍

control_panel_api/serializers.py
jmoz: Yes if we need the extra data for @axemonkey we can do that later.

jmoz: I think for now this is what he needs, let's see how he uses some of the other endpoints if he needs extra data, then we can add this approach or refactor to a shared implementation.

PR #50, opened 2017-10-25
Add kubernetes/helm to the Docker image

xoen:
### What
Added k8s (kubernetes `v1.7.7`) and helm (`v2.6.1`) to the docker image.

Also added the moj-analytics helm repository

**NOTE**: I had to install `openssl` to solve a weird SSL certificate problem when getting the tarballs via `wget`.

### Why
Set up a user within the platform involves the installation of a couple of our helm charts.

### Ticket
https://trello.com/c/8dwIeL6w/455-ensure-helm-client-is-present-in-cp-api-docker-container-3

This is similar to what [we already do in the jenkins-agent image](https://github.com/ministryofjustice/analytics-platform-jenkins-agent/blob/master/Dockerfile#L13).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,30 @@ RUN apk add --no-cache \
 
 # install python3 and 'ca-certificates' so that HTTPS works consistently
 RUN apk add --no-cache \
+        openssl \
         ca-certificates \
         libffi-dev \
         libressl-dev \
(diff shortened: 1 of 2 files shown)


xoen:
Trying to rename the branch to see if it fixes the jenkins job

PR #56, opened 2017-10-27
Port helm code from master branch

jmoz:
## What

User create now also calls helm init and config.

Unit test switches out subprocess for magicmock to prevent accidental running of helm commands when developing and testing.

Other PRs with env var.

The request time for User create is now around 8 seconds.

_Latest commit shows a different method of patching out live api/process calls.  We will have to patch at class level test cases that interact with apis or system libraries, but by passing `MagicMock()` as the second parameter this stops the behaviour of having to add a parameter to all test functions._

## How to review

1. ./manage.py test
2. Create a User via api and check stdout for helm commands.

https://trello.com/c/QiDTzyob/454-when-a-user-is-created-via-api-we-run-helm-commands-3

### DIFF
--- a/control_panel_api/helm.py
+++ b/control_panel_api/helm.py
@@ -1,0 +1,35 @@
+from django.conf import settings
+
+
+def _get_subprocess():
+    """Allow the subprocess module to be switched out at runtime for a
+    magicmock to prevent real calls during testing
+    """
+    return settings.SUBPROCESS_MODULE
+
+
+def init_user(username, email, fullname):
+    helm_upgrade(
+        f'init-user-{username}',
+        'mojanalytics/init-user',
+        '--set', f'NFSHostname={settings.NFS_HOSTNAME}',
+        '--set', f'Username={username}',
+        '--set', f'Email={email}',
+        '--set', f'Fullname={fullname}',
+    )
+
+
+def config_user(username):
+    helm_upgrade(
+        f'config-user-{username}',
+        'mojanalytics/config-user',
+        '--namespace', f'user-{username}',
+        '--set', f'Username={username}',
+    )
+
+
+def helm_upgrade(release, chart, *flags):
+    default_flags = ['--install', '--wait']
+    flags = list(flags) + default_flags
+    _get_subprocess().run(
+        ['helm', 'upgrade', release, chart] + flags, check=True)
--- a/control_panel_api/helm.py
+++ b/control_panel_api/helm.py
@@ -1,11 +1,6 @@
+import subprocess
+
 from django.conf import settings
-
-
-def _get_subprocess():
-    """Allow the subprocess module to be switched out at runtime for a
-    magicmock to prevent real calls during testing
-    """
-    return settings.SUBPROCESS_MODULE
 
 
 def init_user(username, email, fullname):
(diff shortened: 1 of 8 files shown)


control_panel_api/helm.py
```diff
@@ -0,0 +1,35 @@
+from django.conf import settings
+
+
+def _get_subprocess():
+    """Allow the subprocess module to be switched out at runtime for a
+    magicmock to prevent real calls during testing
+    """
+    return settings.SUBPROCESS_MODULE
```
andyhd: In retrospect, I think this approach to swapping out modules for test purposes is not so good.
1. Apart from the tests, the app has no need for this setting. This simplifies the test code at the expense of the app code.
2. It also moves the point where the substitution of the module happens away from the tests in which it is used. The tests should be explicit about their setup.

The tests should instead monkey-patch the `helm_upgrade` function, eg:
```python
@mock.patch('control_panel_api.models.helm.helm_upgrade')
```
or even `subprocess.run`
```python
@mock.patch('control_panel_api.helm.subprocess.run')
```

jmoz: I initially had done that with the aws stuff but required multiple patches on classes, functions etc and still was accidentally called.

How would you suggest a simple and effective way so don't have to litter code with patches and accidentally miss any?  It could get called from a view or model etc

We can patch individual tests, or at class level.  Is there a better way?

andyhd: Yeah, it is a tricky problem and I'm now thinking of a solution which is similar to what you already have.

Instead of having the modules to swap out as settings, maybe we need something like feature flags - settings like `ENABLE_FEATURE=True`, maybe a dict of them if there are lots, and then the code could wrap (in this case) the call to `subprocess.run` with an `if ENABLE_SUBPROCESS` or `if ENABLE_HELM` and set them all to `False` in the test settings.

Although it's basically the same thing as what you've got, it seems a bit more straightforward, and we should probably use feature flags for new features we're developing anyway.

andyhd: I think that would soften the first issue I had, because the code would be self-documenting.
I'm not sure if having a version of the settings for tests is so good, as per my second issue. It would be more explicit to override the settings near the test code.

jmoz: Yes agree it's a tricky one, I sort of see what you're saying.  

The test settings is a pattern I've seen and used before - commonly switching out things like cache and session middlewares with dummy implementations e.g. https://docs.djangoproject.com/en/1.11/topics/cache/#dummy-caching-for-development

There's also `@override_settings` could help.

I was using the magicmock as it just gives you the interface and works - now noticing that's just a lazy version of the empty dummy cache class.

A better design could be a simple command runner class or interface that uses subprocess, which then could be switched out at runtime using the test environment for a dummy runner?

Not sure, further research required!

control_panel_api/models.py
```diff
@@ -37,6 +37,10 @@ def aws_create_role(self):
     def aws_delete_role(self):
         services.delete_role(self.iam_role_name)
 
+    def helm_create_user(self):
```
xoen: This is fine, you may omit the `user` bit as you're in the context of the `User` model, maybe rename to just `helm_create` or something like that.

PR #58, opened 2017-11-02
Users3bucket access level based on is_admin

jmoz:
## What

UserS3Bucket access_level should not be editable and should be based on the is_admin flag.

## How to review

1. ./manage.py test
2. Use api to create UserS3Bucket

https://trello.com/c/WuFheSAR/447-confirm-users-assigned-s3-bucket-policy-is-read-only-2

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -188,10 +188,10 @@ class AccessToS3Bucket(TimeStampedModel):
     class Meta:
         abstract = True
 
-    def has_readwrite_access(self):
-        return self.access_level == self.READWRITE
-
     def aws_role_name(self):
+        raise NotImplementedError
+
+    def has_readwrite_access(self):
         raise NotImplementedError
 
     def aws_create(self):
@@ -234,24 +234,37 @@ class AppS3Bucket(AccessToS3Bucket):
     def aws_role_name(self):
         return self.app.iam_role_name
 
+    def has_readwrite_access(self):
+        return self.access_level == self.READWRITE
+
 
 class UserS3Bucket(AccessToS3Bucket):
     """
     A user can have access to several S3 buckets.
 
     We have two access levels, "readonly" (default) and "readwrite".
-    The `is_admin` field determine if the user has admin privileges on the
-    S3 bucket
-    """
 
+    The `is_admin` field determines the `access_level` to the S3Bucket. The
+    `access_level` can not be changed directly.
+    """
     user = models.ForeignKey(
         User, related_name='users3buckets', on_delete=models.CASCADE)
     is_admin = models.BooleanField(default=False)
 
     class Meta:
-        # one record per user/s3bucket
         unique_together = ('user', 's3bucket')
         ordering = ('id',)
 
     def aws_role_name(self):
         return self.user.iam_role_name
+
+    def has_readwrite_access(self):
+        return self.is_admin
+
+    def save(self, **kwargs):
+        if self.has_readwrite_access():
+            self.access_level = self.READWRITE
+        else:
+            self.access_level = self.READONLY
+
+        super().save(**kwargs)
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -58,6 +58,7 @@ class UserS3BucketSerializer(serializers.ModelSerializer):
     class Meta:
         model = UserS3Bucket
         fields = ('id', 'url', 'user', 's3bucket', 'access_level', 'is_admin')
+        read_only_fields = ('access_level',)
 
     def update(self, instance, validated_data):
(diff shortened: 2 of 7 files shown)


control_panel_api/models.py
```diff
@@ -234,24 +234,37 @@ class Meta:
...
 
     class Meta:
-        # one record per user/s3bucket
         unique_together = ('user', 's3bucket')
         ordering = ('id',)
 
     def aws_role_name(self):
         return self.user.iam_role_name
+
+    def has_readwrite_access(self):
```
xoen: I don't think this is a good name for what the method is doing. This is checking if the user can admin the bucket, not if he can read/write on it.

I know it seems like an admin of an S3 bucket can also read and write but not necessarily.
I may be an admin for an S3 bucket containing sensitive information but not having the ability to read and write it.

control_panel_api/models.py
```diff
@@ -234,24 +234,37 @@ class Meta:
...
+        return self.is_admin
+
+    def save(self, **kwargs):
+        if self.has_readwrite_access():
```
xoen: Again, this feels a bit confusing. When saving one of this records, this would sneakingly change the `access_level` even if I didn't ask for it which is a bit surprisingly.

jmoz: Check the ticket please https://trello.com/c/WuFheSAR/447-confirm-users-assigned-s3-bucket-policy-is-read-only-2

control_panel_api/models.py
jmoz: Suggestions?  As I'm just adapting existing code and template method.

control_panel_api/serializers.py
```diff
@@ -58,6 +58,7 @@ class UserS3BucketSerializer(serializers.ModelSerializer):
     class Meta:
         model = UserS3Bucket
         fields = ('id', 'url', 'user', 's3bucket', 'access_level', 'is_admin')
+        read_only_fields = ('access_level',)
```
xoen: Why do we have this `access_level` field if we don't even use in practice?
As I said in a separate comment being able to RW is independent from being an admin for an S3 bucket.

control_panel_api/models.py
```diff
@@ -234,24 +234,37 @@ class Meta:
...
-    The `is_admin` field determine if the user has admin privileges on the
-    S3 bucket
-    """
 
+    The `is_admin` field determines the `access_level` to the S3Bucket. The
+    `access_level` can not be changed directly.
+    """
     user = models.ForeignKey(
         User, related_name='users3buckets', on_delete=models.CASCADE)
     is_admin = models.BooleanField(default=False)
```
andyhd: I feel field name doesn't quite fit - it should be `user_is_admin`

control_panel_api/models.py
```diff
@@ -234,24 +234,37 @@ class Meta:
...
         return self.user.iam_role_name
+
+    def has_readwrite_access(self):
+        return self.is_admin
+
+    def save(self, **kwargs):
+        if self.has_readwrite_access():
+            self.access_level = self.READWRITE
+        else:
+            self.access_level = self.READONLY
```
andyhd: I don't think the `access_level` should be changed in `save` - this should instead be something like
```python
if self.has_readwrite_access() or self.user_is_admin():
    # save?
else:
    # don't save?
```

control_panel_api/serializers.py
andyhd: I think we need to keep this for flexibility - we may want users who are not admins to be granted write access - possibly

control_panel_api/models.py
andyhd: I think maybe this method doesn't need to be overridden

xoen: I see what you mean about the template method pattern. I still think the access to S3 bucket would be determined by the `access_level` field - I don't remember adding this `had_readwrite_access` method but doesn't really matter - I imagine implementation is the same for both `AppS3Bucket` and `UserS3Bucket`:

```python
    def has_readwrite_access(self):
        return self.access_level == self.READWRITE
```

control_panel_api/models.py
xoen: Yeah I see, I'm a bit confused about that ticket to be honest.

control_panel_api/models.py
jmoz: Well it's on the USERS3Bucket so this isn't an issue to me.  @xoen ?

control_panel_api/models.py
jmoz: It needs to be saved either way. This is the level we can toggle on the is_admin.  Not sure I understand what you're implying.

jmoz:
Annihilating this branch.

PR #59, opened 2017-11-03
On S3Bucket POST create a UserS3Bucket

jmoz:
## What

An associated UserS3Bucket shuold be created when an S3Bucket is created with `is_admin=True` and `access_level="readwrite"`.

## How to review

1. ./manage.py test

https://trello.com/c/WuFheSAR/447-confirm-users-assigned-s3-bucket-policy-is-read-only-2


xoen:
👍

xoen:
Actually I just realised we create the `UserS3Bucket` record in the DB but that doesn't effect AWS (I may be wrong, let's chat about this offline)

PR #60, opened 2017-11-03
User serializer changes

jmoz:
## What

Nested related data is now exposed on User for the frontend.  Create custom explicit serializers that hide the nested User field to prevent data duplication, nesting and recursion.

Refactored tests slightly for a shared assertion.

## How to review

1. ./manage.py test
2. Make a request using swagger.

https://trello.com/c/GjaFQjvn/463-user-needs-the-s3bucket-nested-data-similiar-to-app-and-s3bucket-work-1

### DIFF
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -56,8 +56,7 @@ class UserS3BucketSerializer(serializers.ModelSerializer):
         return super().update(instance, validated_data)
 
 
-class SimpleAppSerializer(serializers.ModelSerializer):
-
+class AppSimpleSerializer(serializers.ModelSerializer):
     class Meta:
         model = App
         fields = (
@@ -71,25 +70,24 @@ class SimpleAppSerializer(serializers.ModelSerializer):
         )
 
 
-class SimpleS3BucketSerializer(serializers.ModelSerializer):
-
+class S3BucketSimpleSerializer(serializers.ModelSerializer):
     class Meta:
         model = S3Bucket
         fields = ('id', 'url', 'name', 'arn', 'created_by')
 
 
-class AppAppS3BucketSerializer(serializers.ModelSerializer):
+class AppS3BucketNestedInAppSerializer(serializers.ModelSerializer):
     """Used from within with AppSerializer to not expose app"""
-    s3bucket = SimpleS3BucketSerializer()
+    s3bucket = S3BucketSimpleSerializer()
 
     class Meta:
         model = AppS3Bucket
         fields = ('id', 'url', 's3bucket', 'access_level')
 
 
-class S3BucketAppS3BucketSerializer(serializers.ModelSerializer):
+class AppS3BucketNestedInS3BucketSerializer(serializers.ModelSerializer):
     """Used from within with S3BucketSerializer to not expose s3bucket"""
-    app = SimpleAppSerializer()
+    app = AppSimpleSerializer()
 
     class Meta:
         model = AppS3Bucket
@@ -97,8 +95,7 @@ class S3BucketAppS3BucketSerializer(serializers.ModelSerializer):
 
 
 class AppSerializer(serializers.ModelSerializer):
-
-    apps3buckets = AppAppS3BucketSerializer(many=True, read_only=True)
+    apps3buckets = AppS3BucketNestedInAppSerializer(many=True, read_only=True)
 
     class Meta:
         model = App
@@ -120,7 +117,8 @@ class AppSerializer(serializers.ModelSerializer):
 
 
 class S3BucketSerializer(serializers.ModelSerializer):
-    apps3buckets = S3BucketAppS3BucketSerializer(many=True, read_only=True)
+    apps3buckets = AppS3BucketNestedInS3BucketSerializer(
+        many=True, read_only=True)
 
     class Meta:
         model = S3Bucket
@@ -147,22 +145,22 @@ class UserAppSerializer(serializers.ModelSerializer):
         return super().update(instance, validated_data)
 
 
-class UserUserAppSerializer(serializers.ModelSerializer):
+class UserAppNestedInUserSerializer(serializers.ModelSerializer):
     """Used from within with UserSerializer to explicitly expose the app
     but hide the User
     """
-    app = SimpleAppSerializer()
+    app = AppSimpleSerializer()
 
     class Meta:
         model = UserApp
         fields = ('id', 'app', 'is_admin')
 
 
-class UserUserS3BucketSerializer(serializers.ModelSerializer):
+class UserS3BucketNestedInUserSerializer(serializers.ModelSerializer):
     """Used from within with UserSerializer to explicitly expose the s3bucket
     but hide the User
     """
-    s3bucket = SimpleS3BucketSerializer()
+    s3bucket = S3BucketSimpleSerializer()
 
     class Meta:
         model = UserS3Bucket
@@ -170,8 +168,9 @@ class UserUserS3BucketSerializer(serializers.ModelSerializer):
 
 
 class UserSerializer(serializers.ModelSerializer):
-    userapps = UserUserAppSerializer(many=True, read_only=True)
-    users3buckets = UserUserS3BucketSerializer(many=True, read_only=True)
+    userapps = UserAppNestedInUserSerializer(many=True, read_only=True)
+    users3buckets = UserS3BucketNestedInUserSerializer(
+        many=True, read_only=True)
 
     class Meta:
         model = User
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -144,7 +144,7 @@ class S3BucketSerializer(serializers.ModelSerializer):
         read_only_fields = ('apps3buckets', 'created_by')
 
 
-class AppUserSerializer(serializers.ModelSerializer):
+class UserAppSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = UserApp
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -11,22 +11,6 @@ from control_panel_api.models import (
 )
 
 
-class UserSerializer(serializers.ModelSerializer):
-
-    class Meta:
-        model = User
-        fields = (
-            'auth0_id',
-            'url',
-            'username',
-            'name',
-            'email',
-            'groups',
-            'userapps',
-            'users3buckets'
-        )
-
-
 class GroupSerializer(serializers.ModelSerializer):
 
     class Meta:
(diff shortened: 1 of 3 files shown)


control_panel_api/serializers.py
```diff
@@ -161,3 +145,43 @@ def update(self, instance, validated_data):
             )
 
         return super().update(instance, validated_data)
+
+
+class UserUserAppSerializer(serializers.ModelSerializer):
```
andyhd: This class name is not very self-documenting. Should it be something like `NestedUserAppSerializer`?

control_panel_api/serializers.py
```diff
@@ -161,3 +145,43 @@ def update(self, instance, validated_data):
...
+    but hide the User
+    """
+    app = SimpleAppSerializer()
+
+    class Meta:
+        model = UserApp
+        fields = ('id', 'app', 'is_admin')
+
+
+class UserUserS3BucketSerializer(serializers.ModelSerializer):
```
andyhd: This class name is not obvious

control_panel_api/tests/test_views.py
```diff
@@ -20,6 +20,26 @@
 )
 
 
+class AssertionsMixin(object):
+    def assert_s3bucket(self, s3bucket):
+        self.assertIn('id', s3bucket)
+        self.assertIn('url', s3bucket)
+        self.assertIn('name', s3bucket)
+        self.assertIn('arn', s3bucket)
+        self.assertIn('created_by', s3bucket)
```
andyhd: This would be more readable as `self.assert('id' in s3bucket)`.

But, is there a way to assert all these fields are present at once? Perhaps using `Set.issubset`?

control_panel_api/tests/test_views.py
```diff
@@ -20,6 +20,26 @@
...
+        self.assertEqual(5, len(s3bucket))
```
andyhd: With this assertion, the previous ones add up to something like `self.assertEqual(s3bucket.field_names, expected_field_names)` - can it be reduced to a single assertion?

control_panel_api/tests/test_views.py
```diff
@@ -20,6 +20,26 @@
 )
 
 
+class AssertionsMixin(object):
+    def assert_s3bucket(self, s3bucket):
```
andyhd: These assertions are quite brittle - if the models change, they will fail, but the tested object may still be a s3bucket or an app. Are the facts that these methods assert relevant?

control_panel_api/serializers.py
jmoz: I did originally name it Nested* but then I have another one so can't call it Nested2* so I prefixed the name of the Serializer that owns it.

So UserUserAppSerializer means the serializer for UserApps on User.  If that makes sense.

control_panel_api/tests/test_views.py
jmoz: We could use a set.

I disagree on the usage of self.assert.  self.assertIn literally has it in the name of the method, it's intention is clear and is standard practice.  It also goes against previous advice you have given in a PR to use the correct test class assertions (if my memory serves me).

control_panel_api/tests/test_views.py
jmoz: That doesn't read as clear to me.

control_panel_api/serializers.py
jmoz: Look at the similiar App and S3Bucket stuff how I've done that.  Remember the intention is to explicitly expose certain nested fields, which differ depending on the owning serializer.  

Agree it's not that clear, but I'm not sure what else.  I know @xoen is the naming God so I'd like your input.

control_panel_api/serializers.py
andyhd: 👍

control_panel_api/tests/test_views.py
andyhd: Ok, you're right about `assertIn` - it's just a bit weird to read "assert in needle haystack" instead of "assert needle in haystack".

control_panel_api/tests/test_views.py
andyhd: Just minor advantages - it would be more concise (fewer lines to read and fewer places for bugs to hide) and it would reveal the intent (assert that the object has only the fields we expect)

andyhd: What I mean about revealing the intent is, this method _says_ it asserts that the passed object is a `s3bucket`, but what it _does_ is assert that the object has only the five fields we specify.

The intention of the tests that call it appears to be to verify that the JSON objects in the response represent an `S3Bucket` model - maybe the assertion should be that deserializing the JSON object to an instance of S3Bucket succeeds?

control_panel_api/tests/test_views.py

control_panel_api/serializers.py
xoen: As showed offline, I was playing with the names of the serializers and I came up with something which may be a bit more explicit.

Here the rules:
1. Serialiser class start with name of the model serialised (e.g. `App...` is serialising an `App`, `AppS3Bucket...` is serialising an `AppS3Bucket`
2. If the serialiser is not serialising relationships we prefix it with `...Simple` (e.g. `AppSimple...` is serializing an `App` without associations, etc...)
3. If the serialiser is going to be used/nested in another serialiser will have prefix `...NestedIn${ParentSerializerName}` (e.g. `AppS3BucketNestedInAppSerializer` is the serialiser for `AppS3Bucket` in the context of an `App`, `AppS3BucketNestedInS3BucketSerializer` is the serialiser for `AppS3Bucket` in the context of an `S3Bucket`)

Examples:
* `UserUserAppSerializer` => `UserAppNestedInAppSerializer`
* `UserUserS3BucketSerializer` => `UserS3BucketNestedInUserSerializer`
* `SimpleS3BucketSerializer` => `S3BucketSimpleSerializer`
* etc...

Benefits:
* Explicit, the name of the class tells you where the serialiser will be nested/used.
* Class names start with the object being serialized

control_panel_api/serializers.py
xoen: @jmoz see other comment with rules I used earlier when I was playing with the serializers class names. They seem to make sense to you earlier but it's just an idea.

control_panel_api/serializers.py
```diff
@@ -161,3 +145,43 @@ def update(self, instance, validated_data):
...
+    """
+    s3bucket = SimpleS3BucketSerializer()
+
+    class Meta:
+        model = UserS3Bucket
+        fields = ('id', 's3bucket', 'access_level', 'is_admin')
+
+
+class UserSerializer(serializers.ModelSerializer):
+    userapps = UserUserAppSerializer(many=True, read_only=True)
```
xoen: Using possible new names this would become:

```python
userapps = UserAppNestedInUserSerializer(many=True, read_only=True)
users3buckets = UserS3BucketNestedInUserSerializer(many=True, read_only=True)
```

control_panel_api/serializers.py
andyhd: I like it, except for `...Simple` - that seems unnecessary

jmoz: It's called simple as it just exposes some of the top level S3Bucket fields, as opposed to the normal S3BucketSerializer which also exposes the many to many and also displays the m2m as a nested representation.

xoen: Alternatives to `...Simple`:
 * `...NoAssociations`

xoen: @jmoz so basically it's the serialisation of the core fields but it doesn't contain the associations (as the purpose of these "simple" serializers is to be used in the context of the parents associations so they're essentially "leaves" in the JSON response "tree").

The more I think about the `...NoAssociations` name the more I like it as it's quite explicit.

jmoz: AppFLATSerializer!?

control_panel_api/tests/test_views.py
jmoz: Well my intention is just to assert the correct fields I would expect back when serializing the app or bucket.  So something like `assert_s3bucket_representation` etc etc?

I do not care about the data in these cases, as that would make the test brittle.  I just want to know that the correct fields have been returned due to serializer usage.

jmoz: Regarding a single assertion it could look like this:

```
expected_fields = {
    'id',
    'url',
    'name',
    'arn',
    'created_by'
}
self.assertEqual(expected_fields, set(s3bucket))
```

So the question is which is more readable?  @xoen

control_panel_api/tests/test_views.py
```diff
@@ -54,6 +76,22 @@ def test_detail(self):
...
+
+        self.assert_app(userapp['app'])
+
+        users3bucket = response.data['users3buckets'][0]
+        self.assertIn('id', users3bucket)
+        self.assertIn('s3bucket', users3bucket)
+        self.assertIn('access_level', users3bucket)
+        self.assertEqual(4, len(users3bucket))
+
+        self.assert_s3bucket(users3bucket['s3bucket'])
```
xoen: I wonder if this `assert_s3bucket` could get the `s3bucket` from the response and the expected s3bucket and check they much ... or may be too much?

jmoz: Yeh bit much, I'm just checking keys and have refactored common code.

control_panel_api/tests/test_views.py
xoen: I wouldn't mind, as far as I can tell they both read well enough to me...(but I see Andy's point about the data).

andyhd: It's more the list of hard-coded field names that bothers me - you will have to edit both the model and the tests when you make changes to the fields.

This method would better match your stated intention if it checked that the JSON is a serialized `S3Bucket`, maybe by deserializing it, or by comparing the field names to a freshly serialized one, or something else.

xoen: Mmmm, I don't know if it's a good idea to get this list implicitly from the model. I'd prefer to be explicit at the small cost of having to add/remove an item from the list.

Plus, we (for whatever reason) may want to hide some of the fields in the model.

Just my 2p.

jmoz: Had a rethink of the serializer field testing - totally forgot about the `Serializer.Meta.fields` var.  I refactored all the tests to use a nice, explicit field assertion pattern.  I also refactored a couple tests that can be included in existing test methods.  It also eliminated the `AssertionsMixin`.  I've pushed up a commit which I'm happy with.

jmoz: I've played with using the Serializer.Meta.fields var, which works, but Aldo is right, the fields should be specified and tested because we should use the test as a contract for the representation - if someone changes the serializer fields accidentally we need to see the test blow up.  If you reduce it to dynamic one liners the tests will always pass, I could just delete a load of important fields and we wouldn't know.

xoen:
There may be some places where the response is not including some of the associations (e.g. `app.userapps`) but that can be addressed independently.

Is this good to go?

jmoz:
@xoen happy with the intention. Refactor as needed.

PR #61, opened 2017-11-06
Fix repo name and add test to cover bug

jmoz:
## What

Fixes bug with empty string being returned when url has trailing slash.

Fix tests to use mommy so valid repo urls are created during test

## How to review

1. Run tests

https://trello.com/c/aQbbFCqu/464-cp-api-handle-trailing-slash-in-apprepourl-1

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -78,11 +78,15 @@ class App(TimeStampedModel):
         "https://github.com/org/a_repo_name" => "a_repo_name"
         "git@github.com:org/repo_2.git" => "repo_2"
         """
-        last_after_slash = self.repo_url \
-            .rstrip('/') \
-            .replace('.git', '') \
-            .rsplit('/', 1)[1]
-        return last_after_slash
+        repo_name = self.repo_url
+
+        if repo_name.endswith('/'):
+            repo_name = repo_name[:-1]
+
+        if repo_name.endswith('.git'):
+            repo_name = repo_name[:-4]
+
+        return repo_name.rsplit('/', 1)[1]
 
 
 class UserApp(TimeStampedModel):
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -68,8 +68,7 @@ class App(TimeStampedModel):
 
     @property
     def _repo_name(self):
-        '''
-        Returns the repo name
+        """Returns the repo name
 
         The name is the part after the last slash in the URL, without
         the '.git' (if present).
@@ -78,10 +77,12 @@ class App(TimeStampedModel):
 
         "https://github.com/org/a_repo_name" => "a_repo_name"
         "git@github.com:org/repo_2.git" => "repo_2"
-        '''
-
-        last_after_slash = self.repo_url.split('/')[-1]
-        return last_after_slash.replace('.git', '')
+        """
+        last_after_slash = self.repo_url \
+            .rstrip('/') \
+            .replace('.git', '') \
+            .rsplit('/', 1)[1]
+        return last_after_slash
 
 
 class UserApp(TimeStampedModel):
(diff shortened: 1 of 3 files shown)


control_panel_api/models.py
```diff
@@ -78,10 +77,12 @@ def _repo_name(self):
...
-        '''
-
-        last_after_slash = self.repo_url.split('/')[-1]
-        return last_after_slash.replace('.git', '')
+        """
+        last_after_slash = self.repo_url \
+            .rstrip('/') \
+            .replace('.git', '') \
+            .rsplit('/', 1)[1]
+        return last_after_slash
```
andyhd: This code is not very readable, especially with the line continuation slashes and method chaining.
It would be more readable like this:
```python
url = self.repo_url

if url.endswith('/'):
    url = url[:-1]

if url.endswith('.git'):
    url = url[:-4]

return url.rsplit('/', 1)
```

control_panel_api/tests/test_models.py
```diff
@@ -262,6 +260,25 @@ def test_aws_delete(self, mock_detach_bucket_access_from_app_role):
...
+        self.assertEqual('repo_2', app._repo_name)
+
+        app.repo_url = 'https://github.com/org/a_repo_name/'
+        self.assertEqual('a_repo_name', app._repo_name)
+
+        app.repo_url = 'http://foo.com'
+        self.assertEqual('foo.com', app._repo_name)
+
+        app.repo_url = 'http://foo.com/'
+        self.assertEqual('foo.com', app._repo_name)
```
andyhd: These test cases should be in a list and tested with something like:
```python
url_test_cases = [
    ('git@github.com:org/a_repo_name', 'a_repo_name'),
    ...
]
for url, expected in url_test_cases:
    app.repo_url = url
    self.assertEqual(expected, app._repo_name)
```

jmoz: 👍

control_panel_api/models.py

control_panel_api/tests/test_models.py
```diff
@@ -262,6 +260,21 @@ def test_aws_delete(self, mock_detach_bucket_access_from_app_role):
             self.app_1.iam_role_name
         )
 
+    def test_repo_name(self):
+        app = mommy.prepare('control_panel_api.App')
+
+        url_test_cases = (
```
xoen: nice :)

control_panel_api/tests/test_services.py
```diff
@@ -78,8 +78,9 @@ def test_delete_bucket_policies(self, mock_detach_policy_from_entities,
 
     @patch('control_panel_api.aws.attach_policy_to_role')
     def test_apps3bucket_create(self, mock_attach_policy_to_role):
-        app = App.objects.create(slug='appslug')
-        s3bucket = S3Bucket.objects.create(name='test-bucketname')
+        app = mommy.make('control_panel_api.App', slug='appslug')
```
xoen: Not sure what model mommy gives us in simple cases like these, but don't mind.

jmoz: Because manually creating you would have had to create certain fields that followed rules for url or whatever.  mommy does it all behind the scenes.  In this instance the tests were all failing because of invalid repo_urls etc.

control_panel_api/tests/test_models.py
```diff
@@ -262,6 +260,21 @@ def test_aws_delete(self, mock_detach_bucket_access_from_app_role):
...
+            ('https://github.com/org/a_repo_name', 'a_repo_name'),
+            ('git@github.com:org/repo_2.git', 'repo_2'),
+            ('https://github.com/org/a_repo_name/', 'a_repo_name'),
+            ('http://foo.com', 'foo.com'),
+            ('http://foo.com/', 'foo.com'),
```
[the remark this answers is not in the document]
jmoz: 👍

control_panel_api/tests/test_services.py
xoen: Makes sense. Thanks for the explanation James.

PR #63, opened 2017-11-10
Expose S3Bucket's users3buckets as in the other serializers

xoen:
### What
This is needed by the UI to filter out (in the form) users that
already have access to this S3 bucket.

### Why
This is required by the UI:
 - UI needs to show users that already have access to S3 bucket (`users3buckets` association of the S3 bucket)
- UI displays a form with list of users who *do not* already have access to this S3 bucket. In order to generate this list we need to know which users already have access to it (again this is in the `users3buckets` association).

### Part of ticket
https://trello.com/c/hgdqE4dG/471-cp-api-present-s3bucketusers3buckets-as-in-other-endpoints-4


jmoz:
LGTM excellent addition 🥇

PR #66, opened 2017-11-20
Allow disabling writes to cluster with an environment variable

andyhd:
## What

* Added setting dictionary `ENABLED` for feature flags
* Added `write_to_cluster` feature flag, taking value from `ENABLE_WRITE_TO_CLUSTER` environment variable (default `True`), to allow disabling AWS calls and Helm commands in development (or for other reasons)
* Refactored AWS and Helm client modules and calling code to honour feature flag
* Refactored tests to mock AWS and Helm clients

## How to review

1. Set `ENABLE_WRITE_TO_CLUSTER` to a falsey value (eg: `False` or `0`)
2. `docker-compose up`
3. Create a user:
    ```shell
    docker-compose exec app python3 manage.py shell
    ```
    then
    ```python
    user = User.objects.create(username='test-user', auth0_id='github|12345')
    user.save()
    ```
4. See that the user is not created in the cluster and no AWS API calls are made

See [Trello #476](https://trello.com/c/ECksDteJ)

### DIFF
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -1,106 +1,109 @@
 import json
 
+import boto3
+
 from django.conf import settings
 
 
-aws_api_client = settings.AWS_API_CLIENT_HANDLER
+class AWSClient(object):
+
+    def __init__(self):
+        self.enabled = settings.ENABLED['write_to_cluster']
+        self.client = boto3.client
+
+    def _do(self, service, method, **kwargs):
+        if self.enabled:
+            client = self.client(service)
+            return getattr(client, method)(**kwargs)
+
+    def create_bucket(self, name, region, acl='private'):
+        self._do('s3', 'create_bucket',
+            Bucket=name,
+            ACL=acl,
+            CreateBucketConfiguration={'LocationConstraint': region})
+
+    def put_bucket_logging(self, name, target_bucket, target_prefix):
+        self._do('s3', 'put_bucket_logging',
+            Bucket=name,
+            BucketLoggingStatus={
+                'LoggingEnabled': {
+                    'TargetBucket': target_bucket,
+                    'TargetPrefix': target_prefix
+                }
+            })
+
+    def create_policy(self, name, policy_document):
+        self._do('iam', 'create_policy',
+            PolicyName=name,
+            PolicyDocument=json.dumps(policy_document))
+
+    def delete_policy(self, policy_arn):
+        self._do('iam', 'delete_policy', PolicyArn=policy_arn)
+
+    def create_role(self, role_name, assume_role_policy):
+        """Creates IAM role with the given name"""
+
+        self._do('iam', 'create_role',
+            RoleName=role_name,
+            AssumeRolePolicyDocument=json.dumps(assume_role_policy))
+
+    def delete_role(self, role_name):
+        """Delete the given IAM role."""
+
+        self._detach_role_policies(role_name)
+        self._do('iam', 'delete_role', RoleName=role_name)
+
+    def _detach_role_policies(self, role_name):
+        """Detaches all the policies from the given role"""
+
+        policies = self._do(
+            'iam', 'list_attached_role_policies', RoleName=role_name)
+
+        if policies:
+            for policy in policies["AttachedPolicies"]:
+                self.detach_policy_from_role(
+                    role_name=role_name,
+                    policy_arn=policy["PolicyArn"]
+                )
+
+    def detach_policy_from_entities(self, policy_arn):
+        """
+        Get all entities to which policy is attached first then call separate
+        detach operations
+        """
+        entities = self._do(
+            'iam', 'list_entities_for_policy', PolicyArn=policy_arn)
+
+        if entities:
+
+            for role in entities["PolicyRoles"]:
+                self.detach_policy_from_role(policy_arn, role["RoleName"])
+
+            for group in entities["PolicyGroups"]:
+                self.detach_policy_from_group(policy_arn, group["GroupName"])
+
+            for user in entities["PolicyUsers"]:
+                self.detach_policy_from_user(policy_arn, user["UserName"])
+
+    def attach_policy_to_role(self, policy_arn, role_name):
+        self._do('iam', 'attach_role_policy',
+            RoleName=role_name,
+            PolicyArn=policy_arn)
+
+    def detach_policy_from_role(self, policy_arn, role_name):
+        self._do('iam', 'detach_role_policy',
+            RoleName=role_name,
+            PolicyArn=policy_arn)
+
+    def detach_policy_from_group(self, policy_arn, group_name):
+        self._do('iam', 'detach_group_policy',
+            GroupName=group_name,
+            PolicyArn=policy_arn)
+
+    def detach_policy_from_user(self, policy_arn, user_name):
+        self._do('iam', 'detach_user_policy',
+            UserName=user_name,
+            PolicyArn=policy_arn)
 
 
-def create_bucket(name, region, acl='private'):
-    aws_api_client('s3').create_bucket(
-        Bucket=name,
-        ACL=acl,
-        CreateBucketConfiguration={'LocationConstraint': region},
-    )
-
-
-def put_bucket_logging(name, target_bucket, target_prefix):
-    aws_api_client('s3').put_bucket_logging(
-        Bucket=name,
-        BucketLoggingStatus={
-            'LoggingEnabled': {
-                'TargetBucket': target_bucket,
-                'TargetPrefix': target_prefix
-            }
-        }
-    )
-
-
-def create_policy(name, policy_document):
-    aws_api_client("iam").create_policy(
-        PolicyName=name,
-        PolicyDocument=json.dumps(policy_document)
-    )
-
-
-def delete_policy(policy_arn):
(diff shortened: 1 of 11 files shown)


control_panel_api/settings/base.py
```diff
@@ -1,6 +1,15 @@
 import os
 
-import boto3
+
+def is_enabled(value):
+    return str(value).lower() not in ('n', 'no', 'off', 'false', '0')
```
xoen: I wonder if we should keep it simple and consider a flag enabled only when equal to string `'true'`.

```python
def is_enabled(value):
  return str(value).lower() == 'true'
```

andyhd: That seems unnecessary as the code is already simple and is more flexible that what you suggest.

control_panel_api/aws.py
```diff
@@ -1,106 +1,109 @@
...
 
-def detach_policy_from_entities(policy_arn):
-    """Get all entities to which policy is attached first then call separate detach operations"""
-    entities = aws_api_client("iam").list_entities_for_policy(PolicyArn=policy_arn)
+        policies = self._do(
+            'iam', 'list_attached_role_policies', RoleName=role_name)
 
-    for role in entities["PolicyRoles"]:
-        detach_policy_from_role(policy_arn, role["RoleName"])
+        if policies:
```
xoen: Is this `if` statement necessary? Clearly if the list is empty the `for` loop would be a no-op.

control_panel_api/aws.py
```diff
@@ -1,106 +1,109 @@
...
-        PolicyName=name,
-        PolicyDocument=json.dumps(policy_document)
-    )
 
+class AWSClient(object):
 
-def delete_policy(policy_arn):
-    aws_api_client("iam").delete_policy(PolicyArn=policy_arn)
+    def __init__(self):
+        self.enabled = settings.ENABLED['write_to_cluster']
```
xoen: Super-minor thing. I wonder if the constructor should receive this `enabled` option instead of reading a global variable directly (`settings`).

control_panel_api/services.py
```diff
@@ -1,6 +1,6 @@
 from django.conf import settings
 
-from . import aws
+from .aws import aws_api_client as aws
```
xoen: I love how as the name of the methods didn't change this means this is the only change in the file 👍

control_panel_api/aws.py
andyhd: It is necessary, because `policies` is a dict. The `for` loops through `policies["AttachedPolicies"]` - accessing the `"AttachedPolicies"` item will raise a `KeyError` if `policies` is `None`.

control_panel_api/aws.py
andyhd: Yep, maybe. Not really necessary, as `aws_api_client` is essentially a singleton - no other instances will be created.

control_panel_api/tests/test_aws.py
```diff
@@ -1,88 +1,87 @@
...
-from unittest.mock import call
+from unittest.mock import MagicMock, call, patch
 
-from control_panel_api import aws
-from control_panel_api.aws import aws_api_client
+from .aws_mock import AWSClientPatcher, aws
 
 
-class AwsTestCase(TestCase):
+class AwsTestCase(AWSClientPatcher, TestCase):
```
jmoz: Personally I would favour a class level `@patch` - it's simpler, cleaner and more obvious.

control_panel_api/settings/base.py
```diff
@@ -1,6 +1,15 @@
...
+
+
+# These are boolean flags to enable/disable features in the API
+ENABLED = {
+    'write_to_cluster':
```
jmoz: Shouldn't we have more granular control over which libraries are enabled?

control_panel_api/settings/test.py
```diff
@@ -1,9 +1,7 @@
 from .base import *
-from unittest.mock import MagicMock
 
 
 ENV = 'test'
-AWS_API_CLIENT_HANDLER = MagicMock()
```
jmoz: 👍

control_panel_api/tests/test_views.py
```diff
@@ -103,17 +107,18 @@ def test_delete(self, mock_aws_delete_role):
...
+    def test_create(self, mock_aws_create_role):
+        username = 'foo'
+        data = {'auth0_id': 'github|2', 'username': username}
         response = self.client.post(reverse('user-list'), data)
         self.assertEqual(HTTP_201_CREATED, response.status_code)
 
         self.assertEqual(data['auth0_id'], response.data['auth0_id'])
 
         mock_aws_create_role.assert_called()
-        mock_helm_create_user.assert_called()
```
jmoz: The view layer tests are asserting that model methods are called (which have no params) so need to remain.  The model layer tests should be asserting the correct params.

control_panel_api/tests/test_views.py
```diff
@@ -11,24 +11,28 @@
...
 
+from .aws_mock import AWSClientPatcher, aws
 from control_panel_api.models import (
     AppS3Bucket,
     S3Bucket,
     User,
     UserApp,
     UserS3Bucket,
 )
+from .helm_mock import HelmClientPatcher, helm
```
jmoz: The naming is now confusing as to what is real or mocked.

control_panel_api/aws.py
```diff
@@ -1,106 +1,109 @@
...
 
+from django.conf import settings
 
-def create_policy(name, policy_document):
-    aws_api_client("iam").create_policy(
-        PolicyName=name,
-        PolicyDocument=json.dumps(policy_document)
-    )
 
+class AWSClient(object):
```
jmoz: I like the refactor to use a client object.

I am somewhat ambivalent towards the usage of feature flags due to the peppering of `if` statements in client objects or libraries.  Have you thought about using polymorphism at runtime to replace with `AWS|HelmDummyClient` implementations which would keep the client code interface the same and cleaner than feature flags?  e.g https://docs.djangoproject.com/en/1.11/topics/cache/#dummy-caching-for-development

control_panel_api/helm.py
```diff
@@ -3,28 +3,37 @@
...
+    def upgrade_release(self, release, chart, *args):
+        default_flags = ['--install', '--wait']
+        flags = list(args) + default_flags
+        self._helm_command('upgrade', release, chart, *flags)
+
+    def init_user(self, username, email, fullname):
+        self.upgrade_release(
+            f'init-user-{username}',
+            'mojanalytics/init-user',
+            '--set', f'NFSHostname={settings.NFS_HOSTNAME}',
```
jmoz: Now we have a client object, could we extract the django settings stuff and use dependency injection?

control_panel_api/settings/base.py
andyhd: These are flags for features, not libraries. The feature set being enabled/disabled here is writing changes to the cluster. Having to set multiple flags for a single concept would be confusing and prone to errors.

control_panel_api/tests/test_views.py
andyhd: Yep, I'll rename the module

control_panel_api/aws.py
andyhd: Yep, that does look like a good approach for this case. Feature flags are desirable for other cases, however. I think I will refactor to this approach for this case later, because I don't want Clive to have to wait.

control_panel_api/helm.py
andyhd: Yep, good idea

control_panel_api/tests/test_views.py
andyhd: True, I may have gone overboard with the test updates. I'll revert.

control_panel_api/tests/test_aws.py
andyhd: I sort-of agree. A class level patch would not be as clean for the helm patches. But you're right, the `Patcher` classes are not really right. I'll redo them as class patches.

control_panel_api/settings/base.py
andyhd: We might need to revisit if we have overlapping feature flags, but hopefully that would be rare as we'd remove old flags when they become unnecessary.

control_panel_api/tests/test_aws.py
jmoz: 👍

control_panel_api/settings/base.py
jmoz: 👍

control_panel_api/tests/test_views.py

control_panel_api/aws.py
jmoz: 👍

control_panel_api/tests/test_aws.py
andyhd: I have tried, but I cannot get a class level patch to work. It also clutters the test method signatures, adding the patched object as a parameter.

control_panel_api/tests/test_views.py
andyhd: I've removed them completely and refactored to use class decorators.

control_panel_api/helm.py
andyhd: Hm, I'm not so sure about this one actually. It would be better if we could remove this NFS_HOSTNAME variable from the API settings altogether. I think I'll leave it for now.

PR #67, opened 2017-11-20
Handle aws and helm errors and add transaction support

jmoz:
## What

- Add explicit transaction support on view functions.
- AWS and helm explicit error handling - let the lower level exceptions bubble up to the view, log, then throw a subclassed APIView exception.
- For 2 stage view operations e.g. aws call then helm call, if it fails on the 2nd helm call, we can revert the aws call.

## How to review

1. Run the tests.
2. Rename the helm command for an obvious error on User create and check that the aws call is rolled back and no db item is saved.

https://trello.com/c/GZphN2qR/479-db-transactions-aws-helm-error-handling-3

### DIFF
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -15,13 +15,16 @@ def ignore_existing(func):
     """Decorates a function to catch and allow exceptions that are thrown for
     existing entities or already created buckets etc, and reraise all others
     """
-    e_names = 'BucketAlreadyOwnedByYou', 'EntityAlreadyExistsException'
+    exception_names = (
+        'BucketAlreadyOwnedByYou',
+        'EntityAlreadyExistsException'
+    )
 
     def inner(*args, **kwargs):
         try:
             func(*args, **kwargs)
         except ClientError as e:
-            if e.__class__.__name__ not in e_names:
+            if e.__class__.__name__ not in exception_names:
                 raise e
 
             logger.error(f"Caught aws exception and ignored: {e}")
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -1,9 +1,32 @@
+import logging
+
+from botocore.exceptions import ClientError
 from django.conf import settings
 
 from .aws import aws
 
 READWRITE = 'readwrite'
 READONLY = 'readonly'
+
+logger = logging.getLogger(__name__)
+
+
+def ignore_existing(func):
+    """Decorates a function to catch and allow exceptions that are thrown for
+    existing entities or already created buckets etc, and reraise all others
+    """
+    e_names = 'BucketAlreadyOwnedByYou', 'EntityAlreadyExistsException'
+
+    def inner(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except ClientError as e:
+            if e.__class__.__name__ not in e_names:
+                raise e
+
+            logger.error(f"Caught aws exception and ignored: {e}")
+
+    return inner
 
 
 def _policy_name(bucket_name, readwrite=False):
@@ -24,6 +47,7 @@ def _policy_arn(bucket_name, readwrite=False):
         _policy_name(bucket_name, readwrite))
 
 
+@ignore_existing
 def create_role(role_name, add_saml_statement=False):
     """See: `sts:AssumeRole` required by kube2iam
     https://github.com/jtblin/kube2iam#iam-roles"""
@@ -127,6 +151,7 @@ def get_policy_document(bucket_name_arn, readwrite):
     }
 
 
+@ignore_existing
 def create_bucket(bucket_name):
     aws.create_bucket(
         bucket_name,
@@ -138,6 +163,7 @@ def create_bucket(bucket_name):
         target_prefix=f"{bucket_name}/")
 
 
+@ignore_existing
 def create_bucket_policies(bucket_name, bucket_arn):
     """Create readwrite and readonly policies for s3 bucket"""
     readwrite = True
--- a/control_panel_api/tests/test_views.py
+++ b/control_panel_api/tests/test_views.py
@@ -134,11 +134,10 @@ class UserViewTest(AuthenticatedClientMixin, APITestCase):
         self.assertEqual(data['username'], response.data['username'])
         self.assertEqual(data['auth0_id'], response.data['auth0_id'])
 
-    @patch('control_panel_api.models.User.aws_delete_role')
     @patch('control_panel_api.models.User.helm_create')
     @patch('control_panel_api.models.User.aws_create_role')
     def test_aws_error_and_transaction(self, mock_aws_create_role,
-                                       mock_helm_create, mock_aws_delete_role):
+                                       mock_helm_create):
         mock_helm_create.side_effect = CalledProcessError(1, 'foo error')
 
         data = {'auth0_id': 'github|3', 'username': 'foo'}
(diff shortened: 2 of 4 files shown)


control_panel_api/views.py
```diff
@@ -22,28 +28,46 @@
...
+    @transaction.atomic
     def perform_create(self, serializer):
         instance = serializer.save()
-        instance.aws_create_role()
-        instance.helm_create()
 
+        try:
+            instance.aws_create_role()
+            instance.helm_create()
+        except ClientError as e:
```
xoen: Only problem with raising and rolling back the DB changes for all boto errors is that if we try to create a user in the CP and its role is already present it will always fail.

Ideally we'd want to ignore "this specific role already exists" exceptions and consider that as a "success".

Boto exceptions are not exactly pretty so you may need to check `e.response["Error"]["Code"]` is not a certain value.

control_panel_api/views.py
```diff
@@ -22,28 +28,46 @@
...
+            logger.error(e)
+            raise AWSException
+        except CalledProcessError as e:
+            logger.error(e)
+            instance.aws_delete_role()
```
xoen: I see that you're trying to manually "roll back" the IAM role creation here.
Only worry of doing this is that if the role was created by the lambda function and `helm upgrade` fails for whatever reason we end up deleting someone perfectly working IAM role.

I would be on the safe side and leave this out, we can always change this later if we *really need* to.

control_panel_api/views.py
```diff
@@ -59,13 +83,25 @@ class AppViewSet(viewsets.ModelViewSet):
...
     permission_classes = (AppPermissions,)
 
+    @transaction.atomic
     def perform_create(self, serializer):
         app = serializer.save(created_by=self.request.user)
-        app.aws_create_role()
 
+        try:
+            app.aws_create_role()
+        except ClientError as e:
```
xoen: As [said in previous comment](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/67/files#r152300693), we may want to ignore the error raised when "app IAM role already exists".

It's probably less of a problem in the case of app IAM roles as a part from when we deleted the DB I would imagine there shouldn't be apps' IAM roles already there.

control_panel_api/views.py
```diff
@@ -74,32 +110,65 @@ class AppS3BucketViewSet(viewsets.ModelViewSet):
 
     def perform_create(self, serializer):
```
xoen: Should this wrapped in a transaction as well?

control_panel_api/views.py
```diff
@@ -74,32 +110,65 @@ class AppS3BucketViewSet(viewsets.ModelViewSet):
...
         apps3bucket = serializer.save()
-        apps3bucket.aws_create()
+
+        try:
+            apps3bucket.aws_create()
+        except ClientError as e:
+            logger.error(e)
+            raise AWSException
 
     def perform_update(self, serializer):
```
xoen: Does this method needs to be wrapped in a transaction?

control_panel_api/views.py
```diff
@@ -74,32 +110,65 @@ class AppS3BucketViewSet(viewsets.ModelViewSet):
...
         apps3bucket = serializer.save()
-        apps3bucket.aws_update()
+
+        try:
+            apps3bucket.aws_update()
+        except ClientError as e:
+            logger.error(e)
+            raise AWSException
 
     def perform_destroy(self, instance):
```
xoen: Does this method needs to be wrapped in a transaction?

control_panel_api/views.py
```diff
@@ -108,14 +177,27 @@ class S3BucketViewSet(viewsets.ModelViewSet):
...
     permission_classes = (S3BucketPermissions,)
 
+    @transaction.atomic
     def perform_create(self, serializer):
         instance = serializer.save(created_by=self.request.user)
-        instance.aws_create()
+
+        try:
+            instance.aws_create()
+        except ClientError as e:
```
xoen: As [said in other comment](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/67/files#r152300693), we may want to be specific when catching this exception to not roll back errors caused by us adding an S3 bucket already existent in the API.

control_panel_api/tests/test_views.py
```diff
@@ -123,6 +127,23 @@ def test_update(self):
...
+    def test_aws_error_and_transaction(self, mock_aws_create_role,
+                                       mock_helm_create, mock_aws_delete_role):
+        mock_helm_create.side_effect = CalledProcessError(1, 'foo error')
+
+        data = {'auth0_id': 'github|3', 'username': 'foo'}
+        response = self.client.post(reverse('user-list'), data)
+        self.assertEqual(HTTP_500_INTERNAL_SERVER_ERROR, response.status_code)
+
+        mock_aws_create_role.assert_called()
+        mock_aws_delete_role.assert_called()
```
xoen: As said in other comment, I wonder if delete the user IAM role on helm error may cause more problems than other.

control_panel_api/views.py
jmoz: Well spotted, thanks.

control_panel_api/views.py
jmoz: 👍

control_panel_api/views.py
jmoz: 👍

control_panel_api/views.py
jmoz: Ok will delete, maybe trying to be too clever.

control_panel_api/views.py
```diff
@@ -22,28 +28,45 @@
...
         instance = serializer.save()
-        instance.aws_create_role()
-        instance.helm_create()
 
+        try:
+            instance.aws_create_role()
+            instance.helm_create()
+        except ClientError as e:
+            logger.error(e)
+            raise AWSException
```
andyhd: This should be `raise AWSException(msg) from e` to chain the exceptions: https://docs.python.org/3/reference/simple_stmts.html#raise

jmoz: mindblown.jpg
👍

jmoz: @andyhd do you want to expose the full message "An error occurred (EntityAlreadyExists) when calling the CreateRole operation: Role with name foo already exists." in the output, or provide the generic "tehre was an AWS error.." and log?

andyhd: More detail is probably better

control_panel_api/services.py
```diff
@@ -1,10 +1,33 @@
...
 READONLY = 'readonly'
 
+logger = logging.getLogger(__name__)
+
+
+def ignore_existing(func):
+    """Decorates a function to catch and allow exceptions that are thrown for
+    existing entities or already created buckets etc, and reraise all others
+    """
+    e_names = 'BucketAlreadyOwnedByYou', 'EntityAlreadyExistsException'
```
andyhd: This should be `exception_names` or maybe just `exceptions`. Also, if it's a tuple, please include the parentheses.

andyhd: I'm assuming this isn't a list of the actual exception classes because boto dynamically creates them?

control_panel_api/services.py
```diff
@@ -24,6 +47,7 @@ def _policy_arn(bucket_name, readwrite=False):
         _policy_name(bucket_name, readwrite))
 
 
+@ignore_existing
```
andyhd: 👍

control_panel_api/services.py
jmoz: 👍

control_panel_api/services.py
jmoz: @xoen 🥇

control_panel_api/services.py
jmoz: Yep sorry should have commented - because they're dynamic I have to catch the parent then check their names.

PR #71, opened 2017-11-23
Really create the user in the database and the cluster

andyhd:
## What

* When an unknown authenticated user makes a request, the user should be created in the database and the cluster, and have an IAM role attached. Now this actually happens.

## How to review

1. Make an API request as an unknown user
2. See that `init-user` and `config-user` helm charts are deployed for this user
3. You can now successfully deploy RStudio for this user

### DIFF
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -48,5 +48,8 @@ class Auth0JWTAuthentication(BaseAuthentication):
                 email=decoded.get(settings.OIDC_FIELD_EMAIL),
                 name=decoded.get(settings.OIDC_FIELD_NAME),
             )
+            user.save()
+            user.aws_create_role()
+            user.helm_create()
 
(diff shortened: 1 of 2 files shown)


control_panel_api/authentication.py
```diff
@@ -48,5 +48,8 @@ def authenticate(self, request):
                 email=decoded.get(settings.OIDC_FIELD_EMAIL),
                 name=decoded.get(settings.OIDC_FIELD_NAME),
             )
+            user.save()
```
xoen: django-newbie question: Do you need the call to `save()` if you already called the `create()`?

andyhd: apparently yes

xoen: Yeah, just tested in a console, `create()` will save to the DB so `save()` is redundant.

PR #75, opened 2017-11-29
Add `/k8s/` proxy endpoint

xoen:
### What

Added the endpoint `/k8s/...` to proxy requests to the kubernetes (k8s) cluster.

### Why

We want users (via the UI) to have visibility over their RStudio instances, restart them, etc...

### How
Requests going `/k8s/something` will be proxied to the kubernetes (same path but without `/k8s` prefix) cluster passing query params and body.
Then response coming from the cluster will be forwarded to the user preserving the response code and body.

### Permissions
Permission model is relatively simple:
1) You need to be logged in to interact with the cluster
2) If you are a super user you can perform all operations in any namespace
3) If you are a normal user you can perform all operations in your namespace

**Future**: When we'll add RBAC to kubernetes we'll pass the auth token for the user and the authorisation will probably happen in kubernetes. Double check permissions doesn't harm anyway!

### How to review

1. Visit http://localhost:8000/k8s/api/v1/namespaces (as superuser or visi)
2. You should see the JSON response with the list of namespaces
3. Profit!

### Ticket
https://trello.com/c/K6QkOhGW/517-5-add-api-method-to-forward-kubernetes-resource-requests-to-kubernetes-api


xoen:
Hold fire on reviewing this, I may some ideas to make it nicer :)

control_panel_api/k8s.py
```diff
@@ -0,0 +1,60 @@
...
+    return Request(request).make()
+
+
+class Request(object):
+
+    def __init__(self, request):
+        self.request = request
+        self._config = Config()
+
+    def make(self):
```
andyhd: A better name for this method might be `dispatch`

control_panel_api/k8s.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(self):
+        try:
+            config.load_incluster_config()
+        except ConfigException as e:
+            config.load_kube_config()
+
+        self.host = client.configuration.host
+        self.authorization = client.configuration.api_key[
+            'authorization']
```
andyhd: No need to split this line

control_panel_api/k8s.py
```diff
@@ -0,0 +1,60 @@
...
+
+def proxy(request):
+    return Request(request).make()
+
+
+class Request(object):
+
+    def __init__(self, request):
+        self.request = request
+        self._config = Config()
```
andyhd: This will load the config on every request - that seems inefficient. Shouldn't we load the config once on startup?

xoen:
I cleaned the code further and fixed the tests so feel free to have a look :)

control_panel_api/k8s.py
```diff
@@ -0,0 +1,60 @@
...
+    def url(self):
+        return f"{self._config.host}{self.path}?{self.querystring}"
+
+    @property
+    def path(self):
+        return self.request.path[4:]  # path without '/k8s' prefix
+
+    @property
+    def querystring(self):
+        return self.request.GET.urlencode()
```
andyhd: Not sure any of these values need to be properties. Does this class conform to a standard request class interface? Seems like these properties are only being used once in `.make()` - they could be variables defined in that method, or even inline values.

control_panel_api/k8s.py
andyhd: You might want to pass the config in as a constructor parameter, then you can pass a test config object instead of monkey-patching the Config class in the tests.

control_panel_api/k8s.py
xoen: Didn't notice that - must have been the auto-PEP8-thingy 👍

control_panel_api/k8s.py
xoen: Yes good point, I'll play with it any see if I can move to the settings.

control_panel_api/k8s.py
xoen: I called it `make` because usually you usually _make_ requests, but I'll change 👍

control_panel_api/k8s.py
xoen: Also, I need to check if the authentication token changes during the lifecycle of the pod running the CP-API.

control_panel_api/k8s.py
xoen: It's mainly to make the code more readable, I'll get rid of the simple properties (`method` and `url`) and inline as it should impact readability that much 👍

control_panel_api/k8s.py
xoen: FYI: Had a chat with Kerin and I should be able to cache the token and config at start 👍

control_panel_api/k8s.py
andyhd: Yeah, I get that, but `make` is a bit unspecific - usually it would mean "build"

control_panel_api/k8s.py
```diff
@@ -0,0 +1,70 @@
...
+def proxy(request):
+    return Request(request).dispatch()
+
+
+class Request(object):
+
+    def __init__(self, request):
+        self.request = request
+        self._load_config()
+
```
andyhd:
```python
@property
def config(self):
    if not self._config:
        config.load()
        self._config = config
```

xoen:
@andyhd I'm folding for now. I tried different things but none satisfies me enough and/or can write tests for it...

...this works and load the config only once, so I think for now will do and we can always iterate (also probably config loading will change a bit anyway when/if RBAC will be enabled)

Let me know if you're happy with this as it stands or you have any other suggestions.

kerin:
@xoen I've been playing with OIDC authentication, and it should be as simple as passing the JWT id_token as the auth credential. With kubectl it looks like:

`kubectl get pods --token=$JWT`

There is also [support for passing a refresh token](https://kubernetes.io/docs/admin/authentication/#option-1---oidc-authenticator) along with the JWT, but I've had trouble getting it to work. I think in our case we don't need it though, as the token refresh will be handled by the frontend app, so the token should always be valid.

kerin:
@xoen also worth knowing - under the hood kubectl uses curl, so the actual request made above looks like:

`curl -k -v -XGET  -H "Accept: application/json" -H "User-Agent: kubectl/v1.8.2 (darwin/amd64) kubernetes/bdaeafa" -H "Authorization: Bearer $JWT" https://192.168.99.101:8443/api/v1/pods`

xoen:
Cheers @kerin - I didn't that much about k8s authentication but hopefully it should be easy to tweak this to just pass over JWT token from request once we have OIDC authentication and RBAC authorisation enabled in the cluster 👍

xoen:
(rebased)

control_panel_api/tests/test_permissions.py
```diff
@@ -386,3 +386,53 @@ def test_update_as_normal_user_responds_403(self):
...
+    def test_user_cant_operate_outside_their_namespace(self, mock_request):
+        self.client.force_login(self.normal_user)
+
+        response = self.client.get('/k8s/api/v1/namespaces/user-other')
+        self.assertEqual(HTTP_403_FORBIDDEN, response.status_code)
+
+    @patch('kubernetes.client.configuration')
+    @patch('kubernetes.config.load_incluster_config')
+    @patch('requests.request')
+    def test_user_can_operate_in_their_namespace(self, mock_request, mock_load_config, mock_config):
```
jmoz: If this is just testing the perms can you patch out at a higher level or object and avoid the lower level patches?

control_panel_api/tests/test_permissions.py
```diff
@@ -386,3 +386,53 @@ def test_update_as_normal_user_responds_403(self):
...
+    @patch('requests.request')
+    def test_superuser_can_do_anything(self, mock_request, mock_load_config, mock_config):
+        self.client.force_login(self.superuser)
+
+        mock_request.return_value.status_code = 200
+
+        response = self.client.get('/k8s/something')
+        self.assertEqual(HTTP_200_OK, response.status_code)
+
+    @patch('requests.request')
```
jmoz: If this is testing a 403 does it need the patch, does it even reach the requests execution?

control_panel_api/permissions.py
```diff
@@ -20,6 +20,26 @@ def has_permission(self, request, view):
...
+class K8sPermissions(BasePermission):
+    """
+    User can operate only in his namespace (unless superuser)
+    """
+
+    def has_permission(self, request, view):
+        if not request.user:
+            return False
+
+        if is_superuser(request.user):
```
jmoz: Can you not compose via mixin `IsSuperuser`?

jmoz: Or specify on the view and add the `IsSuperuser` permission?  Rather than rewriting and adding multiple permission checks.

xoen:
(FYI: Rebase off `master` again after James' merge)

control_panel_api/tests/test_permissions.py
xoen: It slipped through Copy&Paste, well spotted, thanks :)

control_panel_api/views.py
```diff
@@ -18,6 +20,7 @@
     AppPermissions,
     S3BucketPermissions,
     UserPermissions,
+    K8sPermissions,
```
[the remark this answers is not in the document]
xoen: You mean "K" isn't after "U" 😱 ?! I'll sort it out (no pun intended)

PR #76, opened 2017-11-29
Enable normal user to update their own details

andyhd:
## What

* Added `IsSelf` permission mixin, to allow normal user to access and modify their own details
* Added `verified_email` flag to User model to record whether a user has verified their email address

## How to review

1. `docker-compose up`
2. `docker-compose exec app python3 manage.py shell` 
3. Make yourself a normal user, eg:
    ```python
    u = User.objects.get(username='your username')
    u.is_superuser = False
    u.save()
    ```
4. Browse to http://localhost:8000/users/{your_auth0_id}/
5. See that you get a response, and not an error

See [Trello #453](https://trello.com/c/REq10r8F)


control_panel_api/models.py
```diff
@@ -14,6 +14,7 @@
 class User(AbstractUser):
     auth0_id = models.CharField(max_length=128, primary_key=True)
     name = models.CharField(max_length=256, blank=True)
+    verified_email = models.BooleanField(default=False)
```
xoen: Annoying comment on this field name. Wouldn't make more sense to call this `email_verified`? That way would be clear for the name is flag to check if the email was verified instead of an email which is verified.

Also, if both fields start with `email` is clear they're very related.

...said that, you probably already wrote some FE code going with this so probably annoying to change so ignore it if it's too much hassle and you reckon is not worth it.

control_panel_api/permissions.py
```diff
@@ -19,6 +19,9 @@ class IsSuperuser(BasePermission):
     def has_permission(self, request, view):
         return is_superuser(request.user)
 
+    def has_object_permission(self, request, view, obj):
```
xoen: Why to implement this method if it will return always the same result as the `has_permission` method? What does it gives us now?

It's a genuine question because as far as I remember by reading the permissions documentation `has_permission()` is always called first anyway on member endpoints, and `has_object_permission()` is only called for member endpoints if `has_permission()` returned true.

Basically unless I'm missing something or I don't remember correctly if  `has_permissions()` returns `False` (or raise) the `has_object_permission()` is never called.

Worth re-checking the documentation tho.

control_panel_api/permissions.py
```diff
@@ -28,5 +31,23 @@ class S3BucketPermissions(IsSuperuser):
     pass
 
 
-class UserPermissions(IsSuperuser):
+class IsSelf(BasePermission):
+
+    def has_permission(self, request, view):
+        if super().has_permission(request, view):
```
xoen: Do we need to check the parent in this case? `BasePermission#has_permission()` always returns `True`, it may be a bit overkill.

control_panel_api/permissions.py
```diff
@@ -28,5 +31,23 @@ class S3BucketPermissions(IsSuperuser):
...
+            return True
+
+        if hasattr(request.user, 'auth0_id'):
```
xoen: Nice, as far I can tell this should also handle the case where the user is not even logged in so `request.user` is `None`, or something like that...am I wrong?

control_panel_api/permissions.py
```diff
@@ -28,5 +31,23 @@ class S3BucketPermissions(IsSuperuser):
...
+            return view.kwargs.get('pk') == request.user.auth0_id
+
+        return False
+
+    def has_object_permission(self, request, view, obj):
+        if super().has_object_permission(request, view, obj):
```
xoen: Again, why to check `has_object_permission()` on the parent class here?

control_panel_api/permissions.py
```diff
@@ -28,5 +31,23 @@ class S3BucketPermissions(IsSuperuser):
...
 
-class UserPermissions(IsSuperuser):
+class IsSelf(BasePermission):
+
+    def has_permission(self, request, view):
+        if super().has_permission(request, view):
+            return True
+
+        if hasattr(request.user, 'auth0_id'):
+            return view.kwargs.get('pk') == request.user.auth0_id
```
xoen: Don't this break the listing of users?

e.g. this is called on all endpoints (on the list one as well), when I'm making a request to `GET /users`, `view.kwargs.get('pk')` will probably be `None` (or something like that) therefore will not be equal to `request.user.auth0_id`.

In addition, as it's not a member endpoint (as it's not operating on a specific object) the `has_object_permission()` should not be called.

control_panel_api/views.py
```diff
@@ -36,6 +37,11 @@ class UserViewSet(viewsets.ModelViewSet):
     filter_backends = (UserFilter,)
     permission_classes = (UserPermissions,)
 
+    def get_object(self):
```
xoen: What is this for? Is it because otherwise a normal user can't see himself?

I think that's because of the `filters.UserFilter`, that's currently only filter out everything for non-superuser. If this is filtering logic, probably should be moved in `filters.UserFilter`.

xoen: Probably `UserFilter` can now be something like:

```python
    def filter_queryset(self, request, queryset, view):
        queryset = super().filter_queryset(request, queryset, view)

        if is_superuser(request.user):
            return queryset
        elif request.user:
            return queryset.get(pk=request.user.pk)
        else:
           return queryset.none()
``` 

or something like that! (you got the idea)

control_panel_api/permissions.py
andyhd: The parent will be `IsSuperuser`

xoen: This is obviously more complicated than it seems.
I can now see why you were checking `super()`. Annoyingly as far as I can tell from the DRF documentation, if any of the permissions classes in the chain returns `False` or an exception it doesn't check the rest of the classes which in our case may be handy.

control_panel_api/permissions.py
andyhd: Yes. When the user is not logged in, `request.user` is an instance of `AnonymousUser` which does not have an `auth0_id` attribute.

control_panel_api/permissions.py
andyhd: My implementation is unnecessarily complicated - I will refactor it.

control_panel_api/permissions.py
andyhd: Yep, this is unnecessary

control_panel_api/models.py
andyhd: Yep, will change

control_panel_api/permissions.py
andyhd: This is intentional. Normal users are not allowed to list users. I don't think it's useful to `GET /users` and only see yourself listed.

The `has_object_permission` will not be called, because it is not a member endpoint.

control_panel_api/filters.py
```diff
@@ -43,11 +43,27 @@ class S3BucketFilter(SuperusersOnlyFilter):
     pass
 
 
-class UserFilter(SuperusersOnlyFilter):
+class SelfOrSuperuserFilter(DjangoFilterBackend):
```
xoen: As discussed offline we can probably have this as `UserFilter` as this logic is very specific of the user filtering.

control_panel_api/permissions.py
```diff
@@ -31,23 +28,13 @@ class S3BucketPermissions(IsSuperuser):
     pass
 
 
-class IsSelf(BasePermission):
+class IsSelfOrSuperuser(BasePermission):
```
xoen: Again, you may want to rename this to `UserPermissions`.

control_panel_api/permissions.py
```diff
@@ -31,23 +28,13 @@ class S3BucketPermissions(IsSuperuser):
...
 
     def has_permission(self, request, view):
```
xoen: This should work.

I don't have enough experience with DRF permissions to tell if this is good or bad.

My only observation is that my understanding is that this method is supposed to determine if logged user can perform the class operations requested, while the `has_object_permission()` is to check if he can perform that operation on the specific instance of the user. This method is checking both things (as you it checks the `view.kwargs.get('pk')`).

Maybe having something like this would be more correct:

```python
def has_permission(self, request, view):
    if is_superuser(request.user):
        return True

    # Normal users can only update
    if request.method in ['PUT', 'PATCH']:
        return True


def has_object_permission(self, request, view, obj):
    if is_superuser(request.user):
        return True

    # otherwise normal users can potentially delete themselves
    if request.method not in ['GET', 'PUT', 'PATCH']:
        return False

    if hasattr(request.user, 'auth0_id'):
        return request.user == obj
```

I've seen something similar in the documentation.
Something I realised while writing this code is that we may want to check the request method to not allow normal users to perform non-`GET`/`PUT`/`PATCH`(e.g. so they can't somehow delete themselves).

This obviously is a bit longer so maybe have all the logic in `has_permission()` (even the one to check object permissions) is alright. As I said, I don't have enough experience with DRF permissions to know if this is bad or good.

control_panel_api/permissions.py
```diff
@@ -31,23 +28,23 @@ class S3BucketPermissions(IsSuperuser):
...
+        if is_superuser(request.user):
             return True
 
-        if hasattr(request.user, 'auth0_id'):
-            return view.kwargs.get('pk') == request.user.auth0_id
+        if request.user.is_anonymous():
+            return False
 
-        return False
+        return view.action not in ('create', 'destroy', 'list')
```
xoen: Nice that you can check the `view.action`, much nicer than checking the `request.method` 👍

PR #78, opened 2017-12-05
k8s proxy: Allow requests to Deployments endpoints (v1beta2)

xoen:
### What
Kubernetes Deployment endpoints are not part of v1 core API - [they're part of `v1beta2`](https://kubernetes.io/docs/api-reference/v1.8/#deployment-v1beta2-apps).

UI could make requests to get list of deployment or create them at some point.

### How
I tweaked the `K8sPermissions` class to:
1) Allow normale user requests to this endpoints (as long as they're
   operating in their personal namespace)
2) Make it easy to support other API groups if/when will need to for
  whatever reason

### DIFF
--- a/control_panel_api/permissions.py
+++ b/control_panel_api/permissions.py
@@ -25,6 +25,11 @@ class K8sPermissions(BasePermission):
     User can operate only in his namespace (unless superuser)
     """
 
+    ALLOWED_APIS = [
+        'api/v1',
+        'apis/apps/v1beta2',
+    ]
+
     def has_permission(self, request, view):
         if not request.user:
             return False
@@ -37,7 +42,12 @@ class K8sPermissions(BasePermission):
             return False
 
         path = request.path.lower()
-        return path.startswith(f'/k8s/api/v1/namespaces/user-{username}')
+
+        for api in self.ALLOWED_APIS:
+            if path.startswith(f'/k8s/{api}/namespaces/user-{username}'):
+                return True
+
+        return False
 
 
 class AppPermissions(IsSuperuser):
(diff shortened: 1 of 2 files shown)


jmoz:
Can you cover this with a unit test?

xoen:
@jmoz https://github.com/ministryofjustice/analytics-platform-control-panel/pull/78/commits/e699c3ceb030afd41264d7522e3f2949fb9da44b

jmoz:
LGTM 👺

control_panel_api/permissions.py
```diff
@@ -37,7 +42,12 @@ def has_permission(self, request, view):
             return False
 
         path = request.path.lower()
-        return path.startswith(f'/k8s/api/v1/namespaces/user-{username}')
+
+        for api in self.ALLOWED_APIS:
+            if path.startswith(f'/k8s/{api}/namespaces/user-{username}'):
+                return True
```
andyhd: I'd be tempted to change this to get the API from the path (`/(api\/.*)\/namespaces/`) and check it is in `ALLOWED_APIS` instead of looping, but this is fine as is.

control_panel_api/tests/test_permissions.py
```diff
@@ -451,5 +451,12 @@ def test_user_can_operate_in_their_namespace(self, mock_request, mock_load_confi
         mock_request.return_value.status_code = 200
 
         username = self.normal_user.username.lower()
-        response = self.client.get(f'/k8s/api/v1/namespaces/user-{username}')
-        self.assertEqual(HTTP_200_OK, response.status_code)
+
+        api_groups = [
+            'api/v1',
+            'apis/apps/v1beta2',
+        ]
```
andyhd: This would be better if it also tested disallowed APIs

xoen:
@andyhd https://github.com/ministryofjustice/analytics-platform-control-panel/pull/78/commits/073159970572dcf8498ce01c98e27b28058fa423

PR #79, opened 2017-12-05
Add custom pagination where ?page_size=0 returns all items

jmoz:
## What

Add custom pagination where ?page_size=0 returns all items

Keeps existing pagination but overrides and adds the functionality.

## How to review

1. Make a request and add ?page_size=0

https://trello.com/c/amqhVpb5/528-2-paginate-app-bucket-user-lists

### DIFF
--- a/control_panel_api/pagination.py
+++ b/control_panel_api/pagination.py
@@ -13,7 +13,8 @@ class CustomPageNumberPagination(PageNumberPagination):
         execute original parent functionality
         """
         page_size = self.get_page_size(request)
-        if page_size != 0:
+
+        if not self._page_size_is_all(page_size):
             return super().paginate_queryset(queryset, request, view)
 
         paginator = self.django_paginator_class(queryset, queryset.count())
@@ -35,3 +36,8 @@ class CustomPageNumberPagination(PageNumberPagination):
                 pass
 
         return self.page_size
+
+    def _page_size_is_all(self, page_size):
+        """If page_size is specified as 0 we use that to mean return all
+        """
+        return page_size == 0
--- a/control_panel_api/pagination.py
+++ b/control_panel_api/pagination.py
@@ -1,5 +1,37 @@
-from rest_framework.pagination import PageNumberPagination
+from rest_framework.pagination import PageNumberPagination, _positive_int
 
 
 class CustomPageNumberPagination(PageNumberPagination):
+    """This Pagination class allows a request to pass the page_size query param
+    for usual pagination functionality but adds custom whereby a value of
+    page_size=0 will return all items
+    """
     page_size_query_param = 'page_size'
+
+    def paginate_queryset(self, queryset, request, view=None):
+        """Override to check for page_size=0 if so then return all items else
+        execute original parent functionality
+        """
+        page_size = self.get_page_size(request)
+        if page_size != 0:
+            return super().paginate_queryset(queryset, request, view)
+
+        paginator = self.django_paginator_class(queryset, queryset.count())
+        self.page = paginator.page(1)
+
+        return list(self.page)
+
+    def get_page_size(self, request):
+        """Override to set strict=False to allow a 0 value
+        """
+        if self.page_size_query_param:
+            try:
+                return _positive_int(
+                    request.query_params[self.page_size_query_param],
+                    strict=False,
+                    cutoff=self.max_page_size
+                )
+            except (KeyError, ValueError):
+                pass
+
+        return self.page_size
--- a/control_panel_api/tests/test_pagination.py
+++ b/control_panel_api/tests/test_pagination.py
@@ -1,0 +1,29 @@
+from model_mommy import mommy
+from rest_framework.reverse import reverse
+from rest_framework.test import APITestCase
+
+from control_panel_api.tests.test_views import AuthenticatedClientMixin
+
+
+class TestPagination(AuthenticatedClientMixin, APITestCase):
+    def test_pagination(self):
+        mommy.make('control_panel_api.App', 120)  # 120 exceeds the 100 default
+
+        app_list_url = reverse('app-list')
+
+        fixtures = (
+            (app_list_url, 100, True),  # default paging no query param
+            (app_list_url + '?page_size=50', 50, True),  # less than default
+            (app_list_url + '?page_size=130', 120, False),  # more than default
+            (app_list_url + '?page_size=0', 120, False),  # return all
+        )
+
+        for url, expected_page_size, has_next in fixtures:
+            response = self.client.get(url)
+            self.assertEqual(120, response.data['count'])
+            self.assertEqual(expected_page_size, len(response.data['results']))
+            self.assertIsNone(response.data['previous'])
+            if has_next:
+                self.assertIsNotNone(response.data['next'])
+            else:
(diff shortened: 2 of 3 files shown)


control_panel_api/tests/test_views.py
```diff
@@ -783,3 +783,27 @@ def test_proxy(self, mock_request, mock_k8s_config):
...
         )
+
+
+class TestPagination(AuthenticatedClientMixin, APITestCase):
+    def test_pagination(self):
+        mommy.make('control_panel_api.App', 120)  # 120 exceeds the 100 default
+
+        app_list_url = reverse('app-list')
+
+        fixtures = (
```
xoen: nice

control_panel_api/pagination.py
```diff
@@ -1,5 +1,37 @@
...
+        page_size = self.get_page_size(request)
+        if page_size != 0:
+            return super().paginate_queryset(queryset, request, view)
+
+        paginator = self.django_paginator_class(queryset, queryset.count())
+        self.page = paginator.page(1)
+
+        return list(self.page)
+
+    def get_page_size(self, request):
```
xoen: This is all copied from `rest_framework.pagination.PageNumberPagination` right? So you can just not implement this and use the implementation in the parent class.

jmoz: Crucially I changed the boolean, that is all :/

control_panel_api/tests/test_views.py
```diff
@@ -783,3 +783,27 @@ def test_proxy(self, mock_request, mock_k8s_config):
             headers={'authorization': self.K8S_AUTH_TOKEN},
             verify=self.K8S_SSL_CERT_PATH,
         )
+
+
+class TestPagination(AuthenticatedClientMixin, APITestCase):
```
xoen: This test case could potentially be moved in a `test_pagination.py` file.

control_panel_api/pagination.py
xoen: Stupid question, why comment this method make the test fail? Shouldn't it the use `get_page_size()` in the parent class which is same as this?

I'm probably missing something.

xoen: It may have something to do with `_positive_int()` used internally, I'm having a quick look.

control_panel_api/tests/test_views.py
jmoz: Good idea

control_panel_api/pagination.py
xoen: Right, so it has something to do with the internal use of `_positive_int()` then.

jmoz: 👍

control_panel_api/pagination.py
```diff
@@ -1,5 +1,37 @@
...
+    def paginate_queryset(self, queryset, request, view=None):
+        """Override to check for page_size=0 if so then return all items else
+        execute original parent functionality
+        """
+        page_size = self.get_page_size(request)
+        if page_size != 0:
+            return super().paginate_queryset(queryset, request, view)
+
+        paginator = self.django_paginator_class(queryset, queryset.count())
+        self.page = paginator.page(1)
```
xoen: I'm not familiar with Django pagination, etc...so I may miss something here, isn't this going to return always page 1?

jmoz: Yep. Code is to always return the entire result set, which implies just one page which is 1.

xoen: Sorry, didn't see the `return super()...`.

jmoz: 😱

xoen: Don't understand why Python devs hate if/else...

xoen:
Andy may notice something I didn't but looks OK as far as I can tell 👍

control_panel_api/pagination.py
jmoz: That's not a python thing!  Avoids unnecessary nesting.

xoen: 😱

xoen: doge

control_panel_api/pagination.py
```diff
@@ -1,5 +1,43 @@
-from rest_framework.pagination import PageNumberPagination
+from rest_framework.pagination import PageNumberPagination, _positive_int
```
andyhd: Not sure you should import a `_` prefixed function - that implies it is not a part of the public API and could change without notice. I've been burned by this in the past.

control_panel_api/pagination.py
andyhd: I agree with @xoen's original comment - use the parent class implementation, eg:
```python
def get_page_size(self, request):
    try:
        page_size = super().get_page_size(request)

    except ValueError:
        try:
            if int(request.query_params[self.page_size_query_param]) == 0:
                return 0
        except (KeyError, ValueError):
            pass

    return self.page_size
```

This way, you don't have to depend on a private function.

control_panel_api/pagination.py
```diff
@@ -1,5 +1,43 @@
...
+                )
+            except (KeyError, ValueError):
+                pass
+
+        return self.page_size
+
+    def _page_size_is_all(self, page_size):
+        """If page_size is specified as 0 we use that to mean return all
+        """
+        return page_size == 0
```
andyhd: This method seems redundant. 0 is not a magic number.

control_panel_api/tests/test_pagination.py
```diff
@@ -0,0 +1,29 @@
+from model_mommy import mommy
+from rest_framework.reverse import reverse
+from rest_framework.test import APITestCase
+
+from control_panel_api.tests.test_views import AuthenticatedClientMixin
+
+
+class TestPagination(AuthenticatedClientMixin, APITestCase):
+    def test_pagination(self):
+        mommy.make('control_panel_api.App', 120)  # 120 exceeds the 100 default
```
andyhd: Rather than explain this with comments, it would be self-documenting to have a variable `default_page_size = 100` and then `mommy.make('...App', default_page_size + 20)`

control_panel_api/tests/test_pagination.py
```diff
@@ -0,0 +1,29 @@
...
+
+        app_list_url = reverse('app-list')
+
+        fixtures = (
+            (app_list_url, 100, True),  # default paging no query param
+            (app_list_url + '?page_size=50', 50, True),  # less than default
+            (app_list_url + '?page_size=130', 120, False),  # more than default
+            (app_list_url + '?page_size=0', 120, False),  # return all
```
andyhd: The same goes here, `?page_size={default_page_size - 50}` would make the intent obvious

control_panel_api/tests/test_pagination.py
jmoz: 🎖

control_panel_api/pagination.py
jmoz: In this case it is. 0 is not ZERO it is ALL.  Inside a named comparison function that describes this I think is better than a blind comparison to 0 in the main method.

control_panel_api/pagination.py
jmoz: I can't rely on a thrown exception as the parent hides it and always returns a default.

PR #80, opened 2017-12-07
Set root logger to output to console

andyhd:
## What

* Add logging config to log everything to the console by default


jmoz:
Can we have the more detailed format with timestamp and module name etc?  IIRC it's default set to a more simple output - may be wrong.

PR #82, opened 2017-12-12
Add endpoint to deploy tool

xoen:
### What

This adds the `POST /tools/{tool_name}/deployments/` endpoint which allow a user to deploy a tool in their namespace.

### Why
We want our user to potentially deploy their own tools (RStudio only for now).

### How
Deployment of a tool involves the installation/upgrade of the corresponding helm chart.

### Permissions
At this stage this endpoint is relatively simple and will only allow a user to deploy a tool in their namespace. Therefore the authorisation logic is only checking if user is logged in.

### Settings
This needs these new settings in order to work:

 - `TOOLS_DOMAIN` is the domain in which the tool will be available, e.g. `tools.example.com`
 - `RSTUDIO_AUTH_CLIENT_DOMAIN` is the OAuth client domain. It will default to the same domain as in `OIDC_DOMAIN` if not set
 - `RSTUDIO_AUTH_CLIENT_ID` is the OAuth client ID
 - `RSTUDIO_AUTH_CLIENT_SECRET` is the OAuth client secret

**NOTE**: When we'll add more and more tools this approach of having the tool settings in the environment variables *may not* work, we'll cross that bridge when we'll get there.

### How to review

1. Add these environment variables for `dev` in your `.dev` file
2. start the server by `$ python3 manage.py runserver`
3. Go to [http://localhost:8000](http://localhost:8000) and log in
4. User the `POST /tools/{tool_name}/deployments/` endpoint
5. Try to enter an invalid tool name => will return a `400 BAD REQUEST`
6. Try to enter `rstudio` as tool name => you should get a `201 CREATED`
7. Visit `{your_username}-rstudio.{TOOLS_DOMAIN}

### Ticket
https://trello.com/c/77FMKVLE/536-3-cp-api-add-endpoint-to-deploy-rstudio

### Related PR
The following things needs be in place before merging/deploying this:
- [ ] [Add config for cpanel](https://github.com/ministryofjustice/analytics-platform-config/pull/21)
- [ ] [Update `cpanel` helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/61)

### DIFF
--- a/control_panel_api/tools.py
+++ b/control_panel_api/tools.py
@@ -46,7 +46,7 @@ class Tool():
             f'{username}-{self.name}',
             f'mojanalytics/{self.name}',
             '--namespace', user.k8s_namespace,
-            '--set', f'Username={username}',
+            '--set', f'username={username}',
             '--set', f'aws.iamRole={user.iam_role_name}',
             '--set', f'toolsDomain={settings.TOOLS_DOMAIN}',
             '--set', f'authProxy.cookieSecret={auth_proxy_cookie_secret}',
--- a/control_panel_api/tools.py
+++ b/control_panel_api/tools.py
@@ -25,9 +25,9 @@ class Tool():
             '--namespace', user.k8s_namespace,
             '--set', f'Username={user.username.lower()}',
             '--set', f'aws.iamRole={user.iam_role_name}',
-            '--set', 'authProxy.auth.domain=' + self.auth_client["domain"],
-            '--set', 'authProxy.auth.clientId=' + self.auth_client["id"],
-            '--set', 'authProxy.auth.clientSecret=' + self.auth_client["secret"],
+            '--set', 'authProxy.auth.domain=' + self.auth_client['domain'],
+            '--set', 'authProxy.auth.clientId=' + self.auth_client['id'],
+            '--set', 'authProxy.auth.clientSecret=' + self.auth_client['secret'],
         )
 
     def load_auth_client_config(self):
--- a/control_panel_api/tools.py
+++ b/control_panel_api/tools.py
@@ -1,4 +1,5 @@
 import os
+import secrets
 
 from django.conf import settings
 
@@ -10,7 +11,10 @@ class Tool():
     def __init__(self, name):
         self.name = name
         self.helm = Helm()
-        self.load_auth_client_config()
+
+        self.auth_client_domain = _get_auth_client_config('domain')
+        self.auth_client_id = _get_auth_client_config('id')
+        self.auth_client_secret = _get_auth_client_config('secret')
 
     def deploy_for(self, user):
         """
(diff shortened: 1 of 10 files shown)


control_panel_api/tools.py
```diff
@@ -0,0 +1,66 @@
...
+
+class UnsupportedToolException(APIException):
+    tools_list = ", ".join(SUPPORTED_TOOL_NAMES)
+
+    status_code = 400
+    default_detail = f'Unsupported tool, tool_name must be one of: {tools_list}'
+    default_code = 'unsupported_tool'
+
+
+class Tool():
```
andyhd: Although all classes in Python 3 are new-style classes, you should still use `class Tool(object):` - at least until Python 2 has gone away

control_panel_api/tools.py
```diff
@@ -0,0 +1,66 @@
...
+        >>> rstudio = Tool('rstudio')
+        >>> rstudio.deploy_for(alice)
+        """
+
+        username = user.username.lower()
+        auth_proxy_cookie_secret = secrets.token_hex(32)
+        tool_cookie_secret = secrets.token_hex(32)
+
+        self.helm.upgrade_release(
+            f'{username}-{self.name}',
```
andyhd: Not sure, but I think you might need to use `user.k8s_namespace` here instead of `username`, and below in `username={username}`

control_panel_api/tools.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _get_auth_client_config(self, key):
+        setting_key = f'{self.name}_AUTH_CLIENT_{key}'
+        return getattr(settings, setting_key.upper())
+
+    def _validate_name(self, name):
+        if not name in SUPPORTED_TOOL_NAMES:
+            raise UnsupportedToolException
+
+        return name
```
andyhd: Not sure this needs to be a method

control_panel_api/tools.py
```diff
@@ -0,0 +1,66 @@
...
+
+from django.conf import settings
+from rest_framework.exceptions import APIException
+
+from control_panel_api.helm import Helm
+
+
+SUPPORTED_TOOL_NAMES = [
+    'rstudio',
+]
```
andyhd: Should this be in config, or somewhere not in the code? (DB?)

control_panel_api/tools.py
xoen: 👍

control_panel_api/tools.py
xoen: I think we'll iterate on this as we'll introduce more tools.
For now my preference is for this list to be close to the code which uses it.

control_panel_api/tools.py
xoen: Oh so this is for retro-compatibility reason. Copy that 👍

control_panel_api/tools.py
xoen: I used the same naming convention used currently when deploying using the jenkins job to avoid any confusion or bugs: https://github.com/ministryofjustice/analytics-platform-control-panel/blob/jenkins-control-panel/scripts/deploy_rstudio#L31

control_panel_api/tools.py
andyhd: 👍

PR #83, opened 2017-12-13
Auth RS256

jmoz:
## What

Changes auth to RS256 using json web keys via auth0.

Library change to python-jose for more effective jwk handling.

Tests use a fresh rsa key only used for the test.

Also includes logging fix.

## How to review

1. ./manage.py test
2. Run the frontend with the app and login via github.

https://trello.com/c/vhfSv2Vv/546-change-jwt-signing-method-in-cp-api-to-rs256-to-match-k8s-api-2

### DIFF
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -17,8 +17,7 @@ logger = logging.getLogger(__name__)
 
 
 def get_jwks():
-    response = requests.get(
-        f'https://{settings.OIDC_DOMAIN}/.well-known/jwks.json')
+    response = requests.get(settings.OIDC_WELL_KNOWN_URL)
     return response.json()
 
 
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -21,12 +21,18 @@ def get_jwks():
     return response.json()
 
 
-def get_key(jwks, unverified_header):
+def get_key(jwks, kid):
     for jwk in jwks.get('keys'):
-        if jwk['kid'] == unverified_header['kid']:
+        if jwk['kid'] == kid:
             return jwk
 
     return None
+
+
+def get_jwk(kid):
+    jwks = get_jwks()
+
+    return get_key(jwks, kid)
 
 
 class Auth0JWTAuthentication(BaseAuthentication):
@@ -45,12 +51,11 @@ class Auth0JWTAuthentication(BaseAuthentication):
             raise AuthenticationFailed('Error decoding JWT header') from e
 
         try:
-            jwks = get_jwks()
+            key = get_jwk(unverified_header['kid'])
         except RequestException as e:
             logger.error(e)
             raise AuthenticationFailed(e) from e
 
-        key = get_key(jwks, unverified_header)
         if key is None:
             raise AuthenticationFailed('kid matching failure')
 
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -35,15 +35,15 @@ def get_jwk(kid):
     return get_key(jwks, kid)
 
 
-def get_or_create_user(payload):
+def get_or_create_user(decoded_payload):
     try:
-        user = User.objects.get(pk=payload.get('sub'))
+        user = User.objects.get(pk=decoded_payload.get('sub'))
     except User.DoesNotExist:
         user = User.objects.create(
-            pk=payload.get('sub'),
-            username=payload.get(settings.OIDC_FIELD_USERNAME),
-            email=payload.get(settings.OIDC_FIELD_EMAIL),
-            name=payload.get(settings.OIDC_FIELD_NAME),
+            pk=decoded_payload.get('sub'),
+            username=decoded_payload.get(settings.OIDC_FIELD_USERNAME),
+            email=decoded_payload.get(settings.OIDC_FIELD_EMAIL),
+            name=decoded_payload.get(settings.OIDC_FIELD_NAME),
         )
         user.save()
         user.helm_create()
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -35,6 +35,22 @@ def get_jwk(kid):
     return get_key(jwks, kid)
 
 
+def get_or_create_user(payload):
+    try:
+        user = User.objects.get(pk=payload.get('sub'))
+    except User.DoesNotExist:
+        user = User.objects.create(
+            pk=payload.get('sub'),
+            username=payload.get(settings.OIDC_FIELD_USERNAME),
+            email=payload.get(settings.OIDC_FIELD_EMAIL),
+            name=payload.get(settings.OIDC_FIELD_NAME),
+        )
+        user.save()
+        user.helm_create()
+
+    return user
+
+
 class Auth0JWTAuthentication(BaseAuthentication):
     def authenticate(self, request):
         try:
@@ -70,21 +86,9 @@ class Auth0JWTAuthentication(BaseAuthentication):
             logger.error(e)
             raise AuthenticationFailed(e) from e
 
-        sub = decoded.get('sub')
-
-        if sub is None:
+        if decoded.get('sub') is None:
             raise AuthenticationFailed('JWT missing "sub" field')
 
-        try:
-            user = User.objects.get(pk=sub)
-        except User.DoesNotExist:
-            user = User.objects.create(
-                pk=sub,
-                username=decoded.get(settings.OIDC_FIELD_USERNAME),
-                email=decoded.get(settings.OIDC_FIELD_EMAIL),
-                name=decoded.get(settings.OIDC_FIELD_NAME),
-            )
-            user.save()
-            user.helm_create()
+        user = get_or_create_user(decoded)
 
         return user, None
--- a/control_panel_api/authentication.py
+++ b/control_panel_api/authentication.py
@@ -1,22 +1,39 @@
+import logging
+
+import requests
 from django.conf import settings
-import jwt
-from jwt.exceptions import InvalidTokenError
+from jose import jwt
+from jose.exceptions import JWTError
+from requests.exceptions import RequestException
 from rest_framework.authentication import (
     BaseAuthentication,
-    get_authorization_header
+    get_authorization_header,
 )
 from rest_framework.exceptions import AuthenticationFailed
 
 from control_panel_api.models import User
 
+logger = logging.getLogger(__name__)
+
+
+def get_jwks():
+    response = requests.get(
+        f'https://{settings.OIDC_DOMAIN}/.well-known/jwks.json')
+    return response.json()
+
+
+def get_key(jwks, unverified_header):
+    for jwk in jwks.get('keys'):
+        if jwk['kid'] == unverified_header['kid']:
+            return jwk
+
+    return None
+
 
 class Auth0JWTAuthentication(BaseAuthentication):
-
     def authenticate(self, request):
-
         try:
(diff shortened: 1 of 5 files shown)


control_panel_api/authentication.py
```diff
@@ -1,37 +1,70 @@
...
 
 from control_panel_api.models import User
 
+logger = logging.getLogger(__name__)
 
-class Auth0JWTAuthentication(BaseAuthentication):
 
-    def authenticate(self, request):
+def get_jwks():
+    response = requests.get(
```
kerin: I suspect it's not necessarily safe to hit this URL directly, as the JWKS URL isn't part of the OIDC spec. The configuration URL is though, and the response from that contains the JWKS URL:

https://dev-analytics-moj.eu.auth0.com/.well-known/openid-configuration -

`"jwks_uri":"https://dev-analytics-moj.eu.auth0.com/.well-known/jwks.json"`

So it's better to hit the `openid-configuration` URL first, and get the JWKS URL from the response.

control_panel_api/authentication.py
```diff
@@ -1,37 +1,70 @@
...
+def get_key(jwks, unverified_header):
+    for jwk in jwks.get('keys'):
+        if jwk['kid'] == unverified_header['kid']:
+            return jwk
 
+    return None
+
+
+class Auth0JWTAuthentication(BaseAuthentication):
+    def authenticate(self, request):
```
kerin: Should this method be broken up into a few smaller methods? There's quite a lot going on here now

control_panel_api/authentication.py
```diff
@@ -1,37 +1,70 @@
+import logging
+
+import requests
```
andyhd: Import order

requirements.txt
```diff
@@ -15,6 +15,8 @@ django-filter==1.0.4
 django-rest-swagger==2.1.2
 djangorestframework==3.7.3
 docutils==0.14
+ecdsa==0.13
```
kerin: Does the `ecdsa` package need to be included here, or is it already a dependency of `pycrypto` or `python-jose`?

jmoz: I rebuilt the venv to make sure they're all correct, installed as part of jose.

control_panel_api/authentication.py
jmoz: Looks ok?  Auto ordered/grouped by system, site, app, in alpha.

control_panel_api/authentication.py
jmoz: I realise it looks big, but if you look each method call is doing something specific and then the errors are clearly handled in this controlling function block (which I would rather have than some error handling here and some in other functions).

control_panel_api/authentication.py
```diff
@@ -1,37 +1,69 @@
...
+    response = requests.get(settings.OIDC_WELL_KNOWN_URL)
+    return response.json()
+
+
+def get_key(jwks, unverified_header):
+    for jwk in jwks.get('keys'):
+        if jwk['kid'] == unverified_header['kid']:
+            return jwk
 
+    return None
```
xoen: You may just raise `AuthenticationFailed` from here so you don't have to check the return value of this method outside (and save few lines of code).

control_panel_api/authentication.py
```diff
@@ -1,37 +1,69 @@
...
 
 from control_panel_api.models import User
 
+logger = logging.getLogger(__name__)
 
-class Auth0JWTAuthentication(BaseAuthentication):
 
-    def authenticate(self, request):
+def get_jwks():
+    response = requests.get(settings.OIDC_WELL_KNOWN_URL)
```
xoen: Again, I think may would be cleaner catch and handle the `RequestException` from inside this method where it's more clear what's going on.

This would also simplify the caller.

control_panel_api/authentication.py
```diff
@@ -1,37 +1,69 @@
...
 
         if prefix != 'JWT'.encode():
             return None
 
+        try:
+            unverified_header = jwt.get_unverified_header(token)
+        except JWTError as e:
+            raise AuthenticationFailed('Error decoding JWT header') from e
+
+        try:
```
xoen: I'd move the handling of this exception into `get_jwks()` to streamline this method.

control_panel_api/authentication.py
```diff
@@ -1,37 +1,69 @@
...
+            jwks = get_jwks()
+        except RequestException as e:
+            logger.error(e)
+            raise AuthenticationFailed(e) from e
+
+        key = get_key(jwks, unverified_header)
```
xoen: I'd move this logic inside `get_key()` to streamline this method.

control_panel_api/authentication.py
```diff
@@ -1,37 +1,69 @@
...
+        if key is None:
+            raise AuthenticationFailed('kid matching failure')
+
         try:
```
xoen: I'd move this logic and exception handling in a method, something called `get_decoded_jwt()` or something like that.

This would make this `authenticate()` method a little big smaller.

control_panel_api/authentication.py
xoen: I think as well it would be good if we split this in smaller, well-named functions if possible.

control_panel_api/authentication.py
kerin: Actually I think I'm wrong about this - looking at Auth0's own examples, they hit the JWKS endpoint directly - https://github.com/auth0/node-jwks-rsa#usage - and it's also probably safe to assume that if Auth0 change the JWKS endpoint URL they'd give us warning well in advance.

control_panel_api/authentication.py
jmoz: Yeh I think in this case I can put `get_key` and `get_jwks` together. Although usually I'd prefer not to couple it.

control_panel_api/authentication.py
jmoz: I prefer the opposite - exception handling and errors should be top level to avoid having to inspect many different methods where exception handling could be. 

The top level controlling `authenticate` function knows what error bubble up, catches and handles appropriately, in one place.

control_panel_api/authentication.py
jmoz: Ok, cool.

control_panel_api/authentication.py
xoen: I see what you mean and it's a fair point as these methods are "external" to the `Auth0JWTAuthentication` class so raise a `AuthenticationFailed` exception may seem a bit odd.

I suppose **if** you want to make this a bit smaller there are a couple of options:
1. Extract all the logic to get/verify/decode the JWT token in a separate class which raise specific exceptions, e.g. `KidMismatchError`, etc...which then you catch in `Auth0JWTAuthentication` and handle by raising an `AuthenticationFailed` exception
2. You move these helper methods in `Auth0JWTAuthentication` so that raise  an`AuthenticationFailed` exception inside them makes sense.

Option 1 may be a bit overkill.

Obviously, feel free to play with it if you want. It seems to me that this method is basically doing 2 things:
1. Extract/verify/decode JWT token
2. Get/create the user and return it (authenticating him)

Maybe the first part could be moved in a separate method.

control_panel_api/authentication.py
andyhd: `requests` surely comes after `django`?

control_panel_api/authentication.py
andyhd: I wouldn't do that - it's mixing responsibilities. Getting a matching key out of a list doesn't technically have anything to do with authentication.

control_panel_api/authentication.py
jmoz: Good point, didn't notice.  It was some weird from/import grouping option..

control_panel_api/authentication.py
andyhd: I agree, the code in `authenticate` is too low level - it could be split out to look something like this:
```python
def authenticate(self, request):
    jwt = self.get_jwt(request)  # throws AuthenticationFailed if JWT couldn't be found/decoded
    return self.get_or_create_user(jwt.sub)
```

control_panel_api/authentication.py
andyhd: Rather than put the two existing functions together, which I think are already well factored, just add another function which calls them

control_panel_api/authentication.py
andyhd: Just realised that this is what @xoen said earlier

control_panel_api/authentication.py
xoen: Yes as mentioned in other comment, as these functions are outside the `Auth0JWTAuthentication` class raise an `AuthenticationFailed` here is a bit odd.

control_panel_api/authentication.py
andyhd: I agree with this one - partially. `get_jwks` should maybe catch and log (esp. to Sentry) the `RequestException`, but either
1. raise a new more specific exception (eg: `JwksFetchError`), or
2. return an empty dict to pass to `get_key` and let the authentication fail because the key could not be matched.

control_panel_api/authentication.py
kerin: @andyhd I thought that, but there doesn't seem to be any expiry times on the JWKS? https://auth0.com/docs/jwks

requirements.txt
kerin: should drop that from our requirements.txt then, and let jose handle it?

jmoz: It all needs to be pinned to prevent version errors.  This is simply the output of `pip freeze > requirements.txt` which is the standard way to lock it down.

jmoz:
Refactored out the get_user bit.

requirements.txt
kerin: Yep, but then you end up with your dependencies' dependencies in requirements.txt. I've always included only the top-level packages, and let the packages take care of their own dependencies. Am I right @andyhd ? Or does this just not matter very much 😬

jmoz: I think you should simply rely on `pip freeze` and not craft it yourself.  

Crafting it yourself also often leaves random packages added to the list and no one knows what's been added, then have to delete the venv and build again to check etc etc.

andyhd: I usually do pip freeze. It might be worth switching to [pipenv](https://github.com/pypa/pipenv) for this sort of dependency locking.

control_panel_api/authentication.py
```diff
@@ -1,54 +1,94 @@
...
+    return None
+
+
+def get_jwk(kid):
+    jwks = get_jwks()
+
+    return get_key(jwks, kid)
+
+
+def get_or_create_user(payload):
```
xoen: I'd probably rename `payload` => `token`.
Also, shouldn't this be a "private" method of `Auth0JWTAuthentication`?

jmoz: It's the decoded payload.  The token is the encrypted string passed in the header.  After decoding it's split into the payload.  The token also contains a header and secret fyi.

Yes those functions can be part of the class. Not sure why I left them out.

xoen: You're right, I suppose in my mind when I think "token" I keep thinking about the decoded JSON but the token is the string getting in. Still `payload` is a bit generic, maybe something like `decoded_token` or something like that would work better...but nevermind, it was just an idea.

jmoz:
We will need to switch the auth0 api client to the new RS256 one.  Is that about right @kerin?

kerin:
@jmoz probably better to change the algo on the existing client rather than switch over to a new one - give it a whirl in the dev env first though, of course.

PR #84, opened 2017-12-29
Create role oidc provider policy

kerin:
## What

Added an OIDC provider section to user Role trust policies. This is intended to replace the current SAML provider and trust policy, but the SAML section has been left in place for now to maintain compatibility until the SAML provider is removed.

## How to review

1. Run the API
2. Create a user
3. Verify that the user's role trust policy has an OIDC provider section, with the provider referencing the OIDC provider, and the user's OIDC sub/auth0 ID used as a Condition block.


kerin:
@andyhd don't the changes to the existing tests cover it, with [add_oidc_statement here](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/84/files#diff-9301a102f09942cbdbdfbaae09557e82R184)?

control_panel_api/services.py
```diff
@@ -51,7 +51,9 @@ def _policy_arn(bucket_name, readwrite=False):
 
 
 @ignore_existing
-def create_role(role_name, add_saml_statement=False):
+def create_role(
+        role_name, add_saml_statement=False, add_oidc_statement=False,
+        oidc_sub=None):
```
andyhd: This should be formatted with one argument per line

PR #87, opened 2018-01-18
Add helm delete for init-user and delete their namespace on User delete

jmoz:
## What

User delete now deletes the helm chart and the user's namespace.

The helm command subprocess call needs to change to use some shell trickery.

Renamed decorator to more appropriate.

Test for calls.

## How to review

1. Create a user using API
2. Check for helm 
3. Delete user using API and check helm deleted

https://trello.com/c/zO7SL4ts/589-write-script-that-deletes-leanne-as-a-user-in-order-to-recreate-role-and-recognise-where-theres-an-issue

### DIFF
--- a/control_panel_api/helm.py
+++ b/control_panel_api/helm.py
@@ -14,6 +14,14 @@ class Helm(object):
         if self.enabled:
             subprocess.run(['helm', command] + list(args), check=True)
 
+    def _helm_shell_command(self, command_string):
+        if self.enabled:
+            subprocess.run(
+                f'helm {command_string}',
+                shell=True,
+                check=True
+            )
+
     def upgrade_release(self, release, chart, *args):
         default_flags = ['--install', '--wait']
         flags = list(args) + default_flags
@@ -28,6 +36,16 @@ class Helm(object):
             '--set', f'Username={username_slug}',
             '--set', f'Email={email}',
             '--set', f'Fullname={fullname}',
+        )
+
+    def delete_user(self, username):
+        username_slug = sanitize_dns_label(username)
+        self._helm_command('delete', f'init-user-{username_slug}', '--purge')
+
+    def delete_namespace(self, username):
+        username_slug = sanitize_dns_label(username)
+        self._helm_shell_command(
+            f'delete --purge $(helm list -q --namespace user-{username_slug})'
         )
 
     def config_user(self, username):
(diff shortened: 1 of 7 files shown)


control_panel_api/helm.py
```diff
@@ -30,6 +38,16 @@ def init_user(self, username, email, fullname):
             '--set', f'Fullname={fullname}',
         )
 
+    def delete_user(self, username):
```
xoen: Don't know if you want to make this method a bit more explicit and rename to something like `uninstall_init_user_chart()`.

Similarly the `delete_namespace()` method could be renamed to `uninstall_user_charts()`.

Not a big deal, reason why I'm mentioning this is because the user namespace is actually created as part of the installation of the `init_user` helm chart, therefore `delete_namespace` could be a bit misleading.

control_panel_api/tests/test_models.py
```diff
@@ -40,6 +42,20 @@ def test_helm_create_user(self):
         helm.config_user.assert_called_with(username)
         helm.init_user.assert_called_with(username, email, name)
 
+    def test_helm_delete_user(self):
+        username = 'foo'
+        email = 'bar@baz.com'
+        name = 'bat'
+        user = User.objects.create(
```
xoen: In this test you don't need to persist the user strictly speaking.
You could just build a user object with `User(username=username)` and save a trip to the DB :) (which make this test a bit faster).

control_panel_api/tests/test_permissions.py
```diff
@@ -426,10 +426,10 @@ def test_when_not_authenticated_responds_403(self):
         response = self.client.get('/k8s/something')
         self.assertEqual(HTTP_403_FORBIDDEN, response.status_code)
 
-    @patch('kubernetes.client.configuration')
-    @patch('kubernetes.config.load_incluster_config')
+    @patch('kubernetes.client.configuration', MagicMock())
```
xoen: nice 👍

control_panel_api/tests/test_models.py
jmoz: 👍

control_panel_api/helm.py
andyhd: I would go the other way - helm shouldn't know about "users", so these methods (`delete_user`, `config_user`, `init_user`) really belong in elsewhere, probably in the calling code. But this module is really just a collection of helper functions, so I don't think it matters much.

control_panel_api/tests/test_permissions.py
jmoz: No, default causes a param to be injected in test methods. Specifying here prevents that.

PR #89, opened 2018-01-26
S3bucket non app bucket

jmoz:
## What

This adds functionality for a 'data_warehouse' bucket (I don't like this terminology btw) which is just a regular s3bucket with a boolean flag, but which can not be related to apps via apps3buckets.  (I prefer inverting this so an s3bucket is a s3bucket, but an app bucket can only be related - the functionality and special case is for the app bucket.)

You can filter for different bucket types in request.

Please discuss terminology.

## How to review

1. Run tests.
2. Call api using filter.
3. Try relate to app.

https://trello.com/c/8gZXcJTV/603-data-warehouse-s3buckets


control_panel_api/serializers.py
```diff
@@ -36,6 +36,14 @@ def update(self, instance, validated_data):
 
         return super().update(instance, validated_data)
 
+    def create(self, validated_data):
+        if validated_data['s3bucket'].is_data_warehouse:
+            raise serializers.ValidationError(
+                'Only app s3bucket can be related.'
```
xoen: I would tweak the error message to that it's more explicit. Maybe something like `Apps cannot access data warehouse S3 buckets`, or something like that.

control_panel_api/views.py
```diff
@@ -183,6 +183,7 @@ class S3BucketViewSet(viewsets.ModelViewSet):
     serializer_class = S3BucketSerializer
     filter_backends = (S3BucketFilter,)
     permission_classes = (S3BucketPermissions,)
+    filter_fields = ('is_data_warehouse',)
```
xoen: Should we have a test for the filtering by `is_data_warehouse` flag?

xoen: Sorry, I didn't see you added a test in the existing `test_list` 👍

jmoz: 😄

control_panel_api/tests/test_views.py
```diff
@@ -519,7 +544,11 @@ def setUp(self):
     def test_list(self):
         response = self.client.get(reverse('s3bucket-list'))
         self.assertEqual(HTTP_200_OK, response.status_code)
-        self.assertEqual(len(response.data['results']), 2)
+        self.assertEqual(len(response.data['results']), 3)
+
+        response = self.client.get(
+            reverse('s3bucket-list') + '?is_data_warehouse=true')
```
xoen: Stupid question: If the URL is `?is_data_warehouse=false`, would only return the S3 buckets objects with that flag set to `False` right?

jmoz: Yeh of course

PR #90, opened 2018-01-29
Migration to add OIDC statement to all users' trust policies

xoen:
### Why
Because without this trust policy users will not be able to log in to AWS using OIDC.

### What

The migration will loop through the users:
1) read the role's trust policy
2) check if the trust policy has already the OIDC statement
3) If not, it will add it to the document
4) Will write back the new role trust policy (if needed)

### boto3 relevant documentation
* [`IAM.Role()`](https://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.Role)
* [`IAM.Role.assume_role_policy_document`](https://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.Role.assume_role_policy_document)
* [`IAM.AssumeRolePolicy()`](https://boto3.readthedocs.io/en/latest/reference/services/iam.html#assumerolepolicy)
* [`IAM.AssumeRolePolicy.update()`](https://boto3.readthedocs.io/en/latest/reference/services/iam.html#IAM.AssumeRolePolicy.update)
### Related PRs
This is related to Kerin's PR: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/84

### Ticket
https://trello.com/c/Vy7ImBcn/602-4-sso-aws-federated-login-user-roles-migration

### DIFF
--- a/control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
+++ b/control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
@@ -8,6 +8,9 @@ from django.core.management.base import BaseCommand
 from control_panel_api.models import User
 
 
+OIDC_PROVIDER_ARN = f'{settings.IAM_ARN_BASE}:oidc-provider/{settings.OIDC_DOMAIN}/'
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -21,9 +24,11 @@ class Command(BaseCommand):
             iam_role = iam.Role(user.iam_role_name)
             assume_role_policy_document = iam_role.assume_role_policy_document
 
-            if not self._has_oidc_statement(assume_role_policy_document):
+            if _has_oidc_statement(assume_role_policy_document):
+                logger.info(f'OIDC statement already found in "{user.iam_role_name}" trust policy. Skipping.')
+            else:
                 # Add OIDC statement
-                oidc_statement = self._oidc_statement(user.auth0_id)
+                oidc_statement = _oidc_statement(user.auth0_id)
                 assume_role_policy_document['Statement'].append(oidc_statement)
 
                 # Update role trust policy
@@ -32,34 +37,30 @@ class Command(BaseCommand):
                     PolicyDocument=json.dumps(assume_role_policy_document)
                 )
                 logger.info(f'OIDC statement added to "{user.iam_role_name}" trust policy.')
-            else:
-                logger.info(f'OIDC statement already found in "{user.iam_role_name}" trust policy. Skipping.')
 
-    def _oidc_statement(self, oidc_sub):
-        return {
-            "Effect": "Allow",
-            "Principal": {
-                "Federated": self._oidc_provider_arn,
+
+def _has_oidc_statement(document):
+    for statement in document['Statement']:
+        try:
+            oidc_provider_arn = statement['Principal']['Federated']
+            if oidc_provider_arn == OIDC_PROVIDER_ARN:
+                return True
+        except KeyError:
+            pass
+
+    return False
+
+
+def _oidc_statement(oidc_sub):
+    return {
+        "Effect": "Allow",
+        "Principal": {
+            "Federated": OIDC_PROVIDER_ARN,
+        },
+        "Action": "sts:AssumeRoleWithWebIdentity",
+        "Condition": {
+            "StringEquals": {
+                f"{settings.OIDC_DOMAIN}/:sub": oidc_sub,
             },
-            "Action": "sts:AssumeRoleWithWebIdentity",
-            "Condition": {
-                "StringEquals": {
-                    f"{settings.OIDC_DOMAIN}/:sub": oidc_sub,
-                },
-            },
-        }
-
-    def _has_oidc_statement(self, document):
-        for statement in document['Statement']:
-            try:
-                oidc_provider_arn = statement['Principal']['Federated']
-                if oidc_provider_arn == self._oidc_provider_arn:
-                    return True
-            except KeyError:
-                pass
-
-        return False
-
-    @property
-    def _oidc_provider_arn(self):
-        return f"{settings.IAM_ARN_BASE}:oidc-provider/{settings.OIDC_DOMAIN}/"
+        },
+    }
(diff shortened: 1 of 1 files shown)


xoen:
FYI: Tested locally on my usual test user/role (`aldo-justice-test`). 

CP-API role probably needs some permissions in order for this to work.

xoen:
@kerin These are the permissions needed by this migration: https://github.com/ministryofjustice/analytics-platform-ops/pull/124 - We'll get rid of them after we migrate.

(Tested this in dev in a pod with this branch deployed and it worked as expected)

control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
```diff
@@ -0,0 +1,65 @@
...
+    help = "Adds OIDC statement to all users roles trust policies"
+
+    def handle(self, *args, **options):
+        iam = boto3.resource('iam')
+        for user in User.objects.all():
+            # Read role trust policy
+            iam_role = iam.Role(user.iam_role_name)
+            assume_role_policy_document = iam_role.assume_role_policy_document
+
+            if not self._has_oidc_statement(assume_role_policy_document):
```
andyhd: Generally, it's clearer to test for truth (avoid `if not`). I would invert the logic here (and switch the if/else blocks).

control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
```diff
@@ -0,0 +1,65 @@
...
+                assume_role_policy_document['Statement'].append(oidc_statement)
+
+                # Update role trust policy
+                assume_role_policy = iam_role.AssumeRolePolicy()
+                assume_role_policy.update(
+                    PolicyDocument=json.dumps(assume_role_policy_document)
+                )
+                logger.info(f'OIDC statement added to "{user.iam_role_name}" trust policy.')
+            else:
+                logger.info(f'OIDC statement already found in "{user.iam_role_name}" trust policy. Skipping.')
```
andyhd: Is this log output necessary? It seems not, unless it's a noteworthy fact that the OIDC statement is already found.

control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
```diff
@@ -0,0 +1,65 @@
...
+                if oidc_provider_arn == self._oidc_provider_arn:
+                    return True
+            except KeyError:
+                pass
+
+        return False
+
+    @property
+    def _oidc_provider_arn(self):
+        return f"{settings.IAM_ARN_BASE}:oidc-provider/{settings.OIDC_DOMAIN}/"
```
andyhd: There's no need for this function to be a `@property` or even a method, since it doesn't use `self`.

andyhd: Or even a function, if `settings` does not change between loading this script file and running it.

control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
```diff
@@ -0,0 +1,65 @@
...
+                # Update role trust policy
+                assume_role_policy = iam_role.AssumeRolePolicy()
+                assume_role_policy.update(
+                    PolicyDocument=json.dumps(assume_role_policy_document)
+                )
+                logger.info(f'OIDC statement added to "{user.iam_role_name}" trust policy.')
+            else:
+                logger.info(f'OIDC statement already found in "{user.iam_role_name}" trust policy. Skipping.')
+
+    def _oidc_statement(self, oidc_sub):
```
andyhd: This also does not need to be a method, it can just be a function

control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
xoen: Sounds like a good candidate for a "constant". Not sure why I ended up writing a function, I'll change 👍

xoen:
@andyhd tweaked migration as per previous comments: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/90/commits/2e2bf10c1feec800a98601db7fa5679a2e4ae11e

Let me know if there is anything else that doesn't look right.

(tested this with a one user in various scenarios and it seemed to still work after changes but don't take my word for it)

xoen:
Also @kerin - don't know if you had the chance to have a look at this (after all it's related to the work you've done)

xoen:
FYI: In order to run this will have to terraform-apply this when we'll get there: https://github.com/ministryofjustice/analytics-platform-ops/pull/124

control_panel_api/management/commands/add_oidc_statement_to_users_roles.py
```diff
@@ -0,0 +1,66 @@
...
+
+class Command(BaseCommand):
+    help = "Adds OIDC statement to all users roles trust policies"
+
+    def handle(self, *args, **options):
+        iam = boto3.resource('iam')
+        for user in User.objects.all():
+            # Read role trust policy
+            iam_role = iam.Role(user.iam_role_name)
+            assume_role_policy_document = iam_role.assume_role_policy_document
```
andyhd: I usually take comments to indicate that there should be a self-documenting function instead, eg:
```python
for user in User.objects.all():
    policy = read_role_trust_policy(user)

    if has_oidc_statement(policy):
        logger.info('...Skipping.')

    else:
        add_oidc_statement(policy, user)
        update_role_trust_policy(policy, user)
        logger.info('OIDC statement added ...')
```

xoen: I take your point. One reason why I didn't go that route is to avoid making this too opaque.

It's relatively straightforward and I didn't want to over-engineer it but at the same time leave some crumbs behind for the future me/others (therefore the couple of comments).

xoen:
[(4) SSO/AWS federated login user roles migration](https://trello.com/c/Vy7ImBcn/602-4-sso-aws-federated-login-user-roles-migration)

xoen:
Cheers @kerin I'm merging this so you can use later when the other one you're working on is ready.

PR #92, opened 2018-01-30
Clean-up after IAM policy revamp

xoen:
### Why
Permissions to S3 are not in managed IAM policies anymore.
They're in the user role `s3-access` inline policy.

### What
This means we can delete a bunch of functions which attach/detach policies,
etc...
 - `aws.attach_policy_to_role()`
 - `aws.create_policy()`
 - `aws.delete_policy()`
 - `aws.detach_policy_from_entities()`
 - `aws.detach_policy_from_group()`
 - `aws.detach_policy_from_user()`
 - `models.S3Bucket.aws_delete()`
 - `services.create_bucket_policies()`
 - `services.delete_bucket_policies()`
 - `services.get_policy_document()`
 - `views.S3BucketViewSet.perform_destroy()`

I also moved `_policy_name()` and `_policy_aen()` from the `services` module
into the `migrate_s3_access_roles_to_inline_policies` migration which
is the only place using them. This is not strictly necessary in theory
as I already run this migration but I preferred to keep this migration
functioning.

### Related PR
[Use one inline IAM policy per role](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/88)

### Ticket
Kind of part of: https://trello.com/c/2nuxrwiq/586-5-iam-policies-change-from-1-per-bucket-to-1-per-user

### DIFF
--- a/control_panel_api/management/commands/migrate_s3_access_roles_to_inline_policies.py
+++ b/control_panel_api/management/commands/migrate_s3_access_roles_to_inline_policies.py
@@ -8,11 +8,33 @@ from control_panel_api.models import (
     AppS3Bucket,
     UserS3Bucket,
 )
-# Using "private" method to get the policy ARN - no point in duplicating it
-from control_panel_api.services import _policy_arn
+
+
+READWRITE = 'readwrite'
+READONLY = 'readonly'
 
 
 logger = logging.getLogger(__name__)
+
+
+# Copied from `services`
+def _policy_name(bucket_name, readwrite=False):
+    """
+    Prefix the policy name with bucket name, postfix with access level
+    eg: dev-james-readwrite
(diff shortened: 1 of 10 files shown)


xoen:
(FYI: rebase `master`, no conflicts)

control_panel_api/management/commands/migrate_s3_access_roles_to_inline_policies.py
```diff
@@ -8,13 +8,35 @@
...
+
+
+READWRITE = 'readwrite'
+READONLY = 'readonly'
 
 
 logger = logging.getLogger(__name__)
 
 
+# Copied from `services`
```
andyhd: These comments should go

PR #93, opened 2018-01-31
Add tagging to data warehouse bucket on create

jmoz:
## What

Add tagging to data warehouse bucket on create.

Had to rewrite the test patch as that method does not work for multiple mock testing.

## How to review

1. Create a datawarehouse bucket and check the tags aws console

### DIFF
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -50,14 +50,14 @@ class AWSClient(object):
         """Put bucket tagging
 
         :param name: Bucket name
-        :param tags: List of tags (key, value) e.g. [('buckettype', 'datawarehouse')]
+        :param tags: Tags {key: value} e.g. {'buckettype': 'datawarehouse'}
         :type name: str
-        :type tags: list(tuple)
+        :type tags: dict
         """
         self._do('s3', 'put_bucket_tagging',
             Bucket=name,
             Tagging={
-                'TagSet': [{'Key': k, 'Value': v} for k, v in tags]
+                'TagSet': [{'Key': k, 'Value': v} for k, v in tags.items()]
             })
 
     def get_inline_policy_document(self, role_name, policy_name):
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -57,7 +57,7 @@ class AWSClient(object):
         self._do('s3', 'put_bucket_tagging',
             Bucket=name,
             Tagging={
-                'TagSet': [{'Key': k, 'Value': v} for k, v in tags.items()]
+                'TagSet': [{'Key': str(k), 'Value': str(v)} for k, v in tags.items()]
             })
 
     def get_inline_policy_document(self, role_name, policy_name):
--- a/control_panel_api/aws.py
+++ b/control_panel_api/aws.py
@@ -46,6 +46,20 @@ class AWSClient(object):
                 }
             })
 
+    def put_bucket_tagging(self, name, tags):
+        """Put bucket tagging
+
+        :param name: Bucket name
+        :param tags: List of tags (key, value) e.g. [('buckettype', 'datawarehouse')]
+        :type name: str
+        :type tags: list(tuple)
+        """
+        self._do('s3', 'put_bucket_tagging',
+            Bucket=name,
+            Tagging={
+                'TagSet': [{'Key': k, 'Value': v} for k, v in tags]
+            })
+
     def get_inline_policy_document(self, role_name, policy_name):
         if not self.enabled:
             return None
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -105,7 +105,7 @@ def create_bucket(bucket_name, is_data_warehouse):
     if is_data_warehouse:
         aws.put_bucket_tagging(
             bucket_name,
-            tags=[('buckettype', 'datawarehouse')]
+            tags={'buckettype': 'datawarehouse'}
         )
 
 
--- a/control_panel_api/services.py
+++ b/control_panel_api/services.py
@@ -91,7 +91,7 @@ def delete_role(role_name):
 
 
 @ignore_aws_exceptions
-def create_bucket(bucket_name):
+def create_bucket(bucket_name, is_data_warehouse):
     aws.create_bucket(
         bucket_name,
         region=settings.BUCKET_REGION,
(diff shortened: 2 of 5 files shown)


control_panel_api/aws.py
```diff
@@ -46,6 +46,20 @@ def put_bucket_logging(self, name, target_bucket, target_prefix):
                 }
             })
 
+    def put_bucket_tagging(self, name, tags):
+        """Put bucket tagging
+
+        :param name: Bucket name
+        :param tags: List of tags (key, value) e.g. [('buckettype', 'datawarehouse')]
+        :type name: str
+        :type tags: list(tuple)
```
xoen: (from my little understanding of) S3 tags are key/value pairs. 
So after calling `put_bucket_tagging` you may have something like this attached to the bucket:

```python
{
  'buckettype': 'datawarehouse',
  'other-key': 'other-value',
}
```

I think the `tags` param here should be a dictionary, e.g. you call this like this:

```python
aws.put_bucket_tagging('bucket-name', {'k1': 'v1', 'k2': 'v2'})
```

In case like the above use tuples can lead to confusion, e.g. if I see something like this `('k1', 'v1', 'k2', 'v2')` I may not catch the relationship between `k1`/`v1` and `k2`/`v2`.

(Obviously this means the `for` loop needs to be tweaked but it's a little price to pay for clarity IMHO)

control_panel_api/aws.py
```diff
@@ -46,6 +46,20 @@ def put_bucket_logging(self, name, target_bucket, target_prefix):
...
+        """
+        self._do('s3', 'put_bucket_tagging',
+            Bucket=name,
+            Tagging={
+                'TagSet': [{'Key': k, 'Value': v} for k, v in tags]
```
xoen: Maybe it's a problem for another day but as far as I can read from boto documentation the stuff inside `TagSet` needs to be a string. 

I'm not sure what could happen if they're not. We have few options here:
1) We don't do anything, after all we control the client code.
2) Just convert everything to string, e.g. calling `str(k)`/`str(v)`
3) We check the type and we blowup a `ValueError` or something like that (probably a bit too much)

`1)` may be alright but I think I'd prefer option `2)`, what do you reckon?

control_panel_api/services.py
```diff
@@ -175,6 +175,12 @@ def create_bucket(bucket_name):
         target_prefix=f"{bucket_name}/")
     aws.put_bucket_encryption(bucket_name)
 
+    if is_data_warehouse:
+        aws.put_bucket_tagging(
+            bucket_name,
+            tags=[('buckettype', 'datawarehouse')]
```
xoen: See other comment on `tags` argument (e.g. change to a dictionary).

control_panel_api/aws.py
jmoz: Good point, not sure why I went down that route.

control_panel_api/services.py
```diff
@@ -175,6 +175,12 @@ def create_bucket(bucket_name):
         target_prefix=f"{bucket_name}/")
     aws.put_bucket_encryption(bucket_name)
 
+    if is_data_warehouse:
+        aws.put_bucket_tagging(
+            bucket_name,
+            tags={'buckettype': 'datawarehouse'}
```
xoen: The OCD in me thinks `bucket-type`: `data-warehouse` :) - I need to double check, but it should be possible to use `-` here right?

control_panel_api/services.py
```diff
@@ -175,6 +175,12 @@ def create_bucket(bucket_name):
         target_prefix=f"{bucket_name}/")
     aws.put_bucket_encryption(bucket_name)
 
+    if is_data_warehouse:
```
xoen: One thing I wonder is if we should tag the non-data-warehouse buckets (with `{'bucket-type': 'app'}`) or not. I suppose the "problem" with that would be that then some app buckets have this tag but some they don't.

It probably doesn't matter at this stage as we don't use these tags anyway.

control_panel_api/tests/test_services.py
```diff
@@ -19,7 +19,7 @@
 )
 
 
-@patch.object(aws, 'client', MagicMock())
+@patch('control_panel_api.aws.aws.client')
```
xoen: Just curious here, any reason why this changed? I see all tests now need to receive a `mock_client`.

Is it to make this more explicit that you're using a mock instead of calling `aws.client.return_value [...]`?

control_panel_api/aws.py
jmoz: I can `str()` it but also note we don't cast on any of the other calls

control_panel_api/aws.py
```diff
@@ -46,6 +46,20 @@ def put_bucket_logging(self, name, target_bucket, target_prefix):
...
 
+    def put_bucket_tagging(self, name, tags):
+        """Put bucket tagging
+
+        :param name: Bucket name
+        :param tags: Tags {key: value} e.g. {'buckettype': 'datawarehouse'}
+        :type name: str
+        :type tags: dict
+        """
+        self._do('s3', 'put_bucket_tagging',
```
xoen: Bear in mind that merge this would break if you don't update the CP-API IAM role so that it has the permission to perform this action.

You need something like [this](https://github.com/ministryofjustice/analytics-platform-ops/pull/123) with the right S3 action and to do a `terraform apply` in dev/alpha.

control_panel_api/aws.py
```diff
@@ -57,7 +57,7 @@ def put_bucket_tagging(self, name, tags):
         self._do('s3', 'put_bucket_tagging',
             Bucket=name,
             Tagging={
-                'TagSet': [{'Key': k, 'Value': v} for k, v in tags.items()]
+                'TagSet': [{'Key': str(k), 'Value': str(v)} for k, v in tags.items()]
```
xoen: 👍

control_panel_api/tests/test_services.py
jmoz: I wrote in PR description.

Basically the MagicMock() in the patch is good for just mocking out a function and ignoring it.  However if you really want to mock and test the mocked function it's the wrong one as the mock is not cleared between test calls, so using the patch and inject is better.

xoen: Oh I missed that sentence in the PR description 🤦‍♂️ 

Nice one, didn't know that (that's why I asked), so it's basically to avoid state leaking between tests, cheers 👍

jmoz:
Added a branch on ops with the tagging permission, tested with terraform apply and then branch merged, then applied on dev/alpha.

jmoz:
Force pushed a rebase

xoen:
Just saw this. Good stuff 👍

PR #94, opened 2018-01-31
Handle better update of S3 policies when `write_to_cluster` is disabled

xoen:
### What

When this flag is disabled the policy `get_role_policy()` will return
`None` (as expected because we don't talk to AWS).
The problem is that we try to access the `PolicyDocument` of `None` and this
is not obviously working.

**NOTE**: This only affects us locally when `write_to_cluster` is disabled.

### Ticket
https://trello.com/c/MxSKSCuR/608-1-fix-grant-revoke-access-locally-when-writetocluster-is-false


xoen:
That was quick @andyhd - thanks :D

PR #95, opened 2018-02-07
Refactoring: Moved tests for permissions (some) and filters in separate files

xoen:
### What

We used to have massive test files for filters and permissions.
It's cleaner to have different files for different classes tested as this also reduces the chances of conflicts as more people work on close parts of the code (as it is the case currently).

#### **NOTES**
- I did **not** change any of the code, only moved
- I think it's better to have this done before the actual permissions changes so we don't have a massive PR with a diff difficult to follow

### Ticket
Preparation work for this and other permission-related tickets:

https://trello.com/c/JALNEbcN/630-3-buckets-permissions-normal-users-can-create-s3-buckets-becomes-admin


control_panel_api/tests/__init__.py
```diff
@@ -1,5 +1,8 @@
 from django.conf import settings
 
+from control_panel_api.tests.filters import *
+from control_panel_api.tests.permissions import *
```
andyhd: Don't do this `import *` it is an antipattern. You must use explicit imports.

xoen: Should I import like this instead?

```python
import control_panel_api.tests.filters
import control_panel_api.tests.permissions
```

jmoz: Requiring individual testcase imports is going to increase the chance of tests being written and not executed.

So it's a toss up of not restructuring them, or `import *` I'd say.

xoen: Changed: daf525ae12e33a09defa53e1a26324777f581859

control_panel_api/tests/__init__.py
```diff
@@ -1,5 +1,8 @@
 from django.conf import settings
 
+import control_panel_api.tests.filters
```
jmoz: If this import works then disregard what I said - I presumed `... import AppTestCase` etc would be required.

xoen:
As discussed the `import` were not actually necessary: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/95/commits/ed2f068ec66f4b5efb1564774fdd8addf2346962

PR #96, opened 2018-02-08
Normal users can create/retrieve/list S3 buckets

xoen:
### What
Normal users can now:
- Create S3 buckets
- List/Retrieve them, user will see only S3 buckets he has access to

### Background
The authoritative source of true when it comes to which users can access what is the list of `UserS3Bucket` records.
Each of this record grant a given user access to an S3 bucket with a certain access level (readonly/readwrite).

### NOTES
- Django filters are applied before permissions, therefore if a record is filtered out by the filter class the user will not be able to perform any operation on that record (e.g. he would get a `404`)
- All others permissions are maintained as before, e.g. normal users cannot update or delete S3 buckets (tests cover these and did not change)

In order to determine if a user has the ability to see an S3 bucket we check if the user has a record for this S3 bucket for his user (with any access level).

### Ticket
https://trello.com/c/JALNEbcN/630-3-buckets-permissions-normal-users-can-create-s3-buckets-becomes-admin

### DIFF
--- a/control_panel_api/permissions.py
+++ b/control_panel_api/permissions.py
@@ -51,8 +51,22 @@ class AppPermissions(IsSuperuser):
     pass
 
 
-class S3BucketPermissions(IsSuperuser):
-    pass
+class S3BucketPermissions(BasePermission):
+    """
+    - Superusers can do anything
+    - Normal users can only list buckets they have access to
+
+    NOTE: Filters are applied before permissions
+    """
+
+    def has_permission(self, request, view):
+        if is_superuser(request.user):
+            return True
+
+        if request.user.is_anonymous():
+            return False
+
+        return view.action in ('list')
 
 
 class UserPermissions(BasePermission):
--- a/control_panel_api/permissions.py
+++ b/control_panel_api/permissions.py
@@ -54,7 +54,7 @@ class AppPermissions(IsSuperuser):
 class S3BucketPermissions(BasePermission):
     """
     - Superusers can do anything
-    - Normal users can only list buckets they have access to
+    - Normal users can create S3 buckets and list buckets they have access to
 
     NOTE: Filters are applied before permissions
     """
@@ -66,7 +66,7 @@ class S3BucketPermissions(BasePermission):
         if request.user.is_anonymous():
             return False
 
-        return view.action in ('list')
+        return view.action in ('create', 'list')
 
 
 class UserPermissions(BasePermission):
--- a/control_panel_api/permissions.py
+++ b/control_panel_api/permissions.py
@@ -54,7 +54,8 @@ class AppPermissions(IsSuperuser):
 class S3BucketPermissions(BasePermission):
     """
     - Superusers can do anything
-    - Normal users can create S3 buckets and list buckets they have access to
+    - Normal users can create S3 buckets and list/retrieve buckets they have
+      access to
 
     NOTE: Filters are applied before permissions
     """
@@ -66,7 +67,7 @@ class S3BucketPermissions(BasePermission):
         if request.user.is_anonymous():
             return False
 
-        return view.action in ('create', 'list')
+        return view.action in ('create', 'list', 'retrieve')
 
 
 class UserPermissions(BasePermission):
--- a/control_panel_api/tests/filters/test_s3bucket_filter.py
+++ b/control_panel_api/tests/filters/test_s3bucket_filter.py
@@ -7,29 +7,27 @@ from rest_framework.test import APITestCase
 class S3BucketFilterTest(APITestCase):
 
     def setUp(self):
-        # Create users
-        self.superuser = mommy.make(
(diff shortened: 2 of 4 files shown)


xoen:
[(3) Buckets Permissions: "normal" users can create S3 buckets (becomes admin)](https://trello.com/c/JALNEbcN/630-3-buckets-permissions-normal-users-can-create-s3-buckets-becomes-admin)

control_panel_api/permissions.py
```diff
@@ -51,8 +51,23 @@ class AppPermissions(IsSuperuser):
...
+    """
+
+    def has_permission(self, request, view):
+        if is_superuser(request.user):
+            return True
+
+        if request.user.is_anonymous():
+            return False
+
+        return view.action in ('create', 'list', 'retrieve')
```
jmoz: What about deleting their bucket?

control_panel_api/tests/filters/test_s3bucket_filter.py
```diff
@@ -17,6 +17,12 @@ def setUp(self):
             "control_panel_api.S3Bucket", name="test-bucket-1")
         self.s3_bucket_2 = mommy.make(
             "control_panel_api.S3Bucket", name="test-bucket-2")
+        # Normal user has access to s3_bucket_1
```
jmoz: These comments aren't necessary. Intention can be deduced from name etc.  And the comments above also.

control_panel_api/tests/filters/test_s3bucket_filter.py
```diff
@@ -17,6 +17,12 @@ def setUp(self):
...
+        self.normal_user.users3buckets.create(
```
jmoz: Can you keep it consistent and use the smart fixtures library `mommy.make`

control_panel_api/tests/permissions/test_s3bucket_permissions.py
```diff
@@ -28,18 +29,20 @@ def setUp(self):
             "control_panel_api.S3Bucket", name="test-bucket-1")
         self.s3bucket_2 = mommy.make(
             "control_panel_api.S3Bucket", name="test-bucket-2")
+        # Normal user has access to s3bucket_1
```
jmoz: Don't need comment.

control_panel_api/permissions.py
xoen: They can't. Not part of the requirements and deleting is a dangerous operation anyway so denying unless we **really really** find it's a necessity.

jmoz: 👍

control_panel_api/tests/permissions/test_s3bucket_permissions.py
xoen: I still think that these comments can help understand the tests but I'll remove for the love of peace 👍

jmoz:
LGTM 🐈

xoen:
Chairs

xoen:
FYI: Tested in `alpha` and works as expected 👍

control_panel_api/tests/permissions/test_s3bucket_permissions.py
```diff
@@ -18,28 +19,32 @@ class S3BucketPermissionsTest(APITestCase):
...
-        self.normal_user = mommy.make(
-            'control_panel_api.User', is_superuser=False)
-        # Create some S3 buckets
-        self.s3bucket_1 = mommy.make(
-            "control_panel_api.S3Bucket", name="test-bucket-1")
-        self.s3bucket_2 = mommy.make(
-            "control_panel_api.S3Bucket", name="test-bucket-2")
+
+        self.superuser = mommy.make('control_panel_api.User',
+                                    is_superuser=True)
```
andyhd: Just a pet peeve, as this indentation style is ok according to PEP8, but I think indenting like this is better:
```python
self.superuser = mommy.make(
    'control_panel_api.User',
    is_superuser=True)
```
It keeps indentation levels regular throughout the code, instead of having indentation depend on the length of the variable name. Also, if you change the length of the variable name for whatever reason, you won't have to re-indent the following line(s).

xoen: That indentation is coming straight from my PEP8 plugin BTW.

xoen: Re-read your comment, I see. I can change these 👍

xoen: (I just tried to now think too much about it as it was coming from the PEP8 plugin and I didn't want to question the "word" :P  )

xoen: FYI: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/97/commits/91de3d8b29bdb634723d93beeca78c10f74e0ef6

PR #97, opened 2018-02-08
Normal users can list/retrieve UserS3Buckets

xoen:
### What

Allow normal users to see/list `UserS3Bucket`s

Normal users need to have visibility over which user S3 buckets they have access to and also other users having access to same data they have access to.

### How
As per requirements, a user will be able to see all `UserS3Bucket` records for buckets he can access to:
1. Given a list of S3 buckets a user has access to (there is a `UserS3Bucket`
   record for his user)
2. That user can see all `UserS3Bucket` records for buckets above (including `UserS3Bucket` records for other users that have access to same data)

### Notes
`UserS3BucketPermissions` is not tested as James is working on this class as well and he'll add tests.
I added this class mainly to test filtering is working as expected (and without this normal users will just get a `403`).

### Related PRs
This work is based on the changes in https://github.com/ministryofjustice/analytics-platform-control-panel/pull/96 ~I'm waiting for that to be reviewed/merged before I can go ahead with this one.~

### Ticket
https://trello.com/c/21llViWJ/632-3-buckets-permissions-normal-users-can-see-buckets-they-have-access-to-and-other-users-that-have-access-to-the-same-bucket

### DIFF
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -6,6 +6,7 @@ See: http://www.django-rest-framework.org/api-guide/filtering/#custom-generic-fi
 
 from django_filters.rest_framework import DjangoFilterBackend
 
+from control_panel_api.models import S3Bucket
 from control_panel_api.permissions import is_superuser
 
 
@@ -66,3 +67,27 @@ class UserFilter(DjangoFilterBackend):
             return queryset
 
         return queryset.filter(pk=request.user.pk)
+
+
+class UserS3BucketFilter(DjangoFilterBackend):
+    """
+    Filter to get visible UserS3Bucket records.
+
+    - Superusers see everything
+    - Normal users see only records for buckets they have access.
+      Normal users can also see records for other users that have access
+      to same data (but permissions will enforce they can't change these
+      records unless they're admin)
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        queryset = super().filter_queryset(request, queryset, view)
+
+        if is_superuser(request.user):
+            return queryset
+
+        accessible_buckets = S3Bucket.objects.filter(
+            users3buckets__user=request.user,
+        )
+
+        return queryset.filter(s3bucket=accessible_buckets)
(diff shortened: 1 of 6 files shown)


xoen:
[(3) Buckets Permissions: "normal" users can see buckets they have access to and other users that have access to the same bucket](https://trello.com/c/21llViWJ/632-3-buckets-permissions-normal-users-can-see-buckets-they-have-access-to-and-other-users-that-have-access-to-the-same-bucket)

control_panel_api/filters.py
```diff
@@ -66,3 +67,27 @@ def filter_queryset(self, request, queryset, view):
...
+      records unless they're admin)
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        queryset = super().filter_queryset(request, queryset, view)
+
+        if is_superuser(request.user):
+            return queryset
+
+        accessible_buckets = S3Bucket.objects.filter(
```
jmoz: As our base queryset is `UserS3Bucket.objects.all()` I think `queryset.filter(user=request.user)` should suffice.  

Not tested, just guessing.

xoen: That was exactly the query I wrote the when I started to play with it but...

This is a bit of a sneaky one. Users need to see not only `UserS3Bucket` where the user is them but ALSO all other `UserS3Bucket` records for buckets they have access to. That's why the query is like that.

Let me know if it makes sense.

control_panel_api/permissions.py
```diff
@@ -91,6 +91,26 @@ def has_object_permission(self, request, view, obj):
 
         return request.user == obj
 
+# NOTE: James is doing some work related to this class as well,
+#       that's why I'm not testing it - he'll add tests.
+#       Only thing relevant for filtering work is `list`/`retrive` actions
```
andyhd: Don't put notes like this in the code. It doesn't document the code so it doesn't belong here. Add a comment in the PR instead.

xoen:
FYI: Deployed and double check in `alpha` - working as expected 👍

PR #98, opened 2018-02-14
UserS3Bucket permissions

jmoz:
## What

This implements the permissions for users3bucket and also merges Aldo's filter work.

## How to review

1. Run the tests

https://trello.com/c/d7ztSVXv/635-3-users3bucket-permissions-a-normal-user-can-create-a-new-record-if-that-user-is-an-existing-admin-of-the-s3bucket

### DIFF
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -49,7 +49,7 @@ class S3BucketFilter(DjangoFilterBackend):
         if is_superuser(request.user):
             return queryset
 
-        return queryset.filter(users3buckets__user=request.user)
+        return queryset.accessible_by(request.user)
 
 
 class UserFilter(DjangoFilterBackend):
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -86,8 +86,6 @@ class UserS3BucketFilter(DjangoFilterBackend):
         if is_superuser(request.user):
             return queryset
 
-        accessible_buckets = S3Bucket.objects.filter(
-            users3buckets__user=request.user,
-        )
+        accessible_buckets = S3Bucket.objects.accessible_by(request.user)
 
-        return queryset.filter(s3bucket=accessible_buckets)
+        return queryset.filter(s3bucket__in=accessible_buckets)
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -115,9 +115,9 @@ class UserApp(TimeStampedModel):
         ordering = ('id',)
 
 
-class S3BucketManager(models.Manager):
+class S3BucketQuerySet(models.QuerySet):
     def accessible_by(self, user):
-        return self.get_queryset().filter(
+        return self.filter(
             users3buckets__user=user,
         )
 
@@ -136,7 +136,7 @@ class S3Bucket(TimeStampedModel):
     created_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
     is_data_warehouse = models.BooleanField(default=False)
 
-    objects = S3BucketManager()
+    objects = S3BucketQuerySet.as_manager()
 
     class Meta:
         ordering = ('name',)
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -115,6 +115,14 @@ class UserApp(TimeStampedModel):
         ordering = ('id',)
 
 
+class S3BucketManager(models.Manager):
+    def accessible_by_admin(self, user):
+        return self.get_queryset().filter(
+            users3buckets__user=user,
+            users3buckets__is_admin=True,
+        )
+
+
 class S3Bucket(TimeStampedModel):
     name = models.CharField(unique=True, max_length=63, validators=[
         validators.validate_env_prefix,
@@ -123,6 +131,8 @@ class S3Bucket(TimeStampedModel):
     ])
     created_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
     is_data_warehouse = models.BooleanField(default=False)
+
+    objects = S3BucketManager()
 
     class Meta:
         ordering = ('name',)
@@ -272,3 +282,6 @@ class UserS3Bucket(AccessToS3Bucket):
 
     def aws_role_name(self):
         return self.user.iam_role_name
+
+    def user_is_admin(self, user):
+        return self.user == user and self.is_admin
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -116,9 +116,13 @@ class UserApp(TimeStampedModel):
 
 
 class S3BucketManager(models.Manager):
-    def accessible_by_admin(self, user):
+    def accessible_by(self, user):
         return self.get_queryset().filter(
             users3buckets__user=user,
+        )
+
+    def accessible_by_admin(self, user):
+        return self.accessible_by(user).filter(
             users3buckets__is_admin=True,
         )
 
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -121,7 +121,7 @@ class S3BucketQuerySet(models.QuerySet):
             users3buckets__user=user,
         )
 
-    def accessible_by_admin(self, user):
+    def administered_by(self, user):
         return self.accessible_by(user).filter(
             users3buckets__is_admin=True,
         )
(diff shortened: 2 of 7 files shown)


control_panel_api/tests/test_permissions.py
```diff
@@ -292,7 +283,6 @@ def test_normal_user_cant_operate_on_namespaces_with_same_prefix(self):
 
 
 class ToolDeploymentPermissionsTest(APITestCase):
-
```
xoen: Not that it matters from the Python interpreter perspective but I thought methods inside a class were surrounded by a blank line:

> Method definitions inside a class are surrounded by a single blank line.

[PEP8 on blank lines](https://www.python.org/dev/peps/pep-0008/#blank-lines)

control_panel_api/tests/test_permissions.py
```diff
@@ -7,25 +7,21 @@
...
 )
 from rest_framework.test import APITestCase
 
-from control_panel_api.aws import aws
 from control_panel_api.models import AppS3Bucket
 
 
-@patch.object(aws, 'client', MagicMock())
+@patch('control_panel_api.aws.aws.client', MagicMock())
 class AppPermissionsTest(APITestCase):
```
xoen: I wonder if it's a good time to move these last 4 test case classes in separate files under tests/permissions as well.

control_panel_api/filters.py
```diff
@@ -86,8 +86,6 @@ def filter_queryset(self, request, queryset, view):
         if is_superuser(request.user):
             return queryset
 
-        accessible_buckets = S3Bucket.objects.filter(
-            users3buckets__user=request.user,
-        )
+        accessible_buckets = S3Bucket.objects.accessible_by(request.user)
 
-        return queryset.filter(s3bucket=accessible_buckets)
+        return queryset.filter(s3bucket__in=accessible_buckets)
```
xoen: Good stuff on fixing this and the tests, sorry I thought Django ORM would behave similarly to Rails' one by converting this in an SQL `IN` statement. Thanks @jmoz

control_panel_api/tests/permissions/test_users3bucket_permissions.py
```diff
@@ -0,0 +1,165 @@
...
+            is_admin=True,
+        )
+        self.normal_user_other_bucket_non_admin = mommy.make(
+            "control_panel_api.UserS3Bucket",
+            user=self.normal_user,
+            s3bucket=self.s3bucket_2,
+            access_level=AccessToS3Bucket.READWRITE,
+            is_admin=False,
+        )
+        self.other_user_normal_user_has_admin = mommy.make(
```
xoen: You may want to call this variable `other_user_admin` for consistency with `normal_user_admin`.

control_panel_api/tests/permissions/test_users3bucket_permissions.py
```diff
@@ -0,0 +1,165 @@
...
+            "control_panel_api.UserS3Bucket",
+            user=self.user_1,
+            s3bucket=self.s3bucket_1,
+            access_level=AccessToS3Bucket.READWRITE,
+            is_admin=True,
+        )
+
+    def test_create(self):
+        fixtures = [
```
xoen: Very nice 👍

control_panel_api/tests/permissions/test_users3bucket_permissions.py
```diff
@@ -0,0 +1,165 @@
...
+    def test_delete_normal_user_owner_non_admin_403(self):
+        self.client.force_login(self.normal_user)
+
+        response = self.client.delete(
+            reverse('users3bucket-detail',
+                    (self.normal_user_other_bucket_non_admin.id,)))
+        self.assertEqual(HTTP_403_FORBIDDEN, response.status_code)
+
+    def test_update(self):
+        fixtures = [
```
xoen: Nice one 👍

control_panel_api/models.py
```diff
@@ -115,6 +115,18 @@ class Meta:
         ordering = ('id',)
 
 
+class S3BucketQuerySet(models.QuerySet):
+    def accessible_by(self, user):
+        return self.filter(
+            users3buckets__user=user,
+        )
+
+    def accessible_by_admin(self, user):
```
xoen: Maybe name could be `administered_by()` as "accessible" is a given when you're an admin.

control_panel_api/models.py
```diff
@@ -124,6 +136,8 @@ class S3Bucket(TimeStampedModel):
     created_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
     is_data_warehouse = models.BooleanField(default=False)
 
+    objects = S3BucketQuerySet.as_manager()
```
xoen: Didn't know this trick. I quite like this approach to add custom queries 👍

control_panel_api/permissions.py
```diff
@@ -93,24 +95,35 @@ def has_object_permission(self, request, view, obj):
...
 
         if request.user.is_anonymous():
             return False
 
-        return view.action in ('list', 'retrieve')
+        if view.action != 'create':
+            return True
 
+        serializer = UserS3BucketSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
```
xoen: Validate the data is not a "permissions" responsibility so it shouldn't be done here (it's already done in the views)

control_panel_api/tests/test_permissions.py
jmoz: I auto reformatted the page and for some reason it doesn't add blank to after class def.

control_panel_api/filters.py
jmoz: 👍

control_panel_api/tests/permissions/test_users3bucket_permissions.py
jmoz: Yeh it's named differently, because it's from the perspective of the normal_user. This one an other user owns it, but the bucket related the normal user has admin access.

control_panel_api/tests/permissions/test_users3bucket_permissions.py
jmoz: 👍

control_panel_api/tests/permissions/test_users3bucket_permissions.py
jmoz: 👍

control_panel_api/models.py
xoen: One thing to watch out for is that this double application of filter cause a double `JOIN` in the resulting SQL query. e.g. for each of the `UserS3Bucket` records with user/is_admin matching there will be two results.

As far as I can tell we only use this in an if in the `UserS3BucketPermissions` class so this issue should not be user-facing (e.g. users shouldn't see duplicated entries when listing `UserS3Bucket`).

A possible solution could be to write this method as:

```
def administered_by(self, user):
        return self.filter(
            users3buckets__user=user,
            users3buckets__is_admin=True,
        )
```

I know this is a bit of a duplication. Maybe there is a way to tell django to return unique results.

control_panel_api/permissions.py
jmoz: Without this the bad request wouldn't even get to the view to be checked, it would just error out with a KeyError or type mismatch depending on how bad the submitted data is.

This is sort of a chicken and egg situation.  Still need to check the data as this permission check relies on existing data when creating.

control_panel_api/permissions.py
```diff
@@ -93,24 +95,35 @@ def has_object_permission(self, request, view, obj):
...
 
-class ToolDeploymentPermissions(BasePermission):
+        accessible_buckets = S3Bucket.objects.administered_by(request.user)
+
+        if serializer.validated_data['s3bucket'] in accessible_buckets:
```
xoen: We could avoid importing/using the serialiser at all by changing the `if` to query for existence, e.g.:

```python
# "ugly" proof of concept
if accessible_buckets.filter(pk=request.data['s3bucket']).exists():
    return True
```

Making a query has also the benefit of being more efficient for users with access to a large number of S3 buckets (maybe add something like that to to the `S3BucketQuerySet`? I don't know)

control_panel_api/models.py
jmoz: Oh damn, I thought it would merge them.  I'll go back to original implementation with slight dupe code.

control_panel_api/permissions.py
xoen: I see what you mean, potentially the input data may not have an `s3bucket` field, etc...

PR #100, opened 2018-02-20
S3Bucket deep link

jmoz:
## What

Return the location_url from the s3bucket creation.

Add missing test for tagging.

## How to review

1. Create a bucket and check response.

https://trello.com/c/fs2XcWKe/652-s3bucket-deep-links


control_panel_api/tests/test_models.py
```diff
@@ -192,13 +192,18 @@ def test_arn(self):
 
     @patch('control_panel_api.services.create_bucket')
     def test_bucket_create(self, mock_create_bucket):
+        url = 'http://foo.com/'
```
andyhd: No need to change this unless you want to, but it's always best to use `example.com` for test domains. `foo.com` is a real domain.

jmoz: Good point - bad habit.

PR #106, opened 2018-03-16
App Customers (Auth0 group members)

jmoz:
## What

Endpoint to manage app customers (create and delete), via auth0 authorization extension using groups and users.

## How to review

1. Run tests.
2. Create an app, then add customers to the app.
3. Check auth0 authorization extension groups and users.

https://trello.com/c/eI4eInCe/430-add-ability-to-see-list-of-emails-that-have-access-to-an-app-and-change

### DIFF
--- a/control_panel_api/tests/test_views.py
+++ b/control_panel_api/tests/test_views.py
@@ -343,7 +343,7 @@ class AppViewTest(AuthenticatedClientMixin, APITestCase):
             "baz": "bat",
         }]
 
-        response = self.client.get(reverse('app-customers', (self.fixture.id,)))
+        response = self.client.get(reverse('appcustomers-list', (self.fixture.id,)))
 
         self.assertEqual(HTTP_200_OK, response.status_code)
         mock_get_group_members.assert_called()
--- a/control_panel_api/tests/test_views.py
+++ b/control_panel_api/tests/test_views.py
@@ -28,6 +28,7 @@ from control_panel_api.tests.test_authentication import (
     build_jwt_from_user,
     mock_get_keys,
 )
+from moj_analytics.auth0_client import Group, User as Auth0User
 
 
 class AuthenticatedClientMixin(object):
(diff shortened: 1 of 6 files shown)


moj_analytics/auth0_client.py
```diff
@@ -278,9 +282,15 @@ def add_users(self, users):
         if 'error' in response:
             raise AddGroupMemberError(response['message'])
 
-        print('Added users({users}) to group({group})'.format(
-            users=', '.join(user['email'] for user in users),
-            group=self['name']))
+    def delete_users(self, users):
+        response = self.api.request(
+            'DELETE',
+            'groups/{_id}/members'.format(**self),
```
andyhd: Should this be changed to `f'groups/{self._id}/members',`?

jmoz: Ta 👍

control_panel_api/tests/test_views.py
```diff
@@ -28,6 +28,7 @@
     build_jwt_from_user,
     mock_get_keys,
 )
+from moj_analytics.auth0_client import Group, User as Auth0User
```
xoen: You may also want to import `Group` as `Auth0Group` for consistency and clarity.

control_panel_api/views.py
```diff
@@ -130,20 +131,46 @@ def perform_destroy(self, instance):
 
         instance.aws_delete_role()
 
-    @detail_route(permission_classes=[IsSuperuser])
-    def customers(self, request, pk=None):
+
+class AppCustomersAPIView(GenericAPIView):
+    queryset = App.objects.all()
+    serializer_class = GroupMemberSerializer
```
xoen: I would also consider to rename `Group`/`GroupMember` to avoid any possible confusion with Django's Groups. `Auth0GroupSerializer`/`Auth0GroupMemberSerializer` may be a bit long winded but less ambiguous.

jmoz: Oh yes I totally missed that, thanks.

control_panel_api/tests/test_views.py
```diff
@@ -361,6 +369,87 @@ def test_customers(self, mock_get_group_members):
...
+    def test_post_create_user(self, mock_api):
+        api_create_user_response = {'user_id': 123}
+
+        mock_api.authorization.get.return_value = None
+        mock_api.management.create.return_value = api_create_user_response
+
+        data = {'email': 'foo@example.com'}
+
+        response = self.client.post(
+            reverse('appcustomers-list', (self.fixture.id,)), data)
```
xoen: Just an observation, wouldn't be better to call this variable `app` instead of `fixture`?

PR #109, opened 2018-03-21
Lambda functions "data" migrations

xoen:
### What
We want to dismiss lambda functions which currently handles GitHub events to create team buckets, etc...

This means we need to add the records for these buckets, who has access to them, etc...to the control panel DB.

#### 1. S3 Buckets
The first of the migrations (`migrate_lambdas_data_1_buckets.py`) loops through the IAM policies which are attached to at least an IAM role and for each of them:
1. checks if 
   1. policy name starts with the current env name
   2. policy name ends with `-readwrite`, as lambda functions only used these policies
   3. ignore policies for app data buckets ("app buckets" are not "data warehouse buckets" and they're added directly via CP)
2. checks if S3 bucket already exists in CP
3. if not, it creates the `S3Bucket` record with `is_data_warehouse` flag set to true.

#### 2. Create `UserS3Bucket` records
Once we have all the data warehouse `S3Bucket` records in the DB we need to add the `UserS3Bucket` records so that CP-API knows who can access what.

1. For each of the users
2. We determine the user IAM role name and list the IAM policies attached to it
3. For each of the relevant policies
3a. We filter out policies for other environments, non-readwrite policies, policies for "app" buckets and in general ignore any IAM policy with a name not matching the template as these are probably policies for things like AWS Glue, etc...
4. For each of these IAM policy (named something like `dev-*-readwrite`) we create a `UserS3Bucket` record for the corresponding user with access level `readwrite` and to the corresponding S3 bucket (which at this point, after migration above, should already have a record in DB).

#### 3. Update roles to grant access as in `UserS3Bucket` records
Once we have these new `UserS3Bucket` records in the DB, we'll need to update the IAM roles to update their `s3-access` inline policy to grant access to these S3 bucket they had access before via managed policies attached to their role.

We already wrote and used [this migration](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/control_panel_api/management/commands/migrate_s3_access_roles_to_inline_policies.py) when we initially moved from managed policies to a single inline policy so we're going to re-use it.

This migration loops through `UserS3bucket` (and `AppS3Bucket`) records and for each of them:
1. read the corresponding IAM role `s3-access` inline policy
2. grant access to the S3 bucket
3. writes the inline policy back to the role
3. Detaches the corresponding managed policy

#### About `DRYRUN` option
These migrations have a `DRYRUN` option which allow them to be run without actually having side effects. This is very useful to preview what changes the migrations will make before actually migrating the data (by looking at the logs)

```bash
$ DRYRUN=true python3 manage.py ...
```

#### Duplication / Design
I preferred to split this migration/task in smaller migration to avoid having a massive complex migration doing a lot of things. Yes, the steps are related but relatively self contained.

As a result, there are a couple of very simple utility functions (`_is_eligible()` and `_bucket_name()`) which are duplicated in migration 1 and 2. I don't think it's worth adding some sort of shared library for this. And I also like the idea of these migration being self contained so that changes outside them are not affecting them (e.g. we may need to run them again later at some point).

These migrations should also be idempotent.

### Ticket
https://trello.com/c/Me2RD6ek/696-5-script-that-copies-data-access-permissions-from-existing-lambda-created-policies-to-to-s3-access-inline-policy-and-adds-these

### DIFF
--- a/control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
+++ b/control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
@@ -1,0 +1,58 @@
+import json
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+from django.core.management.base import BaseCommand
+
+from control_panel_api.models import User
+
+
+READWRITE = 'readwrite'
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    NOTE: Needs permission to perform `iam:ListAttachedRolePolicies` action
+    """
+
+
+    help = ("For each of the users in the DB, it collects its attached IAM "
+            "policies and prints this data to stdout")
+
+    def handle(self, *args, **options):
+        iam = boto3.client('iam')
+
+        dump = {}
+
+        for user in User.objects.all():
+            username = user.username
+            role_name = user.iam_role_name
+
+            dump[username] = {
+                'iam_role_name': role_name,
+                'attached_policies': [],
+            }
+
+            try:
+                policies = iam.list_attached_role_policies(
+                    RoleName=role_name,
+                    MaxItems=1000,
+                )
+                if policies:
+                    dump[username]['attached_policies'] = policies["AttachedPolicies"]
+            except ClientError as e:
+                if e.response['Error']['Code'] in ('NoSuchEntity', 'ValidationError'):
+                    logger.warning(
+                        f'Error reading policies from IAM role "{role_name}": '
+                        f'user "{user.username}" ("{user.pk}"): {e}'
+                    )
+                    dump[username]['error'] = str(e)
+                    continue
+                else:
+                    raise e
+
+        print(json.dumps(dump, indent=4))
--- a/control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
+++ b/control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
@@ -8,20 +8,18 @@ from django.core.management.base import BaseCommand
 from control_panel_api.models import User
 
 
-READWRITE = 'readwrite'
-
-
 logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
     """
-    NOTE: Needs permission to perform `iam:ListAttachedRolePolicies` action
+    For each of the users in the DB, it collects its attached IAM policies
+    and prints this data to stdout.
+
+    NOTE: Needs permission to perform `iam:ListAttachedRolePolicies` action.
     """
 
-
-    help = ("For each of the users in the DB, it collects its attached IAM "
-            "policies and prints this data to stdout")
+    help = __doc__
 
     def handle(self, *args, **options):
         iam = boto3.client('iam')
--- a/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
+++ b/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
@@ -15,7 +15,7 @@ READWRITE = 'readwrite'
 logger = logging.getLogger(__name__)
 
 
-def _eligible(policy_name):
+def _is_eligible(policy_name):
     return (policy_name.startswith(f'{settings.ENV}-') and
             not policy_name.startswith(f'{settings.ENV}-app-') and
             policy_name.endswith(READWRITE))
@@ -44,7 +44,7 @@ class Command(BaseCommand):
         for policy in results['Policies']:
             policy_name = policy['PolicyName']
 
-            if _eligible(policy_name):
+            if _is_eligible(policy_name):
                 bucket_name = _bucket_name(policy_name)
 
                 if not S3Bucket.objects.filter(name=bucket_name).exists():
--- a/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
+++ b/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
@@ -1,41 +1,32 @@
 import logging
-import os
 
 import boto3
-from django.core.management.base import BaseCommand
 
 from control_panel_api.management.commands.migrate_lambdas_data_utils import (
     bucket_name,
     is_eligible,
 )
+from control_panel_api.management.commands.dry_runnable import DryRunnable
 from control_panel_api.models import S3Bucket
 
 
 logger = logging.getLogger(__name__)
 
 
-class Command(BaseCommand):
+class Command(DryRunnable):
     """
-    NOTE: Needs permission to perform `iam:ListPolicies` action
+    Add records for S3 buckets created by AWS Lambda functions.
+
+    NOTE: Needs permission to perform `iam:ListPolicies` action.
     """
 
-
-    help = "Add records for S3 buckets created by AWS Lambda functions."
-
-    def add_arguments(self, parser):
-        parser.add_argument(
-            '--dry-run',
-            dest='dry_run',
-            help="Run command without side effects",
-            default=False,
-            action='store_true',
-        )
+    help = __doc__
 
     def handle(self, *args, **options):
         iam = boto3.client('iam')
         results = iam.list_policies(
             Scope='Local',  # Customer managed policies only
-            OnlyAttached=True,  # Ignore non-attached policies
+            OnlyAttached=True,  # Ignore IAM policies not used
             MaxItems=1000,
         )
 
--- a/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
+++ b/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import os
 
 import boto3
 from django.conf import settings
@@ -7,7 +8,7 @@ from django.core.management.base import BaseCommand
 
 from control_panel_api.models import S3Bucket
 
-
+DRYRUN = os.environ.get('DRYRUN', 'false').lower() == 'true'
 READWRITE = 'readwrite'
 
 
@@ -25,12 +26,16 @@ def _bucket_name(policy_name):
 
 
 class Command(BaseCommand):
+    """
+    NOTE: Needs permission to perform `iam:ListPolicies` action
+    """
+
+
     help = "Add records for S3 buckets created by AWS Lambda functions."
 
     def handle(self, *args, **options):
-        iam_client = boto3.client('iam')
-
-        results = iam_client.list_policies(
+        iam = boto3.client('iam')
+        results = iam.list_policies(
             Scope='Local',  # Customer managed policies only
             OnlyAttached=True,  # Ignore non-attached policies
             MaxItems=1000,
@@ -43,10 +48,11 @@ class Command(BaseCommand):
                 bucket_name = _bucket_name(policy_name)
 
                 if not S3Bucket.objects.filter(name=bucket_name).exists():
-                    S3Bucket.objects.create(
-                        name=bucket_name,
-                        is_data_warehouse=True,
-                    )
+                    if not DRYRUN:
+                        S3Bucket.objects.create(
+                            name=bucket_name,
+                            is_data_warehouse=True,
+                        )
                     logger.info(f'Created S3 bucket "{bucket_name}" record for IAM policy "{policy_name}"')
                 else:
                     logger.debug(f'S3 bucket "{bucket_name}" record for IAM policy "{policy_name}" already exists')
--- a/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
+++ b/control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
@@ -47,7 +47,8 @@ class Command(BaseCommand):
             if _is_eligible(policy_name):
                 bucket_name = _bucket_name(policy_name)
 
-                if not S3Bucket.objects.filter(name=bucket_name).exists():
+                s3bucket = S3Bucket.objects.filter(name=bucket_name).first()
+                if not s3bucket:
                     if not DRYRUN:
                         S3Bucket.objects.create(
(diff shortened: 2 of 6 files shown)


xoen:
FYI: @andyhd let me add some more details about the second migration and the re-use of the other migration in the PR description - it shouldn't take long but it would help you to understand better what I have in mind to migrate this stuff.

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
```diff
@@ -0,0 +1,108 @@
...
+                        f'user "{user.username}" ("{user.pk}"): {e}'
+                    )
+                    continue
+                else:
+                    raise e
+
+            if policies:
+                for policy in policies["AttachedPolicies"]:
+                    policy_name = policy["PolicyName"]
+                    if _is_eligible(policy_name):
```
xoen: I'm wondering if here (and in the other migration) I should avoid this massive indented if  body by having:

```python
if not _is_eligible(policy_name):
    continue

# ... if body is not not indented.
bucket_name = _bucket_name(policy_name)
# ...
```

I think it may make things easier to follow.

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
```diff
@@ -0,0 +1,108 @@
...
+                        bucket_name = _bucket_name(policy_name)
+                        s3bucket = S3Bucket.objects.filter(name=bucket_name)
+                        if not s3bucket.exists():
+                            logger.warning(
+                                f'S3 bucket "{bucket_name}" not found: '
+                                f'corresponding to IAM policy "{policy_name}"'
+                            )
+                            continue
+
+                        s3bucket = s3bucket.first()
```
andyhd: Reassigning to the same variable could be confusing here, because you're going from a collection to a single value. In the previous file, you had `s3bucket = S3Bucket.objects.filter(...).first()` - why the change?

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
```diff
@@ -0,0 +1,108 @@
...
+                                        access_level=READWRITE,
+                                    )
+                                logger.info(
+                                    f'UserS3Bucket created: '
+                                    f'({user.username}, {bucket_name})'
+                                )
+                            except Exception as e:
+                                logger.critical(
+                                    f'Failed to create UserS3Bucket: '
+                                    f'for ({user.username}, {bucket_name}): {e}'
```
andyhd: `logger.critical` but it doesn't break out of the loop or stop the migration?

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
+import logging
+import re
+import os
```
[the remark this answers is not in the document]
xoen: Italian alphabet has 'r's before 'o's...just kidding, it slipped :P

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+
+import boto3
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from control_panel_api.models import S3Bucket
+
+DRYRUN = os.environ.get('DRYRUN', 'false').lower() == 'true'
```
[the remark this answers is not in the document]
xoen: I'll have a play to see how command line options work, etc...

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+READWRITE = 'readwrite'
+
+
+logger = logging.getLogger(__name__)
+
+
+def _is_eligible(policy_name):
```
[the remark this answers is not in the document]
xoen: Alright.

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+    return (policy_name.startswith(f'{settings.ENV}-') and
+            not policy_name.startswith(f'{settings.ENV}-app-') and
+            policy_name.endswith(READWRITE))
+
+
+def _bucket_name(policy_name):
+    return re.sub(f'-{READWRITE}$', '', policy_name)
```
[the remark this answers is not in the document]
xoen: For some reason I find the use of `sub` more explicit than trimming by a certain number of chars, but I'll play with it.

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+        results = iam.list_policies(
+            Scope='Local',  # Customer managed policies only
+            OnlyAttached=True,  # Ignore non-attached policies
+            MaxItems=1000,
+        )
+
+        for policy in results['Policies']:
+            policy_name = policy['PolicyName']
+
+            if _is_eligible(policy_name):
```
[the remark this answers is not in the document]
xoen: That's what I thought as well in my other comment as it's clear this is only happening when eligible is `true` and skip the other policies entirely.

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+                    if not DRYRUN:
+                        S3Bucket.objects.create(
+                            name=bucket_name,
+                            is_data_warehouse=True,
+                        )
+                    logger.info(
+                        f'Created S3 bucket "{bucket_name}" record '
+                        f'for IAM policy "{policy_name}"'
+                    )
+                elif not s3bucket.is_data_warehouse:
```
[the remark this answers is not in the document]
xoen: I know what you mean, and I thought of inverting them but then I realised I like to have the "happy path" branch, with the logic which creates the S3 bucket above as this is the main objective of the migration.

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def _is_eligible(policy_name):
+    return (policy_name.startswith(f'{settings.ENV}-') and
+            not policy_name.startswith(f'{settings.ENV}-app-') and
+            policy_name.endswith(READWRITE))
+
+
+def _bucket_name(policy_name):
+    return re.sub(f'-{READWRITE}$', '', policy_name)
```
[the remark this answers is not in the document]
xoen: As far as I can tell with my limited knowledge of Django commands, it's a migration per file. I could obviously have a single big migration but I prefer not (as mentioned in PR description). I could maybe have a `migrate_lambda_data_utils` module.

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+    help = ("Add UserS3Bucket records to reflect access to team S3 buckets "
+            "granted by AWS Lambda functions.")
+
+    def handle(self, *args, **options):
+        iam = boto3.client('iam')
+
+        users = User.objects.all()
+        for user in users:
```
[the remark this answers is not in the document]
xoen: I knew you were going to comment on this, it's mainly as I was testing this on a limited number of users (basically my and Robin's user) by replacing this variable, but I'll change.

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
xoen: Uh, I thought I changed this like in the other migration. I'll change here too 👍

control_panel_api/management/commands/migrate_lambdas_data_2_users3buckets.py
xoen: I see what you mean but I still think it would be annoying and draconian to stop the migration halfway through in production just because we can't create a single record for example.

The idea was to output this as critical to highlight the problem and investigate/fix.

This is probably "never" going to happen but I wanted to handle anyway to be defensive. Although I just realised that I swallow the exception here, and that's bad as I wouldn't know what went wrong in the unlikely event there is an exception. I will change the message to display the error in the message.

xoen: Oh actually, I didn't swallow the exception error, it was just the weird GH diff formatting 🎉

xoen:
@andyhd I updated the migrations to address most of your comments (but not the `--dry-run` option *yet*) in [this commit](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109/commits/944ffb93444440b4818bb83546455557b82f41df)

And also added that command to dump the state of users' IAM roles and attached policies I mentioned earlier: [`migrate_lambda_data_0_dump_user_policies`](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109/commits/ac2321f51be324ad1572bd55488c97f4808589f8).

xoen:
@andyhd and this is the [change to add the `--dry-run` flag](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109/commits/bf51a0e1e1d21d7444da03521674b1fc771285b4).

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+    help = "Add records for S3 buckets created by AWS Lambda functions."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dry-run',
+            dest='dry_run',
+            help="Run command without side effects",
+            default=False,
```
andyhd: You don't need the `dest` and `default` arguments here, because these are the default values.

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+            action='store_true',
+        )
+
+    def handle(self, *args, **options):
+        iam = boto3.client('iam')
+        results = iam.list_policies(
+            Scope='Local',  # Customer managed policies only
+            OnlyAttached=True,  # Ignore non-attached policies
```
andyhd: This comment just repeats what the code says. Why ignore non-attached policies?

control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class Command(BaseCommand):
+    """
+    NOTE: Needs permission to perform `iam:ListAttachedRolePolicies` action
+    """
+
+
+    help = ("For each of the users in the DB, it collects its attached IAM "
+            "policies and prints this data to stdout")
```
andyhd: Just an idea - it might be good to have `help` be equal to `__doc__` and put the help text in the class doc comment. Eg:
```python
class Command(BaseCommand):
    """
    For each of the users in the DB, ...
    NOTE: Needs permission ...
    """
    help = __doc__

File: control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py [hunk 15]
```diff
@@ -0,0 +1,58 @@
...
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+from django.core.management.base import BaseCommand
+
+from control_panel_api.models import User
+
+
+READWRITE = 'readwrite'
```
andyhd: Don't need this in this file any more

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    NOTE: Needs permission to perform `iam:ListPolicies` action
+    """
+
+
+    help = "Add records for S3 buckets created by AWS Lambda functions."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dry-run',
```
andyhd: Just a suggestion - you could extract `add_arguments` into a mixin class

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
xoen: > Explicit is better than implicit.

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
xoen: Because we may have IAM policies for test S3 buckets not attached to anything. We don't need to create records for them in the CP and we can safely ignore those.

control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
xoen: 👍

control_panel_api/management/commands/migrate_lambda_data_0_dump_user_policies.py
xoen: That was an artefact of the Copy&Paste. Well spotted.

control_panel_api/management/commands/migrate_lambdas_data_1_buckets.py
xoen: Moved to `DryRunnable` 👍

xoen:
@andyhd last round of amendments as per suggestions: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109/commits/b782a3e2a0714fd20c729bbc9aeb6af127b713e2

xoen:
Thanks for the review Andy :) - merging.

PR #111, opened 2018-03-27
Fixed --dry-run in migrate_s3_access_roles_to_inline_policies

xoen:
`options` is an argument of `handle()` but not of the helper
functions. DOH!


xoen:
(Discussed this with Andy offline. Merging this)

PR #113, opened 2018-04-03
Normal users can list users

xoen:
### What
Updated the User permissions/filtering to allow normal users to list all
users. This is a requirement coming from the Clive/UI.

The reason for this is to allow normal users to grant access to other users
to S3 bucket they are admin of.

The changes are in two areas:
- Users are not filtered anymore, e.g. all users are returned by `UserFilter`
- `UserPermissions` now allows `list` actions for everyone. Note that normal
  users will still only be able to retrieve/update their own user record only.

**NOTE**: I inverted the logic in `UserPermissions` to be a whitelist of
actions rather. It's easier to see what a normal user can do and also
generaly whitelists are safer.

**NOTE (on tests)**: Unauthorised users used to get a `404 NOT FOUND` because records were filtered out before reaching the permissions layer. Now that all users can list all users, they'll get a `403 UNAUTHORIZED`.

### Ticket
https://trello.com/c/ivwsh6FI/715-3-normal-users-can-list-users

### DIFF
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -56,17 +56,11 @@ class UserFilter(DjangoFilterBackend):
     """
     Filter to get visible users.
 
-    Currently superusers see everything, normal users see only themselves,
-    unauthenticated users see nothing.
+    Everyone can see all users.
     """
 
     def filter_queryset(self, request, queryset, view):
-        queryset = super().filter_queryset(request, queryset, view)
-
-        if is_superuser(request.user):
-            return queryset
-
-        return queryset.filter(pk=request.user.pk)
+        return super().filter_queryset(request, queryset, view)
 
 
 class UserS3BucketFilter(DjangoFilterBackend):
--- a/control_panel_api/filters.py
+++ b/control_panel_api/filters.py
@@ -52,17 +52,6 @@ class S3BucketFilter(DjangoFilterBackend):
         return queryset.accessible_by(request.user)
 
 
-class UserFilter(DjangoFilterBackend):
-    """
-    Filter to get visible users.
-
-    Everyone can see all users.
-    """
-
-    def filter_queryset(self, request, queryset, view):
-        return super().filter_queryset(request, queryset, view)
-
-
 class UserS3BucketFilter(DjangoFilterBackend):
     """
     Filter to get visible UserS3Bucket records.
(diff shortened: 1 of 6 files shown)


xoen:
@axemonkey FYI

control_panel_api/filters.py
```diff
@@ -56,17 +56,11 @@ class UserFilter(DjangoFilterBackend):
...
     """
 
     def filter_queryset(self, request, queryset, view):
-        queryset = super().filter_queryset(request, queryset, view)
-
-        if is_superuser(request.user):
-            return queryset
-
-        return queryset.filter(pk=request.user.pk)
+        return super().filter_queryset(request, queryset, view)
```
andyhd: This makes this class a no-op. Probably it should be removed and the reference to it from `views.py` should be removed.

control_panel_api/tests/permissions/test_user_permissions.py
```diff
@@ -36,11 +35,11 @@ def test_list_as_superuser_responds_OK(self):
         response = self.client.get(reverse('user-list'))
         self.assertEqual(HTTP_200_OK, response.status_code)
 
-    def test_list_as_normal_user_responds_403(self):
+    def test_list_as_normal_user_responds_OK(self):
```
andyhd: Just a comment - no need to do anything now - these tests seem to be asking to be data-driven, eg:
```python
def test_list_users(self):
    cases = [
        (self.normal_user, HTTP_200_OK),
        (self.superuser, HTTP_200_OK),
    ]
    for user, expected_response in cases:
        self.client.force_login(user)
        response = self.client.get(reverse('user-list'))
        self.assertEqual(expected_response, response.status_code)
```

xoen:
FYI: Fixed failing authentication test: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/113/commits/413a0cfc44fa728034e63ec6e95309cf71b0fcaa#diff-c9903a591c3d4ace11e12d1f7b071d93R9 (the test case for when a new user is created on the fly by the authentication layer).

control_panel_api/filters.py
xoen: Let me have a look, can't remember what is the default filter class in the views.

xoen: FYI: Got rid of that class and using the parent class directly now https://github.com/ministryofjustice/analytics-platform-control-panel/pull/113/commits/c1a1230c17a00c1d41b7ac524b849f539002969d

PR #114, opened 2018-04-05
Update helm repo before each installation

xoen:
### What
Trello: https://trello.com/c/UcRhiW3Q
As currently the helm repository is updated on Docker image build, newly
released helm chart versions may not be available for the CP-API until
the next release.

I'm now running `helm repo update` before upgrading/install any release.

### Kind of related ticket
We originally thought a [user issue](https://trello.com/c/rsFE4N0z/716-iron-out-issues-from-rbac-investigate-ci-cp) was caused by a non-up-to-date CP installing an older version of an helm chart.

This was not the case but unless we don't do an `helm repo update` before each installation it's still potentially a problem.

### DIFF
--- a/control_panel_api/helm.py
+++ b/control_panel_api/helm.py
@@ -23,6 +23,8 @@ class Helm(object):
             )
 
     def upgrade_release(self, release, chart, *args):
+        self._helm_shell_command('repo update')
+
         default_flags = ['--install', '--wait']
         flags = list(args) + default_flags
         self._helm_command('upgrade', release, chart, *flags)


xoen:
FYI @andyhd

xoen:
When trying to deploy my own RStudio instance locally, `helm upgrade` worked but `helm repo update` returned non-zero exit code.

I wonder why, we may need to run subprocess with `shell=True`, I'm having a look. (to be honest I have zero experience with python's `subprocess` but I'll find a way)

xoen:
Actually `helm upgrade` did **not** work, `views` swallowed the exception.

xoen:
FYI: Using `Helm._helm_shell_command()` instead of `Helm._helm_command()` (which is using `            subprocess.run()` with `shell=True`) and the `helm repo update` is performed correctly now.

PR #115, opened 2018-04-06
When deleting an S3 bucket, update the permissions on associations

xoen:
### Background
Permissions in DB are kept in two models `UserS3Bucket` and `AppS3Bucket`.
These records contains a reference to the S3 bucket.
When these records are deleted the user/app IAM role needs to be updated
to reflect the fact that these users/apps don't have access to the
corresponding S3 bucket.

### Problem
This happened when these records were deleted directly but not when deleted
inderectly when the deletion of an S3 bucket cascaded and deleted all the
associated records.

As we don't actually delete the S3 bucket in AWS a user/app would still
have access to that "deleted" S3 bucket.

### Solution
The solution is to listen to the `post_delete` signal which is [still triggered](https://stackoverflow.com/questions/8944207/overridden-delete-method-of-child-model-does-not-get-called-when-deleting-parent)
when a record is deleted by the cascading.

**NOTE**: I'm passing `dispatch_uid` when connecting the signal to avoid these being registered multiple times.

**NOTE**: As update the IAM roles when access to S3 bucket is revoked is now
performed by the signal, I deleted the custom `perform_destroy()` in the
views which were added to do that.

### Ticket
https://trello.com/c/f0gPNADq/720-4-delete-data-source-not-updating-permissions

### Issue
https://github.com/ministryofjustice/analytics-platform/issues/26

### DIFF
--- a/control_panel_api/models.py
+++ b/control_panel_api/models.py
@@ -3,6 +3,7 @@ import re
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser
 from django.db import models
+from django.db.models.signals import post_delete
 from django.template.defaultfilters import slugify
 from django_extensions.db.fields import AutoSlugField
 from django_extensions.db.models import TimeStampedModel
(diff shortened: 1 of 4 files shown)


xoen:
Cheers @andyhd

PR #116, opened 2018-04-09
Allow normal users to list apps

xoen:
### What
- Updated `AppPermissions` to allow normal users to `list`
- `AppFilter` was a no-op, removed and use `django_filters.rest_framework.DjangoFilterBackend`
   directly

### Why
The reason to allow normal users to list apps is because this list
will allow them to grant access to apps for buckets they admin.

### Ticket
https://trello.com/c/c8faMLoL/725-2-normal-users-to-be-able-to-list-apps


control_panel_api/views.py
```diff
@@ -112,10 +111,10 @@ class GroupViewSet(viewsets.ModelViewSet):
 class AppViewSet(viewsets.ModelViewSet):
     queryset = App.objects.all()
     serializer_class = AppSerializer
-    filter_backends = (AppFilter,)
+    filter_backends = (DjangoFilterBackend,)
```
andyhd: Should we instead change `settings.DEFAULT_FILTER_BACKENDS` to `DjangoFilterBackend` and explicitly set `filter_backends` to `SuperusersOnlyFilter` on the viewsets that require it?

xoen: I personally think we shouldn't.

The reason why have the default filter to `SuperusersOnlyFilter` is to be very conservative and have only super users be able to see everything by default. The idea is to require some "effort" if we really want to open an endpoint to everyone (as in this case).

andyhd: 👍

PR #117, opened 2018-04-11
Improve README

davidread:
I added to the README things I would have found useful as I worked on getting it running (which I didn't complete)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # analytics-platform-control-panel
-Control panel contains admin functions like creating users and granting access to apps
+The logic/backend of the Control Panel app. Serves an API for creating users,
+apps, tools & managing permissions, which are achieved by calling through to
+Kubernetes API (via Helm) and AWS IAM API.
 
+The frontend is:
+https://github.com/ministryofjustice/analytics-platform-control-panel-frontend
 
 ## Running with Docker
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,10 @@
 # analytics-platform-control-panel
-Control panel contains admin functions like creating users and granting access to apps
+The logic/backend of the Control Panel app. Serves an API for creating users,
+apps, tools & managing permissions, which are achieved by calling through to
```
xoen: If we're adding more details here you may also add a small mention to the fact that permissions are also updated in AWS IAM.

README.md
```diff
@@ -26,15 +30,33 @@ docker-compose -f docker-compose.test.yml up --abort-on-container-exit
...
+You need to tell Helm to use the AP charts:
+```sh
+helm repo add mojanalytics https://ministryofjustice.github.io/analytics-platform-helm-charts/charts/
+helm repo update
+```
+
 ### <a name="env"></a>Environment variables
 
+#### Bare minimum (for local testing)
+export DEBUG=True
```
xoen: You may also want to wrap these commands (and the others added) in a <tt>```sh</tt> block as well.

Not a big deal really so don't worry too much but it may make skimming through this document a bit more pleasant :)

README.md
```diff
@@ -55,6 +77,7 @@ pip install -r requirements.txt
 | `DB_PASSWORD` | Postgres password | |
 | `DB_HOST` | Hostname of postgres server | `127.0.0.1` |
 | `DB_PORT` | Postgres port | `5432` |
+| `ENABLED` | Feature flags (see base.py) | |
```
xoen: Good one and I like that you point to the file where these are 👍 - I would probably change the name in the README to `ENABLE_*` so it's even clearer that you're referring to the environment variables starting with `ENABLE_`.

README.md
```diff
@@ -73,6 +96,7 @@ python manage.py migrate
 ```sh
 python manage.py createsuperuser
 ```
+NB `Username` needs to be your GitHub username
```
xoen: Good one :)

PR #118, opened 2018-04-13
Allow users' IAM roles to read inline policies

xoen:
### What / Why
Users in RStudio using new version of S3Browser gets a permission error caused by their role not having the permission to read their `s3-access` inline policy (this inline policy is used to determine who can access which S3 buckets)

[See ticket](https://trello.com/c/TOj5GLE7/735-4-permission-fix-for-users-roles-not-able-to-read-their-inline-policy-in-rstudio) for more details.

### User creation changes
As before, when a user is created its IAM role is created.
In addition to this, we now attach the IAM policy which grants them access to the inline policies.

This IAM policy is created by a django command (see below).

### `grant_aws_roles_permission_to_read_inline_policies` - Django command
This Django command (`grant_aws_roles_permission_to_read_inline_policies`)
will:
- create an IAM policy (`${ENV}-read-user-roles-inline-policies`) to grant permission to read inline policies from
users' IAM roles
- then attach this IAM policy to each or the users' IAM roles

**NOTE**: There are no IAM policy variables to access the role ARN/name at runtime so this policy allow reading all users' roles inline policies
which is a bit loose but shouldn't be a security problem.
The alternative would be to have a policy per user but that would be
overkill.

### Ticket

https://trello.com/c/TOj5GLE7/735-4-permission-fix-for-users-roles-not-able-to-read-their-inline-policy-in-rstudio

### DIFF
--- a/control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
+++ b/control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
@@ -7,10 +7,10 @@ from django.conf import settings
 
 from control_panel_api.management.commands.dry_runnable import DryRunnable
 from control_panel_api.models import User
+from control_panel_api.services import READ_INLINE_POLICIES_POLICY_NAME as POLICY_NAME
 
 
 IAM = boto3.client('iam')
-POLICY_NAME = f'{settings.ENV}-read-user-roles-inline-policies'
 POLICY_ARN = f"{settings.IAM_ARN_BASE}:policy/{POLICY_NAME}"
 POLICY_DOCUMENT = {
     "Version": "2012-10-17",
--- a/control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
+++ b/control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
@@ -1,0 +1,81 @@
+import json
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+from django.conf import settings
+
+from control_panel_api.management.commands.dry_runnable import DryRunnable
+from control_panel_api.models import User
+
+
+IAM = boto3.client('iam')
+POLICY_NAME = f'{settings.ENV}-read-user-roles-inline-policies'
+POLICY_ARN = f"{settings.IAM_ARN_BASE}:policy/{POLICY_NAME}"
+POLICY_DOCUMENT = {
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "CanReadUserRolesInlinePolicies",
+            "Effect": "Allow",
+            "Action": [
+                "iam:GetRolePolicy"
+            ],
+            "Resource": [
+                f"{settings.IAM_ARN_BASE}:role/{settings.ENV}_user_*",
+            ],
+        },
+    ],
+}
+
+
+logger = logging.getLogger(__name__)
+
+
+def create_policy():
+    try:
+        IAM.create_policy(
+            PolicyName=POLICY_NAME,
+            PolicyDocument=json.dumps(POLICY_DOCUMENT),
+        )
+        logger.info(f'IAM policy "{POLICY_NAME}" created')
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'EntityAlreadyExists':
+            logger.warning(f'IAM policy "{POLICY_NAME}" already exists')
+        else:
+            raise e
+
+
+def attach_policy(role_name):
+    try:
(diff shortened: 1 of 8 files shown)


control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
```diff
@@ -0,0 +1,81 @@
...
+    ],
+}
+
+
+logger = logging.getLogger(__name__)
+
+
+def create_policy():
+    try:
+        IAM.create_policy(
```
andyhd: Why doesn't this code use `control_panel_api.aws.AWSClient`?

control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
```diff
@@ -0,0 +1,81 @@
...
+    Creates the "${ENV}-read-users-inline-policies" IAM role and attaches
+    it to all users' IAM roles
+
+    NOTE: Needs permissions to `iam:CreatePolicy` and `iam:AttachRolePolicy`
+    """
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        logger.info(f'Creating IAM policy "{POLICY_NAME}"...')
```
andyhd: This will be a bit verbose, because you're logging "Creating IAM policy ..." and then "IAM policy created/already exists".

control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
```diff
@@ -0,0 +1,81 @@
...
+logger = logging.getLogger(__name__)
+
+
+def create_policy():
+    try:
+        IAM.create_policy(
+            PolicyName=POLICY_NAME,
+            PolicyDocument=json.dumps(POLICY_DOCUMENT),
+        )
+        logger.info(f'IAM policy "{POLICY_NAME}" created')
```
andyhd: Consider removing this log line, it will make output a bit verbose. The usual guideline is to log only if something is unexpected. So since you're already logging "Creating policy", there's no real need to log "created" (or even "already exists"), just unexpected things like errors.

control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
xoen: I thought about that but then I preferred to keep these command de-coupled from the rest of the code so we can't accidentally break them.

control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
```diff
@@ -0,0 +1,81 @@
...
+def attach_policy(role_name):
+    try:
+        IAM.attach_role_policy(
+            RoleName=role_name,
+            PolicyArn=POLICY_ARN,
+        )
+    except ClientError as e:
+        if e.response['Error']['Code'] in ('NoSuchEntity', 'ValidationError'):
+            logger.warning(f'IAM role "{role_name}" not found or invalid')
+        else:
```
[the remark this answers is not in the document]
xoen: Essentially some users may not have an IAM role (e.g. we have a Jenkins user) or be test users created manually from the console with bad usernames, in these cases `attach_role_policy` will raise `NoSuchEntity` or `ValidationError`.

This logs it and carry on.

control_panel_api/management/commands/grant_aws_roles_permission_to_read_inline_policies.py
xoen: Removed 👍

control_panel_api/tests/test_models.py
```diff
@@ -51,9 +52,17 @@ def test_aws_create_role_calls_service(self):
...
+        aws_client.create_role.assert_called_with(
             RoleName=expected_role_name,
             AssumeRolePolicyDocument=json.dumps(USER_IAM_ROLE_ASSUME_POLICY))
+        aws_client.attach_role_policy.assert_called_with(
+            RoleName=expected_role_name,
+            PolicyArn=(
+                f'{settings.IAM_ARN_BASE}:policy/'
+                f'{settings.ENV}-read-user-roles-inline-policies'
+            ),
+        )
```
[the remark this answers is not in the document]
xoen: I see what you mean. In this case is really testing the contract (`User.aws_create_role()`), the fact that the User model has this method and it's performing the operation it's supposed to do.

Having said that, there is always room for improvement. As discussed, the opposite extreme (have all the tests in `test_views`) could mean testing a lot in a single place.

Worth playing with some ideas tho :)

PR #120, opened 2018-05-01
Elasticsearch access logs endpoint

jmoz:
## What

Use ES to fetch the kibana logs of aws s3bucket access.

Hit `/s3buckets/{bucket_id}/access_logs/?day_range=30` and filter by a day_range. No filter returns all.

We interpret a GET on the bucket or object as a hit.  We eliminate hits from AWS bot useragent.

Use ES aggregations for bucket user.  Then parse for the username, group on username, and sum counts (aws role id gets recreated and is not unique).

The test uses a sample response and is separated into a fixture file to eliminate bloat on the __init__.py we use for other fixtures.  AWS fixtures could be split out similarly.

The test is checking that names have been grouped correctly and sums calculated and user type.

## How to review

1. Run tests.
2. Request a s3bucket with corresponding kibana logs e.g. /s3buckets/3/access_logs/?day_range=30

https://trello.com/c/SzRk2fJK/751-expose-patterns-of-data-access-to-data-access-group-admins


control_panel_api/elasticsearch.py
```diff
@@ -0,0 +1,26 @@
+from elasticsearch import Elasticsearch
+from elasticsearch_dsl import Q, Search
+from elasticsearch_dsl.query import Range
+
+
+class ElasticSearch(object):
```
andyhd: This module now contains `Elasticsearch` and `ElasticSearch` classes. It could potentially result in difficult to catch bugs if someone makes a quite likely typo. Maybe `import Elasticsearch as _Elasticsearch` or import the module and use `self.client = elasticsearch.Elasticsearch()` or change the name of this class.

control_panel_api/elasticsearch.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, connection):
+        self.client = Elasticsearch(connection)
+
+    def bucket_hits_aggregation(self, index, bucket_name, day_range=None):
```
andyhd: This method could be a function - just move the `client = Elasticsearch(connection)` in here and you don't use `self` for anything

control_panel_api/elasticsearch.py
```diff
@@ -0,0 +1,26 @@
...
+
+        s = s.query(q1 & q2 & ~q3)
+
+        if day_range:
+            s = s.filter(Range(time_received={"gte": f"now-{day_range}d/d"}))
+
+        s.aggs.bucket('bucket_hits', 'terms', field='requester_id.keyword',
+                      size=100)
+
+        return s.execute()
```
andyhd: Since you don't use the whole result object (just `aggregations.bucket_hits`), you could return just what is used and then the calling code doesn't need to know about the surrounding data structure

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
...
+        match = re.search('alpha_(app|user)_([\w-]+)/', key)
+
+        if match:
+            return match.group(1), match.group(2)
+
+        return 'unknown', key
+
+
+class S3BucketAccessLogsQueryParamsSerializer(serializers.Serializer):
+    day_range = serializers.IntegerField(required=False)
```
andyhd: `day_range` might be better named `num_days` or similar, as it only represents one limit of the range.

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
...
+        results = [
+            {'accessed_by': k, 'count': v, 'type': accessed_by_type_map[k]}
+            for k, v in accessed_by_sum_map.items()
+        ]
+
+        sorted_by_sum = sorted(results, key=itemgetter('count'), reverse=True)
+
+        return sorted_by_sum
+
+    def _extract_fields(self, key):
```
andyhd: This function name is a bit generic and not intent-revealing. Maybe something like `get_accessed_by`?

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
...
+
+        for result in aggregations.bucket_hits:
+            role_type, accessed_by = self._extract_fields(result.key)
+
+            accessed_by_type_map[accessed_by] = role_type
+
+            if accessed_by not in accessed_by_sum_map:
+                accessed_by_sum_map[accessed_by] = result.doc_count
+            else:
+                accessed_by_sum_map[accessed_by] += result.doc_count
```
andyhd: If you use `.setdefault()` or `defaultdict(int)` you can avoid this test, eg:
```python
accessed_by_sum_map = collections.defaultdict(int)
for result in ...:
    ...
    accessed_by_sum_map[accessed_by] += result.doc_count
```

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
             'nickname',
             'name',
         )
+
+
+class ESBucketHitsSerializer(serializers.BaseSerializer):
+
+    def to_representation(self, aggregations):
+        accessed_by_sum_map = {}
+        accessed_by_type_map = {}
```
andyhd: These names are a bit obtuse. Maybe `access_count` and `accessor_role`?

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
...
+                accessed_by_sum_map[accessed_by] += result.doc_count
+
+        results = [
+            {'accessed_by': k, 'count': v, 'type': accessed_by_type_map[k]}
+            for k, v in accessed_by_sum_map.items()
+        ]
+
+        sorted_by_sum = sorted(results, key=itemgetter('count'), reverse=True)
+
+        return sorted_by_sum
```
andyhd: Could just `return sorted(results, ...)` - it's fairly self-explanatory.

jmoz:
Thanks @andyhd , good review 👍

control_panel_api/views.py
```diff
@@ -227,6 +233,25 @@ def perform_create(self, serializer):
 
         instance.create_users3bucket(user=self.request.user)
 
+    @detail_route()
+    def access_logs(self, request, pk=None):
```
xoen: Isn't this a `list_route`? As far as I can tell `pk` is not used.
I've only just started to have a look at your PR, so apologies if I'm missing something.

control_panel_api/views.py
```diff
@@ -227,6 +233,25 @@ def perform_create(self, serializer):
...
+        es = ElasticSearch(settings.ELASTICSEARCH_CONN)
+
+        try:
+            result = es.bucket_hits_aggregation(
+                settings.ELASTICSEARCH_INDEX_S3LOGS,
+                self.get_object().name,
+                query_params_serializer.validated_data.get('day_range'),
+            )
+        except TransportError as e:
+            raise ESException(e) from e
```
xoen: I don't know in which cases `TransportError` is raised and if it contains any useful details but it may be a good idea to log the original error message before swallowing/wrapping it into `ESException`.

Depending on how much details there are in the original error, having them in the logs may help in the future.

control_panel_api/tests/test_views.py
```diff
@@ -24,6 +24,7 @@
     UserApp,
     UserS3Bucket,
 )
+from control_panel_api.tests.fixtures.es import BUCKET_HITS_AGGREGATION
```
xoen: Nice one 👍

control_panel_api/views.py
jmoz: `detail_route` applies to the s3bucket so you get the {id} part and object, this is just a sub route

control_panel_api/settings/base.py
```diff
@@ -167,6 +167,9 @@ def is_enabled(value):
 RSTUDIO_AUTH_CLIENT_ID = os.environ.get('RSTUDIO_AUTH_CLIENT_ID')
 RSTUDIO_AUTH_CLIENT_SECRET = os.environ.get('RSTUDIO_AUTH_CLIENT_SECRET')
 
+ELASTICSEARCH_CONN = os.environ.get('ELASTICSEARCH_CONN')
```
xoen: Let's not forget to add this config to the helm chart/config or it will blow up 💥 😄

control_panel_api/views.py
jmoz: The original is logged by the es library. The error message will get passed to this exception and stacktrace is maintaned due to the use of `from e`. This error is just a nice wrapper and enables the framework to handle error output.

control_panel_api/views.py
```diff
@@ -227,6 +233,25 @@ def perform_create(self, serializer):
...
+    def access_logs(self, request, pk=None):
+        query_params_serializer = S3BucketAccessLogsQueryParamsSerializer(
+            data=request.query_params)
+        query_params_serializer.is_valid(raise_exception=True)
+
+        es = ElasticSearch(settings.ELASTICSEARCH_CONN)
+
+        try:
+            result = es.bucket_hits_aggregation(
+                settings.ELASTICSEARCH_INDEX_S3LOGS,
```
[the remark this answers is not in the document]
jmoz: Yep i'll move index inside. 👍

control_panel_api/settings/base.py
jmoz: 😀

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
...
+
+    def _extract_fields(self, key):
+        match = re.search('alpha_(app|user)_([\w-]+)/', key)
```
xoen: Luckily [GH](https://github.com/isaacs/github/issues/906) doesn't allow `_` in username or this RE would break for users with an underscore 🤓

xoen: Shouldn't `alpha_` come from the `settings.ENV` otherwise it will only match keys in `alpha`.

jmoz: Well spotted, I meant to do that but forgot!

control_panel_api/serializers.py
```diff
@@ -266,3 +269,41 @@ class Meta:
...
         )
+
+
+class ESBucketHitsSerializer(serializers.BaseSerializer):
+
+    def to_representation(self, aggregations):
+        accessed_by_sum_map = {}
+        accessed_by_type_map = {}
+
+        for result in aggregations.bucket_hits:
```
xoen: I may miss something but looking at the fixture file (`es.py`) shouldn't this be `aggregations.bucket_hits.buckets`?

I mean, test seems to work so maybe the `for ... in` in Python is doing something clever in this case and loop through the `buckets` array implicitly?

jmoz: The elasticsearch library does some overloading of magic methods so it gives you this sometimes unintuitive interface.

control_panel_api/elasticsearch.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ElasticSearch(object):
+    def __init__(self, connection):
+        self.client = Elasticsearch(connection)
+
+    def bucket_hits_aggregation(self, index, bucket_name, day_range=None):
+        s = Search(using=self.client, index=index)
+
+        q1 = Q('term', **{'bucket.keyword': bucket_name})
+        q2 = Q('terms',
```
xoen: Are these the only operations logged at the moment? Are we interested in other operations? (You probably already asked the same question to Robin)

control_panel_api/serializers.py
xoen: Or maybe something like `get_role_name_and_type()`...not sure about the name but I would definitely keep the `_` prefix so it's clear it's only used internally and the future us know they can go nuts on refactoring/renaming if necessary.

control_panel_api/elasticsearch.py
xoen: You could move the `index` param last and default to `settings.ELASTICSEARCH_INDEX_S3LOGS` so that the client code doesn't need to pass it unless it wants to change the default behaviour.

control_panel_api/elasticsearch.py
```diff
@@ -0,0 +1,26 @@
+from elasticsearch import Elasticsearch
+from elasticsearch_dsl import Q, Search
+from elasticsearch_dsl.query import Range
+
+
+class ElasticSearch(object):
+    def __init__(self, connection):
```
xoen: As said in the `views.py` comment, you could make `connection` optional and default to `settings.ELASTICSEARCH_CONN` so client code can be simplified and not pass it unless it really wants to change default.

control_panel_api/serializers.py
xoen: Magic 🎩

control_panel_api/serializers.py
xoen: [Ruby's hashes](https://ruby-doc.org/core-2.5.0/Hash.html) has something similar which is handy in these cases and I was wondering if you could do the same here.

Nice tip 👍

xoen: (Troll mode on: In Ruby you would just do `accessed_by_sum_map = Hash(0)` because it's a beautiful language :trollface:  but it's the same thing)

control_panel_api/elasticsearch.py
xoen: 👍

control_panel_api/views.py
xoen: 👍

xoen:
I had another look and it seems OK to me. Good stuff.

Not a big deal (but I can imagine Andy will point it out anyway) but some of the [`import` order](https://www.python.org/dev/peps/pep-0008/#imports) is a bit off, e.g. `django.*` imports are with import of standand libraries - feel free to fix if you want.

I'd also hold fire on merging this until the configs are in place.

PR #146, opened 2018-05-11
Replace invalid chars in docker tag names

andyhd:
## What

* Replace `/` with `-` in branch name when using it to tag the Docker image in the Circle CI script - because quay.io doesn't allow `/` in tag names (and maybe neither does Docker generally?)


xoen:
👍

PR #180, opened 2018-05-25
Fix app customers infinite login loop

andyhd:
## Why

* Customers created via the Control Panel API cannot login to the apps they are invited to - they receive a login link via email, but it does not log them in successfully.
* The reason is that the customers are (sometimes) added to an Auth0 Authorization Group with an incorrect name.
* The Control Panel API uses the app **name** (eg: `Crime_in_prisons`) to get _or create_ a Group.
* The Concourse deployment pipeline creates Groups with the app **slug** (eg: `crime-in-prisons`).
* This problem does not occur if the app name is identical to the slug.
* Because of this, customers are members of a valid Group, but which may not have permission to access the app, so they receive a login link email, but cannot use it to login.

# What

* Changed app customers code to get/create Auth0 authorization groups using the app slug, rather than the app name.

### DIFF
--- a/control_panel_api/auth0.py
+++ b/control_panel_api/auth0.py
@@ -45,7 +45,7 @@ class Auth0(object):
         return group.get_members()
 
     def add_group_member(self, group_name, email):
-        group = self.authorization_api.get_or_create(Group(name=group_name))
+        group = self.authorization_api.get(Group(name=group_name))
 
         user = self.authorization_api.get(User(
             email=email,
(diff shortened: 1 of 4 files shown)


control_panel_api/auth0.py
```diff
@@ -45,7 +45,7 @@ def get_group_members(self, app_name):
         return group.get_members()
 
     def add_group_member(self, group_name, email):
-        group = self.authorization_api.get_or_create(Group(name=group_name))
+        group = self.authorization_api.get(Group(name=group_name))
```
xoen: What about when we genuinely need to create the group the first time?

andyhd: The group is created by the app deployment pipeline

andyhd: But I guess as long as we're all using `slug`, this can remain `get_or_create`

PR #181, opened 2018-05-25
Add patch for kubernetes client bug

andyhd:
# What

* Added a temporary patch for [a bug in the current version of the Python Kubernetes client](https://github.com/kubernetes-client/python/issues/525) which fails to correctly decode the base64 encoded OIDC token. This causes the API to raise an exception on access.
* The Kubernetes client has also changed how configuration is available, so I have changed the API code around the initialization of the client to match the new interface

### DIFF
--- a/control_panel_api/k8s_patch.py
+++ b/control_panel_api/k8s_patch.py
@@ -1,0 +1,45 @@
+import base64
+import datetime
+import json
+from datetime import timezone
+
+import kubernetes
+from kubernetes.config.kube_config import _is_expired
+
+
+def load_token(self):
+    if 'auth-provider' not in self._user:
+        return
+
+    provider = self._user['auth-provider']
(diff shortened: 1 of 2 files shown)


control_panel_api/k8s_patch.py
```diff
@@ -0,0 +1,45 @@
...
+            or provider['name'] != 'oidc'):
+        return
+
+    parts = provider['config']['id-token'].split('.')
+
+    if len(parts) != 3:  # Not a valid JWT
+        return None
+
+    jwt_attributes = json.loads(
+        base64.b64decode(parts[1] + '==').decode('utf-8')
```
andyhd: This is the key part of this patch - the official client code is missing the ` + '=='` and so base64 complains about `Invalid padding`

PR #247, opened 2018-07-05
Bump versions

andyhd:
## What

Dependabot suggested version increments. Applied all at once and tested.

## How to review

1. `docker-compose -f docker-compose.test.yaml up`


andyhd:
Stderr stuff in by mistake

PR #248, opened 2018-07-05
Bump versions

andyhd:
## What

Really bump versions - last PR was strangely closed when I rebased against master

## How to review

1. `docker-compose build`
2. `docker-compose -f docker-compose.test.yaml up`


requirements.txt
```diff
@@ -30,23 +30,23 @@ kubernetes==6.0.0
 MarkupSafe==1.0
 model-mommy==1.6.0
 openapi-codec==1.3.2
-psycopg2==2.7.4
-py==1.5.3
+psycopg2==2.7.5
```
r4vi: we should probably change this to psycopg2-binary

requirements.txt
```diff
@@ -16,12 +16,12 @@ django-rest-swagger==2.2.0
 djangorestframework==3.8.2
 docutils==0.14
 ecdsa==0.13
-elasticsearch==6.2.0
+elasticsearch==6.3.0
 elasticsearch-dsl==6.1.0
```
r4vi: might as well bump this to 6.2.1?

requirements.txt
andyhd: Or change this to `psycopg2==2.7.5 --no-binary psycopg2`? Which do you think? I'm not sure about psycopg2-binary...

r4vi: for now it's fine but when it hits 2.8 we'll have to decide: http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/

PR #265, opened 2018-07-10
Prevent KeyError 'app' when updating apps3bucket

andyhd:
## What

Fix guards for changing `app` and `s3bucket` fields of apps3bucket object.
Update requests contain only fields to update and the apps3bucket id, and may (should) omit `app` and `s3bucket` fields.

See [Sentry](https://sentry.service.dsd.io/mojds/control-panel-api/issues/32536/)

### DIFF
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -30,11 +30,11 @@ class AppS3BucketSerializer(serializers.ModelSerializer):
         fields = ('id', 'url', 'app', 's3bucket', 'access_level')
 
     def update(self, instance, validated_data):
-        if instance.app != validated_data['app']:
+        if 'app' in validated_data and instance.app != validated_data['app']:
             raise serializers.ValidationError(
                 "App is not editable. Create a new record."
             )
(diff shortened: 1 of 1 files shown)


xoen:
What about [`UserS3Bucket`](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/control_panel_api/serializers.py#L59)? I see we're using `get()` there so it shouldn't raise an exception:

```python
if user != validated_data.get('user', user):
  # ...
```

But what if using `PATCH` in these endpoints? Wouldn't `validated_data.get('user', user)` (and same for the `s3bucket` field) return `None`? Therefore this condition would be `False`/prevent the patch right?

andyhd:
I'm actually going to have another go at this - the code is making these fields read-only, so they should be made `read_only` fields in the Serializer, rather than have this code here.

xoen:
@andyhd yeah, it's good that a fresh pair of eyes look at this 👍

andyhd:
Closing for now - will have another attempt

PR #270, opened 2018-07-12
Fix adding platform users as customers

andyhd:
## What

* Refactored customer management as methods on App model - the View was dealing with too low a level of abstraction, reading and writing Auth0 Authorization API resources. Instead, this has been abstracted by a simpler interface on the App model: `App.get_customers`, `App.add_customers` and `App.delete_customers`
* Fixed the logic in `add_customers` to ensure the user resource(s) from Auth0 have `connection: 'email'` - which is what differentiates "customers" from platform users.

### DIFF
--- a/control_panel_api/auth0.py
+++ b/control_panel_api/auth0.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 from django.conf import settings
 
 from moj_analytics.auth0_client import (
@@ -36,34 +38,50 @@ class Auth0(object):
 
         return self.api.management
 
-    def get_group_members(self, app_name):
-        group = self.authorization_api.get(Group(name=app_name))
+    def get_group_members(self, group_name):
+        group = self.authorization_api.get(Group(name=group_name))
 
         if group is None:
             return None
 
         return group.get_members()
 
-    def add_group_member(self, group_name, email):
-        group = self.authorization_api.get_or_create(Group(name=group_name))
+    def add_group_members(self, group_name, emails, user_options={}):
+        group = self.authorization_api.get(Group(name=group_name))
 
-        user = self.authorization_api.get(User(
-            email=email,
-        ))
+        all_users = self.authorization_api.get_all(User)
+        user_lookup = {
+            user['email']: user
+            for user in all_users if 'email' in user}
 
-        if user is None:
-            user = self.management_api.create(User(
-                connection='email',
-                email=email,
-                email_verified=True,
-            ))
+        def has_options(user):
+            return all(
+                item in user.items()
+                for item in user_options.items())
 
-        group.add_users([{'user_id': user['user_id']}])
+        # sorted users for easier testing
+        users_to_add = OrderedDict()
 
-    def delete_group_member(self, group_name, user_id):
+        for email in emails:
+            user = user_lookup.get(email)
+
+            if user and has_options(user):
+                users_to_add[email] = user
+
+            else:
+                users_to_add[email] = self.management_api.create(User(
+                    email=email,
+                    email_verified=True,
+                    **user_options))
+
+        group.add_users(list(users_to_add.values()))
+
+    def delete_group_members(self, group_name, user_ids):
         group = self.authorization_api.get(Group(name=group_name))
 
         if group is None:
             return None
 
-        group.delete_users([{'user_id': user_id}])
+        group.delete_users([
+            {'user_id': user_id}
+            for user_id in user_ids])
--- a/control_panel_api/auth0.py
+++ b/control_panel_api/auth0.py
@@ -74,7 +74,7 @@ class Auth0(object):
                     email_verified=True,
                     **user_options))
 
-        group.add_users(list(users_to_add.values()))
+        group.add_users(users_to_add.values())
 
     def delete_group_members(self, group_name, user_ids):
         group = self.authorization_api.get(Group(name=group_name))
--- a/control_panel_api/tests/test_models.py
+++ b/control_panel_api/tests/test_models.py
@@ -203,7 +203,7 @@ class AppTestCase(TestCase):
 
         self.assertEqual(
             expected_customer_emails,
-            sorted(map(itemgetter('email'), customers)))
+            list(map(itemgetter('email'), customers)))
 
     @patch('control_panel_api.auth0.Auth0Client')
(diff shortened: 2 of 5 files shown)


control_panel_api/auth0.py
```diff
@@ -36,34 +38,50 @@ def management_api(self):
...
+                users_to_add[email] = user
+
+            else:
+                users_to_add[email] = self.management_api.create(User(
+                    email=email,
+                    email_verified=True,
+                    **user_options))
 
-        group.add_users([{'user_id': user['user_id']}])
+        group.add_users(list(users_to_add.values()))
```
r4vi: why does the queryset need to be cast to a list? `.add_users` should be able to handle any iterable

control_panel_api/auth0.py
```diff
@@ -36,34 +38,50 @@ def management_api(self):
...
 
-    def delete_group_member(self, group_name, user_id):
+    def delete_group_members(self, group_name, user_ids):
         group = self.authorization_api.get(Group(name=group_name))
 
         if group is None:
             return None
 
-        group.delete_users([{'user_id': user_id}])
+        group.delete_users([
```
r4vi: why does `delete_users` take `[{'user_id': user_id}]` when your new `delete_group_members` takes a list of user_ids.  If it's not too big a deal can we refactor `delete_users` to take a list of user_ids too? 
The new `delete_group_members` feels cleaner and the old function, and possibility anywhere else we ultimately need a list of user_ids should be changed to get rid of the dictionary

control_panel_api/auth0.py
```diff
@@ -1,3 +1,5 @@
+from collections import OrderedDict
```
xoen: Why `users_to_add` needs to be an ordered dictionary? As far as I can tell order shouldn't matter. What am I missing?

control_panel_api/auth0.py
```diff
@@ -36,34 +38,50 @@ def management_api(self):
...
 
-        if user is None:
-            user = self.management_api.create(User(
-                connection='email',
-                email=email,
-                email_verified=True,
-            ))
+        def has_options(user):
+            return all(
+                item in user.items()
```
xoen: I don't know if it's worth be more explicit and rename `has_options()` to `has_all_options()` and `item` => `option` (or maybe `user_option)`, e.g.

```python
def has_all_options(user):
    return all(
        option in user.items()
        for option in user_options.items())
```

control_panel_api/auth0.py
xoen: Do you have to cast to a `list` and use `values()` here because it's an `OrderedDict` collection?

I quickly tested in the console and it seems like `for key in ...` works with an `OrderedDict` so you should be able to just pass it to `group.add_users()`.

control_panel_api/tests/test_models.py
```diff
@@ -188,6 +190,83 @@ def test_aws_delete_role_calls_service(self):
...
+        ]
+
+        app = App.objects.create(repo_url='https://example.com/repo_name')
+        customers = app.get_customers()
+
+        expected_customer_emails = ['test@example.com']
+
+        self.assertEqual(
+            expected_customer_emails,
+            sorted(map(itemgetter('email'), customers)))
```
xoen: You can probably avoid to sort here in this test right?

control_panel_api/views.py
```diff
@@ -141,14 +140,13 @@ class AppCustomersAPIView(GenericAPIView):
...
 
     def get(self, request, *args, **kwargs):
-        instance = self.get_object()
+        app = self.get_object()
+        customers = app.get_customers()
 
-        members = Auth0().get_group_members(instance.slug)
-
-        if members is None:
+        if not customers:
```
xoen: Isn't this true when there are no customers for an app?
Shouldn't this respond with a `200 OK` and empty list of customers when there are no customers instead?

control_panel_api/views.py
```diff
@@ -157,10 +155,8 @@ def post(self, request, *args, **kwargs):
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
 
-        Auth0().add_group_member(
-            self.get_object().slug,
-            serializer.validated_data['email']
-        )
+        app = self.get_object()
+        app.add_customers([serializer.validated_data['email']])
```
xoen: How does this `validated_data['email']` works when posting a JSON array like in this case? I'm not quite following what's happening here sorry (how the POSTed data is returned, how you convert to a list of emails, why there are `[]` around this expression, etc...)

control_panel_api/auth0.py
andyhd: Yeah, lazy test writer. I'll change it.

control_panel_api/views.py
```diff
@@ -170,7 +166,8 @@ class AppCustomersDetailAPIView(GenericAPIView):
     permission_classes = (IsSuperuser,)
 
     def delete(self, request, *args, **kwargs):
-        Auth0().delete_group_member(self.get_object().slug, kwargs['user_id'])
+        app = self.get_object()
+        app.delete_customers([kwargs['user_id']])
```
xoen: So `delete` is still done one by one 👍

control_panel_api/auth0.py
andyhd: As it says in the comment on L62, the dict is sorted to make testing easier.

control_panel_api/auth0.py
andyhd: Not sure changing the name is necessary. Changing `item` to `option` is incorrect, because the `user.items` are not options.

control_panel_api/auth0.py
andyhd: No, it's cast to a `list` to make testing easier.

control_panel_api/views.py
andyhd: `validated_data['email']` is a `str`, not a JSON array.
I've written `App.add_customers` to take a list of email addresses as the next thing I will do is implement bulk adding customers, but this view still only permits adding one customer at a time.

control_panel_api/views.py
andyhd: Yep, I just changed the API to make it possible to do bulk adds and deletions, but I haven't changed the views.

PR #277, opened 2018-07-16
List more than 25 customers for an app

andyhd:
## What

* Fix for [issue #38](https://github.com/ministryofjustice/analytics-platform/issues/38)
* Refactored `get_all` to get all pages of results, not just the first.
* Removed `get_all_pages` function

## How to review

1. Add 25 customers to an App in the Control Panel
2. Add 1 more customer
3. See that more than 25 customers are listed

### DIFF
--- a/moj_analytics/auth0_client.py
+++ b/moj_analytics/auth0_client.py
@@ -93,7 +93,7 @@ class API(object):
         response = self.request('POST', endpoint, json=resource)
 
         if 'error' in response:
-            raise CreateResourceError(response['message'])
+            raise CreateResourceError(response)
 
         print('Created {resource}({attrs})'.format(
             resource=resource.__class__.__name__.lower(),
@@ -269,7 +269,7 @@ class Group(AuthzResource):
         )
 
         if 'error' in response:
-            raise AddGroupRoleError(response['message'])
+            raise AddGroupRoleError(response)
 
         print('Added role({role}) to group({group})'.format(
             role=role['name'],
@@ -283,7 +283,7 @@ class Group(AuthzResource):
         )
 
         if 'error' in response:
-            raise AddGroupMemberError(response['message'])
+            raise AddGroupMemberError(response)
 
     def delete_users(self, users):
         response = self.api.request(
@@ -293,7 +293,7 @@ class Group(AuthzResource):
         )
 
         if 'error' in response:
-            raise DeleteGroupMemberError(response['message'])
+            raise DeleteGroupMemberError(response)
 
     def get_members(self):
         results = self.api.request(
--- a/moj_analytics/auth0_client.py
+++ b/moj_analytics/auth0_client.py
@@ -6,6 +6,10 @@ class AccessTokenError(Exception):
     pass
 
 
+class APIError(Exception):
+    pass
+
+
 class CreateResourceError(Exception):
     pass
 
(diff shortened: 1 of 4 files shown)


moj_analytics/auth0_client.py
```diff
@@ -104,26 +104,29 @@ def create(self, resource):
...
+        defaults = {
+            'page': 0,
+            'per_page': 100,
+            'include_totals': True
+        }
 
-        if endpoint in resources:
-            resources = resources[endpoint]
+        params = defaults.copy()
+        params.update(kwargs)
```
r4vi:
```python
params = {
  **defaults
  **kwargs
}
``` 
cleaner than .copy & .update?

andyhd: more concise, but less intent-revealing

moj_analytics/auth0_client.py
```diff
@@ -104,26 +104,29 @@ def create(self, resource):
...
-        resources = self.request('GET', endpoint, params=kwargs)
+        defaults = {
+            'page': 0,
+            'per_page': 100,
+            'include_totals': True
+        }
 
-        if endpoint in resources:
-            resources = resources[endpoint]
+        params = defaults.copy()
```
xoen: Kind of a Python question: Why to make a copy here?

Isn't `defaults = {...}` creating a new dictionary at every `get_all()` call?

moj_analytics/auth0_client.py
```diff
@@ -104,26 +104,29 @@ def create(self, resource):
...
-                resource_class, per_page=per_page, page=page, **kwargs)
+            response = self.request('GET', endpoint, params=params)
 
-            for item in items:
-                yield item
+            if endpoint in response:
+                resources.extend(response[endpoint])
 
-            if len(items) < per_page:
+            if len(resources) == response['total']:
```
xoen: I think this is fine but *potentially* we could end up with duplicates and missing elements if something is added to a previous page between the two requests...but I don't know if there is any easy way to prevent this and it doesn't sound like something would happen that frequently anyway.

moj_analytics/auth0_client.py
```diff
@@ -104,26 +104,29 @@ def create(self, resource):
...
 
-        return [resource_class(self, r) for r in resources]
+        resources = []
 
-    def get_all_pages(self, resource_class, per_page=100, **kwargs):
-        page = 0
         while True:
-            items = self.get_all(
-                resource_class, per_page=per_page, page=page, **kwargs)
+            response = self.request('GET', endpoint, params=params)
```
xoen: Hope to not sound annoying, but how would this behave if there are errors?

Presumably if the server doesn't respond `requests` raise an exception (if that happen halfway through that bubbles up which is good).

If status code is not `200 OK`, probably response doesn't have `endpoint` key, this is handled by skipping the page. Wouldn't this cause the loop to never end?

moj_analytics/auth0_client.py
xoen: Another thing I'm wondering is if this `==` to break out of the loop could potentially cause the loop to never ends if:
1. One of the requests fail and we skip a page (see comment above)
2. An item is added between two requests. If this item would be part of a previous page we may never get it but it would (I'm guessing here) be included/counted by the latest `response['total']` by causing this `==` condition to never be true

I know these are a bit of edge cases.

moj_analytics/tests/conftest.py
```diff
@@ -226,3 +226,27 @@ def given_a_role_exists(request, config, mock_resources, state):
...
+    for i in range(0, total, per_page):
+        responses.add_callback(
+            'GET',
+            f'https://{config["domain"]}/api/v2/users',
+            json_response({
+                'start': 0,
+                'limit': per_page,
+                'length': (total - i),
+                'total': total,
+                'users': data[i:i+100]
```
xoen: `100` is `per_page` here right?

moj_analytics/tests/conftest.py
```diff
@@ -226,3 +226,27 @@ def given_a_role_exists(request, config, mock_resources, state):
...
+def given_more_than_100_users_exist(request, config, responses):
+    total = 150
+    per_page = 100
+
+    data = [
+        User(email=f'test{i}@example.com', connection='email')
+        for i in range(total)
+    ]
+
+    for i in range(0, total, per_page):
```
xoen: Is `i` the `start` here right?

moj_analytics/auth0_client.py
andyhd: It's being explicit about using a copy of the `defaults` dictionary, and not modifying it. It's not really necessary here because, as you say, the defaults dictionary is created just before.

moj_analytics/auth0_client.py
andyhd: You're right - this code does not handle modification of the list between requests for pages. I will try to make it more robust.

moj_analytics/auth0_client.py
andyhd: Good point - request does not raise an exception for non-200 responses.

moj_analytics/tests/conftest.py
andyhd: Yep, I'll fix that

moj_analytics/tests/conftest.py
andyhd: Yep

PR #294, opened 2018-07-25
Switch from pycrypto dep because it is not maintained

davidread:
## What

Switch from the 'pycrypto' dependency, because it is not maintained.
See discussion:
https://trello.com/c/hcCBYGHn/1274-pycrypto-in-controlpanel-is-not-maintained

## How to review

The tests pass, and includes some stuff on authenticating JWKs. I've not dug into it any further than this. I've not got controlpanel API and front-end running locally - perhaps someone who has try it or maybe we simply try on dev? I think logging in with Github should be enough of a test.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,14 +35,13 @@ py==1.5.4
 pyasn1==0.4.3
 pyasn1-modules==0.2.2
 pycparser==2.18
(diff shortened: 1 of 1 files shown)


davidread:
This deployed fine to dev and I can login fine, so all seems fine.

PR #295, opened 2018-07-25
Enable bulk adding customers

andyhd:
## What

* Enable adding customers in bulk by submitting a delimited string of email addresses. Delimiters may be any of space, comma or semi-colon, and may be mixed in the same string.
* Raise a ValidationError if any of the email addresses are invalid - no customers will be created or added to the app if any of the emails is invalid
* Also fixed a bug with listing customers multiple times

## How to review

1. Using an HTTP client, POST a list of email addresses to the add-customers API endpoint, eg:
```shell
curl --header "Content-Type: application/json" \
     --user "your-username:your-password" \
     --request POST \
     --data '{"email":"test1@example.com, test2@example.com, test3@example.com"}' \
     http://localhost:8000/apps/1/customers/
```
2. See that multiple customers are added to the app

See [Trello #1275](https://trello.com/c/N1bJpFMl)

### DIFF
--- a/control_panel_api/serializers.py
+++ b/control_panel_api/serializers.py
@@ -262,7 +262,7 @@ class UserSerializer(serializers.ModelSerializer):
 
 
 class AppCustomerSerializer(serializers.Serializer):
-    email = serializers.EmailField()
+    email = serializers.CharField()
     user_id = serializers.CharField(max_length=64, required=False)
     nickname = serializers.CharField(max_length=64, required=False)
     name = serializers.CharField(max_length=64, required=False)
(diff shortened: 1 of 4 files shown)


control_panel_api/serializers.py
```diff
@@ -262,7 +262,7 @@ class Meta:
 
 
 class AppCustomerSerializer(serializers.Serializer):
-    email = serializers.EmailField()
```
r4vi: why not? `EmailSerializer(many=True)` or does that only work with ORM backed models?

andyhd: What is `EmailSerializer`? I can't find it in DRF. `EmailField` does not support `many=True` as an argument.

r4vi: I was guessing there would be one but `many=True` thing works for any serializer but this is a field so :+1: LGTM

PR #343, opened 2018-10-15
allow JupyterLab deployment via API endpoint

r4vi:
## Allow the cpanel to deploy jupyter in addition to rstudio.

allows the cpanel frontend to query what tools are available to deploy and
then deploy them.

1. Run this locally alongside the corresponding [frontend](https://github.com/ministryofjustice/analytics-platform-control-panel-frontend/tree/deploy-jupyter-lab)
2. go to the apps tab in the frontend
3. you should see the option to deploy jupyterlab (clicking it may not work unless you have the [correct environment](https://github.com/ministryofjustice/analytics-platform-config/pull/94) variables set)

https://trello.com/c/2i9vRoTQ/1338-stretch-jupyter-deployable-from-the-control-panel


r4vi:
[Stretch:  Jupyter deployable from the control panel](https://trello.com/c/2i9vRoTQ/1338-stretch-jupyter-deployable-from-the-control-panel)

control_panel_api/tools.py
```diff
@@ -56,8 +74,61 @@ def deploy_for(self, user):
...
+
+class RStudio(BaseTool):
+    name = 'rstudio'
+
+
+class JupyterLab(BaseTool):
+    name = 'jupyter-lab'
+
+    def release_name(self, username):
+        return f'{self.name}-{username}'
```
xoen: Do we want the release name to be different for different tools? I'd say it would be best to be consistent unless there is some major reason to not to. My 2p of course.

control_panel_api/tools.py
```diff
@@ -56,8 +74,61 @@ def deploy_for(self, user):
...
+
+    def deploy_params(self, user):
```
xoen: I see that this is different from RStudio params, I wonder if method in base class could return list of common ones and then we merge with tool-specific params in the subclass.

Alternatively if we override in both subclasses maybe the base class could be kind of abstract and these method would just raise an exception to tell to implement these methods.

Not that this is not work or anything, just a thought.

control_panel_api/tools.py
```diff
@@ -19,35 +50,22 @@ class UnsupportedToolException(APIException):
     default_code = 'unsupported_tool'
 
 
-class Tool(object):
-
-    def __init__(self, name):
-        if not name in SUPPORTED_TOOL_NAMES:
-            raise UnsupportedToolException
+class BaseTool(HelmToolDeployMixin, Auth0ClientConfigMixin):
+    name = None
```
xoen: This somehow confuses me. Sorry I know it's a stupid one. 

`name` is a class variable as it's attached to the class (e.g. to the `RStudio`), but then it's accessed using `self.name`, as if it was an instance variable. 

Is it because Python somehow fallbacks to class variables when it doesn't find them at instance scope?

control_panel_api/utils.py
```diff
@@ -30,3 +30,17 @@ def sanitize_dns_label(label):
...
+    make a string with hyphens into a string where those hyphens have been replaced with _
+    :param name: name of variable to sanitize
+    :type str
+    :return: sanitized name
+    :rtype str
+    """
+    # uppercase
+    name = name.upper()
+    # replace - with _
+    name = name.replace('-', '_', )
```
xoen: What's with the trailing comma?

control_panel_api/views.py
```diff
@@ -270,12 +270,19 @@ class UserAppViewSet(viewsets.ModelViewSet):
...
+    tools = [{"name": n} for n in SUPPORTED_TOOL_NAMES]
+    return JsonResponse({"results": tools})
+
 
 @api_view(['POST'])
 @permission_classes((ToolDeploymentPermissions,))
 @handle_external_exceptions
 def tool_deployments_list(request, tool_name):
-    tool = Tool(tool_name)
+    tool = Tools[tool_name]()
```
xoen: I wonder if `Tools` => `ToolsCatalog` or registry to be on the explicit side of things.

control_panel_api/tools.py
```diff
@@ -1,13 +1,44 @@
...
+    def helm(self):
+        return Helm()
+
+
+class Auth0ClientConfigMixin:
+
+    def _get_auth_client_config(self, key):
+        setting_key = sanitize_environment_variable(
+            f'{self.name}_AUTH_CLIENT_{key}')
+        return getattr(settings, setting_key.upper())
```
xoen: I may be wrong but you don't need `upper()` here as it's done inside `utils.sanitize_environment_variable()`

control_panel_api/tools.py
```diff
@@ -56,8 +74,61 @@ def deploy_for(self, user):
...
+    name = 'jupyter-lab'
+
+    def release_name(self, username):
+        return f'{self.name}-{username}'
+
+    def deploy_params(self, user):
+        auth_proxy_cookie_secret = secrets.token_hex(32)
+
+        return [
+            '--set', f'Username={user.username.lower()}',
```
xoen: Maybe this is a case for factoring the common params in the common class, we had *a lot* of silly bugs caused by usernames case in the past 😎

PR #344, opened 2018-10-18
Improve resilience of 'user create'

davidread:
## What

Setting up accelerator I have seen all sorts of errors when a user first logs into the control panel. It's a pain to have to manually reset it by deleting the user from the control panel's db. Errors happen occasionally, so it's good to be resilient.


control_panel_api/authentication.py
```diff
@@ -60,15 +60,38 @@ def get_or_create_user(decoded_payload):
...
         user = User.objects.get(pk=decoded_payload.get('sub'))
     except User.DoesNotExist:
-        user = User.objects.create(
-            pk=decoded_payload.get('sub'),
-            username=decoded_payload.get(settings.OIDC_FIELD_USERNAME),
-            email=decoded_payload.get(settings.OIDC_FIELD_EMAIL),
-            name=decoded_payload.get(settings.OIDC_FIELD_NAME),
-        )
-        user.save()
+        pass
```
xoen: Are you sure about this logic? Maybe I'm missing something but the indentation may be off.

For example, we ignore the case when the user doesn't exist (it's new) because of the `pass`. But then outside the `exept` block we log `User "{}" is new [...]` which may not be the case.

Then we call `User.objects.create()` regardless of whethere the user was already in the DB or not. I don't remember how Django will handle that, it may raise an exception which we don't handle. Even if it doesn't we're just ignoring the result of `User.obects.get()`.

control_panel_api/authentication.py
```diff
@@ -60,15 +60,38 @@ def get_or_create_user(decoded_payload):
...
+        email=decoded_payload.get(settings.OIDC_FIELD_EMAIL),
+        name=decoded_payload.get(settings.OIDC_FIELD_NAME),
+    )
+    user.save()
+    logger.info('User {} saved to db as: "{}"'
+                .format(decoded_payload.get('sub'), user))
+    try:
         user.aws_create_role()
         user.helm_create()
+    except Exception:
```
xoen: I'm not a great fan of `except Exception` as it's very generic and usually is a bad idea.

I see that we're kind of rolling back the creation of the user if there was an exception, a couple of thoughts on this:
- We're assuming the user wasn't there (bear in mind the method is called `get_or_create_user()`), but what if it was? e.g. a random exception occurred, and then my user is deleted which is not good.
- IAM role creation and helm charts installation were previously done only when the user was created, this behaviour now changed and this could cause more exceptions (e.g. because trying to create an IAM role which is already there)
- As we only rollback the user creation (or at least that's the intent) maybe this should be instead wrapped in a DB transaction so that the DB/django takes care of it. Revert changed to AWS/helm could be more complicated and probably not worth.

xoen:
Uh sorry @davidread I just realised this was a work in progress and you did *NOT* requested my review - sorry, I got an email from GitHub saying you requested my review :( 

BTW: Sorry for the annoying comments, if they're confusing we can have a chat on Monday :)

control_panel_api/authentication.py
davidread: My fault - I forgot to add a return when the user does exist. That changes the logic drastically!

davidread: My fault - I forgot to add a return when the user does exist. That changes the logic drastically!

davidread:
Sorry, I asked for a review and then changed my mind when the tests failed. Not ideal that you still have an email in your inbox in this case - I should have said. I'll let you know when it's ready to look at again and sorry for the head scratching this has caused...

control_panel_api/authentication.py
davidread: I like the idea to use a db transaction to roll-back the creation of the user

andyhd:
I'm closing this PR (but keeping the branch) as it is now quite far out of sync with the master branch and requires a rewrite.

PR #421, opened 2018-11-28
Fix psycopg2 warning over module rename

davidread:
The python module has changed name. This is the warning I was seeing: `/Users/davidread/ap/analytics-platform-control-panel/venv/lib/python3.7/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.`


davidread:
Strange, despite the warning saying it was being renamed, the docs it links to says: "The -binary package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements. If you are the maintainer of a publish package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency. For production use you are advised to use the source distribution."

This suggests they changed their minds about a straight rename. Indeed with psycopg2==2.7.6.1 the warning is gone, so let's stick with that.

davidread:
Oh the warning is back. I don't know.

PR #492, opened 2019-01-17
pyyaml upgrade

davidread:
Upgrade to the latest (pre-release) pyyaml for security fix.

pyyaml is used by the kubernetes python client. They [looked at upgrading their requirements](https://github.com/kubernetes-client/python/pull/718) but decided in favour of [changing their calls of yaml.load to yaml.safe_load](https://github.com/kubernetes-client/python-base/pull/111), however that is not yet released.

So in the meantime I think it is fine to use this pre-release pyyaml. The only checks I've done so far are seeing it pass the tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,7 +43,7 @@ pytest-watch==4.2.0
 python-dateutil==2.7.5
 python-jose[cryptography]==3.0.1
 pytz==2018.9
-PyYAML==3.13
+PyYAML==4.2b4
 raven==6.10.0
 requests==2.21.0
 rsa==4.0


andyhd:
As @r4vi [said on my PR](https://github.com/ministryofjustice/analytics-platform-ops/pull/246#discussion_r245681002), it feels weird to pin to a beta version, so maybe do `>=4.2b4,<5`

davidread:
@andyhd Yeah there's defo a case for that. It's the classic deps debate... We could consider having both a loose-defined top-level requirements file, which is used to generate a precisely pinned packages.lock file (e.g. with pip freeze, pip-tools or pipenv). But since we only have one requirements file, and I think it's key purpose is reproducibility between environments, so I reckon we should stick to pinning exactly, even if a beta version. At least for this repo, Dependabot will upgrade it to the next non-beta version anyway, as soon as it's released.

davidread:
Released in https://github.com/ministryofjustice/analytics-platform-control-panel/releases/tag/v0.13.3

PR #504, opened 2019-01-25
Fix kubernetes client patch for RBAC

andyhd:
## What

* kubernetes python client has changed and the patch no longer applies, so I updated it
* I haven't checked whether this patch is still necessary

### DIFF
--- a/control_panel_api/k8s_patch.py
+++ b/control_panel_api/k8s_patch.py
@@ -7,15 +7,8 @@ import kubernetes
 from kubernetes.config.kube_config import _is_expired
 
 
-def load_token(self):
-    if 'auth-provider' not in self._user:
-        return
-
-    provider = self._user['auth-provider']
-
-    if ('name' not in provider
-            or 'config' not in provider
-            or provider['name'] != 'oidc'):
+def load_token(self, provider):
+    if 'config' not in provider:
         return
 
     parts = provider['config']['id-token'].split('.')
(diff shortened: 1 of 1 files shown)


control_panel_api/k8s_patch.py
```diff
@@ -7,15 +7,8 @@
...
-def load_token(self):
-    if 'auth-provider' not in self._user:
-        return
-
-    provider = self._user['auth-provider']
-
-    if ('name' not in provider
-            or 'config' not in provider
-            or provider['name'] != 'oidc'):
+def load_token(self, provider):
```
xoen: I see that `provider` was taken out from the `self._user['auth-provider']` before but now it's a parameter of `load_token()` - I can't see the change to the client code calling this tho, what am I missing?

andyhd: `load_token` is a monkey patch for [`KubeConfigLoad._load_oid_token`](https://github.com/kubernetes-client/python-base/blob/bd9a8525e9215f7f01c32a321beb9a605cf0402b/config/kube_config.py#L258) in the Python Kubernetes client library. They have a [fix for the issue in a PR](https://github.com/kubernetes-client/python-base/pull/79), but it has been waiting to be merged for months. Hopefully they will merge it soon and we can remove this patch.
In the meantime, they changed the signature of `_load_oid_token` to take the `provider` dict, so this change is to keep up.

andyhd:
Proper fix in another PR

PR #508, opened 2019-01-29
Tidy up, fix warnings and use a Makefile

andyhd:
## What

* Tidy up
  * Removed scripts to run server and tests - moved into Makefile
  * ~~Moved scripts from Dockerfile into Makefile~~ removed to avoid docker image dependency on `make`
  * Updated README
  * Reduced Docker layers
  * Removed deprecated workaround for building `cryptography`
  * Moved static directory out of source code directory to avoid Docker image build issues
&nbsp;
* Fixed Kubernetes client patch for OIDC auth

* Fixed warnings
  * `psycopg2` now called `psycopg2-binary`
  * DRF `field_filter` now called `fieldset_filter`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,63 +2,42 @@ FROM alpine:3.7
 
 MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
 
-# install build dependencies (they'll be uninstalled after pip install)
-RUN apk add --no-cache \
-        --virtual build-deps \
-        gcc \
-        musl-dev
-
-# install python3 and 'ca-certificates' so that HTTPS works consistently
-RUN apk add --no-cache \
-        openssl \
-        ca-certificates \
-        libffi-dev \
-        python3-dev
-
-# Temporary bugfix for libressl
-# Postgres needs libressl-dev, but cryptography only works with openssl-dev
-RUN apk add --no-cache --virtual temp-ssl-fix \
-        openssl-dev \
-    && pip3 install cryptography==2.2.2 \
-    && apk del temp-ssl-fix \
-    && apk add --no-cache \
-        libressl-dev \
-        postgresql-dev
-
-# Install helm
 ENV HELM_VERSION 2.9.1
-RUN wget https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VERSION-linux-amd64.tar.gz \
-    && tar xzf helm-v$HELM_VERSION-linux-amd64.tar.gz \
-    && mv linux-amd64/helm /usr/local/bin \
-    && rm -rf helm-v$HELM_VERSION-linux-amd64.tar.gz linux-amd64
-
-# Configure helm
 ENV HELM_HOME /tmp/helm
-RUN helm init --client-only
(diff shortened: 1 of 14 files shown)


andyhd:
Yeah, I'll change the README to use pip and manage.py.

Dockerfile
```diff
@@ -2,63 +2,42 @@ FROM alpine:3.7
...
-
-ADD manage.py manage.py
-ADD run_api run_api
-ADD run_tests run_tests
-ADD wait_for_db wait_for_db
-ADD control_panel_api control_panel_api
-ADD moj_analytics moj_analytics
+COPY control_panel_api control_panel_api
+COPY moj_analytics moj_analytics
+COPY manage.py wait_for_db ./
```
xoen: Is this `wait_for_db` still used?

r4vi: if you're using docker-compose it's useful

xoen: Right. I was following the `Makefile` dependencies 🎢  and the `wait_for_db` task is a dependency of `migrations` which is run by `docker-compose` 😕  👍

Dockerfile
```diff
@@ -2,63 +2,42 @@ FROM alpine:3.7
...
 
-# install build dependencies (they'll be uninstalled after pip install)
-RUN apk add --no-cache \
-        --virtual build-deps \
-        gcc \
-        musl-dev
+ENV HELM_VERSION 2.9.1
+ENV HELM_HOME /tmp/helm
+ENV DJANGO_SETTINGS_MODULE "control_panel_api.settings"
+ENV USE_VENV=false
```
xoen: Why there is this flag? Maybe it's a stupid question but why someone would want to enable/disable this while running the Docker image?

Dockerfile
xoen: Yeah, I eventually found from where this script was invoked 👍

Dockerfile
andyhd: This is because if you run `make dependencies` outside of the docker container, it will use a virtualenv by default. In the docker container, we don't want to use a virtualenv, so this environment variable prevents that.

andyhd: I've removed this now, because I removed the `make` dependency in the docker image.

Dockerfile
andyhd: I removed `make` from the docker image, so the `docker-compose.yml` now runs `wait_for_db` directly.

Dockerfile
xoen: Thanks for the clarification @andyhd 👍 

Just out of curiosity, why we don't want to use virtualenv in the Docker container? What are the drawbacks of that?

Isn't virtualenv pretty much creating a glorified simlink to a directory? That should't have much of a performance hit - but maybe I'm oversimplifying here.

andyhd: Mainly for convenience - if you want to run `manage.py` for example inside the docker container, you would have to write `docker run -it control-panel venv/bin/python3 manage.py shell`. A virtualenv is unnecessary inside a docker container, so why create it in the first place?

xoen: Oh of course! Unless we don't symlink that `venv/bin/python3` binary. But you're right, I see why it's more convenient this way 👍

PR #610, opened 2019-03-27
Merge Control Panel frontend webapp into REST API webapp

andyhd:
## ⚠️  This is a very large PR!
I've provided a walkthrough to point out the changes in easy to follow way

# 🚦 [Start the Walkthrough](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-43e3cccac20654f672ba2b01e462abf9R24)
Or jump to a section using the links below.

1. **[Move REST API into its own app](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-43e3cccac20654f672ba2b01e462abf9R24)**
    I restructured the Django project into 3 main apps, `api`, `frontend` and `kubeapi`. The `api` app is the old `control_panel_api` app moved and renamed. There were a couple of tweaks necessary to achieve this:

    1a. **[Fix migrations](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-43e3cccac20654f672ba2b01e462abf9R24)**
         Renaming the app meant the existing migrations refer to a non-existent app, so they had to be tweaked to allow them to run without complaints or creating duplicate tables.

    1b. **[Update models](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-0fe5e5e3915cf36fbd27ef47844ab9d9R25)**
        The models also needed to refer to the old table names to avoid creating duplicate tables.

    And of course, I can't touch a codebase without refactoring large chunks of it. I managed to restrain myself to the following:

    1c. **[Refactor Helm client module](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-879606017c4e0c05d0a2e8a160fa114cR19)**

    1d. **[Refactor Auth0 client module](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-0fe5e5e3915cf36fbd27ef47844ab9d9R117)**

    1e. **[Refactor Tools code](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-aad7b41260663fb400ff942a02fd3d14R34)**

2. **[Move Kubernetes API proxy into its own app](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-a42ab3147196db399d3b62ff2d2acd73R43)**
    Next, I separated the Kubernetes API proxy. It seems a logical separation, as it shares no code with the REST API or the frontend.

    2a. **[Monkey-patching the Kubernetes Python client](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-8c49a844762f3e9fdeb511f2a34ca23eR19)**
    Unfortunately OIDC support in the official Kubernetes Python library has a few bugs, so we have to monkey-patch them while we wait for the fixes.

3. **[Replace custom OIDC client code with `mozilla-django-oidc`](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-33610d8b51443bfb29fe9662b597d96eR63)**
    Removing our own code in favour of off-the-shelf libraries is a no-brainer. The Mozilla module integrates nicely with Django and DRF.

    3a. **[Recognise users by their Auth0 `sub` claim](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-27bdf3c6238a9803b6ead711b23e6490R10)**
    Just a slight tweak to the defaults was needed, to find the authenticated user in the database by using the Auth0 `sub` claim.

4. **[Create `frontend` app](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-7917d2dd3d7b1cefc943f94b82f977deR6)**
    Finally, the point of this PR. Rather than maintaining a Node.JS webapp as well as a Django REST API, we combine the two - reducing the number of deployment pipelines, dependencies to track, and moving parts in the Platform generally. Fortunately, the Node.JS app uses Nunjucks templates, so it was a case of tweaking these to work with Jinja2, rather than starting from scratch.

    4a. **[Views]()**
    These are the meat of the frontend app.

    4b. **[Jinja2 templates]()**
    I've used the Django Jinja2 template renderer, because the templates were written in Nunjucks, which is based on Jinja.

    4c. **[GOV.UK Frontend and Design System]()**
    The Node.JS app used the now deprecated GOV.UK Elements stylesheets. I have upgraded to the [GOV.UK Design System](https://design-system.service.gov.uk/) and created some custom components based on their system.

    4d. **[Server Sent Events and ASGI]()**
    This is a work in progress. In the Node.JS app, we use websockets to update the Tool status in pseudo-realtime. I've started work on using [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) to achieve the same, as we don't need full two-way websocket functionality.

5. **[Miscellaneous]()**
    I tried to tidy up the codebase and documentation - specifically the Django settings module. There's also a nice Makefile and up-to-date Dockerfile and docker-compose setup for development, testing and deployment.


controlpanel/api/migrations/0001_initial.py
```diff
@@ -21,6 +21,8 @@ class Migration(migrations.Migration):
         ('auth', '0008_alter_user_username_max_length'),
     ]
 
+    replaces = [('control_panel_api', '0001_initial')]
+
```
andyhd: ## Move REST API into its own app
Unfortunately, simply moving the code into a new subfolder and renaming the app leaves the migrations pointing to the old app name (`control_panel_api`). I wanted to keep the existing migrations and avoid having to run fakes because nothing has actually changed in the models. Fortunately, I found [this handy Stack Overflow question and answer on how to move/rename Django apps with migrations](https://stackoverflow.com/questions/44957167/django-renaming-app-and-migrations).

`replaces` makes sure these "new" migrations don't run if the old ones are already in the database.

controlpanel/api/migrations/0001_initial.py
```diff
@@ -40,6 +42,7 @@ class Migration(migrations.Migration):
                 ('groups', models.ManyToManyField(blank=True, help_text='The groups this user belongs to. A user will get all permissions granted to each of their groups.', related_name='user_set', related_query_name='user', to='auth.Group', verbose_name='groups')),
             ],
             options={
+                'db_table': 'control_panel_api_user',
```
andyhd: I kept the old table names to avoid disturbing the database too.

controlpanel/api/migrations/0001_initial.py
```diff
@@ -68,9 +72,10 @@ class Migration(migrations.Migration):
                 ('created', django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created')),
                 ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified')),
                 ('access_level', models.CharField(choices=[('readonly', 'Read-only'), ('readwrite', 'Read-write')], default='readonly', max_length=9)),
-                ('app', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='apps3buckets', to='control_panel_api.App')),
+                ('app', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='apps3buckets', to='api.App')),
```
andyhd: All the foreign keys needed to be updated to point to the moved models

controlpanel/api/migrations/0006_user_last_name_length.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('api', '0005_s3bucket_url'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='last_name',
+            field=models.CharField(blank=True, max_length=150, verbose_name='last name'),
```
andyhd: This new migration seems to be a left-over from an old change which was never committed.

### [ ⏩ Next ]()

controlpanel/api/models.py
```diff
@@ -24,6 +22,7 @@ class User(AbstractUser):
     REQUIRED_FIELDS = ['email', 'auth0_id']
 
     class Meta:
+        db_table = 'control_panel_api_user'
```
andyhd: The models all specify the old table names so that a bunch of duplicate tables called `api_user`, etc are not created.

### [ ⏩ Next ]()

controlpanel/api/models.py
```diff
@@ -50,27 +53,51 @@ def aws_delete_role(self):
...
+            f"init-user-{user_slug}",
+            "mojanalytics/init-user",
+            f"--set={values}",
+        )
+        helm.upgrade_release(
+            f"config-user-{user_slug}",
+            "mojanalytics/config-user",
+            "--namespace=user-{user_slug}",
+            "--set=Username={user_slug}",
+        )
```
andyhd: #### 1b. Update models

> Lines 55 - 75

I refactored our Helm client module and this calling code to fix the levels of abstraction. The module now only provides an interface to Helm. The model knows how to invoke Helm when creating a User.

[Next →]()

controlpanel/api/models.py
```diff
@@ -50,27 +53,51 @@ def aws_delete_role(self):
...
-class App(TimeStampedModel):
-    def _slugify(name):
-        """Create a slug using standard django slugify but we override with
-        extra replacement so it's valid for s3"""
-        return re.sub(r'_+', '-', slugify(name))
+    def is_bucket_admin(self, bucket_id):
+        return self.users3buckets.filter(
+            s3bucket__id=bucket_id,
+            is_admin=True,
+        ).count() != 0
```
andyhd: #### 1b. Update models

> Lines 82 - 89

These new methods are used in the frontend templates

[Next →]()

controlpanel/api/helm.py
```diff
@@ -0,0 +1,150 @@
...
+            timeout = kwargs.pop('timeout')
+
+        try:
+            proc = subprocess.Popen(
+                ["helm", *args],
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                encoding='utf8',
+                **kwargs,
+            )
```
andyhd: ### Refactor Helm client module

> Lines 1 - 88

I simplified the Helm client code, which had two wrapper methods around `subprocess` calls. Instead, we have a single method to execute Helm and potentially run an asynchronous process for later use in Server Sent Events.

[Next →]()

controlpanel/api/helm.py
```diff
@@ -0,0 +1,150 @@
...
+                *resources.get(resource_type, []),
+                *[row],
+            ]
+
+    return {
+        'last_deployed': last_deployed,
+        'namespace': namespace,
+        'resources': resources,
+        'notes': '\n'.join(notes),
+    }
```
andyhd: #### b. Parse Helm output

> Lines 91 - 147

This is WIP code. My intention is to parse the output of the Helm command and be able to pass a JSON object to the client via SSEs.

[Next →]()

controlpanel/api/models.py
```diff
@@ -85,39 +112,25 @@ def aws_delete_role(self):
...
-        if repo_name.endswith('.git'):
-            repo_name = repo_name[:-4]
-
-        return repo_name.rsplit('/', 1)[1]
-
-    def get_customers(self):
-        return Auth0().get_group_members(group_name=self.slug)
+    @property
+    def customers(self):
+        return auth0.AuthorizationAPI().get_group_members(group_name=self.slug)
```
andyhd: #### 1d. Refactor Auth0 client module

> Lines 117 - 119

Our Auth0 client code was in a separate module `moj_analytics.auth0_client`, which was copied from another codebase. This code was too complicated - I wrote it 😞

[Next →]()

controlpanel/api/auth0.py
```diff
@@ -0,0 +1,190 @@
...
+    base_url = None
+    audience = None
+
+    def __init__(self, **kwargs):
+        self.client_id = kwargs.get("client_id", settings.AUTH0["client_id"])
+        self.client_secret = kwargs.get(
+            "client_secret", settings.AUTH0["client_secret"]
+        )
+        self.domain = kwargs.get("domain", settings.AUTH0["domain"])
+        self._access_token = None
```
andyhd: #### 1d. Refactor Auth0 client module

> Lines 1 - 190

The old client code was too indirect, and relied on subclassing `dict` and some difficult-to-follow inheritance and functional style.

Now we have a straightforward base class which provides methods to make authenticated API requests and wraps fetching all results across pagination...

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-4abd63b561b5260e074659dad215fa1cR109)

controlpanel/api/auth0.py
```diff
@@ -0,0 +1,190 @@
...
+            return None
+
+        return groups[0]
+
+    def get_group_members(self, group_name):
+        group = self.get_group(group_name)
+        if group:
+            return self.get_all(
+                f'groups/{group["_id"]}/members', key="users", per_page=25
+            )
```
andyhd: #### 1d. Refactor Auth0 client module

> Lines 109 - 173

...and subclasses for the Authorization API, providing convenience methods wrapping API requests to manage group members...

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-4abd63b561b5260e074659dad215fa1cR176)

controlpanel/api/auth0.py
```diff
@@ -0,0 +1,190 @@
...
+        response = self.request(
+            "POST",
+            "users",
+            json={"email": email, "email_verified": email_verified, **kwargs},
+        )
+
+        if "error" in response:
+            raise Auth0Error("create_user", response)
+
+        return response
```
andyhd: #### 1d. Refactor Auth0 client module

> Lines 176 - 190

...and for the Management API, providing a convenience method to create a user in Auth0.

controlpanel/api/tools.py
```diff
@@ -0,0 +1,212 @@
...
+
+class Tool(metaclass=ToolMeta):
+    name = None
+
+    @classmethod
+    def create(cls, name):
+        try:
+            return tools_repository[name]()
+        except KeyError:
+            raise UnsupportedToolException()
```
andyhd: #### 1e. Refactor Tools code
> Lines 41 - 130
I refactored the ToolRepository class into a simple dict, with entries automatically added by the `ToolMeta` class when a `Tool` subclass is defined. *RETHINK THIS*

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+        return super().__getitem__(key)
+
+    def __setitem__(self, key, value):
+        if key not in SUPPORTED_TOOL_NAMES:
+            raise UnsupportedToolException(key)
+
+        if key in self.tools:
+            raise DuplicateTool(key)
+
+        super().__setitem__(key, value)
```
andyhd: #### 1e. Refactor Tools code
> Lines 34 - 52

I updated the ToolRepository class to disallow adding unsupported tool names or duplicates

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-aad7b41260663fb400ff942a02fd3d14R55)

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+
+    # repository is shared across all Tool subclasses
+    _repository = ToolRepository()
+
+    def __init__(cls, name, bases, clsdict):
+        super().__init__(name, bases, clsdict)
+        if 'name' not in clsdict:
+            raise AttributeError(f"{name}.name")
+        if name != "Tool":
+            cls._repository[clsdict['name']] = cls
```
andyhd: #### 1e. Refactor Tools code
> Lines 55 - 66

This metaclass automatically adds subclasses of `Tool` to a shared `ToolRepository` instance. This removes (hides) the module-level `Tools` variable and encapsulates it to avoid it being messed with. Also, we no longer have to remember to add new `Tool` subclasses to the repository manually.

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-aad7b41260663fb400ff942a02fd3d14R69)

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+    def get_user_deployment(self, user):
+        deployments = client.AppsV1Api().list_namespaced_deployment(
+            namespace=user.k8s_namespace,
+            label_selector=f"app={self.name}"
+        )
+
+        if len(deployments.items) < 1:
+            raise ToolNotDeployed()
+
+        return ToolDeployment(deployments.items[0], user)
```
andyhd: #### 1e. Refactor Tools code
> Lines 123 - 132

The main change to the base `Tool` class is to add a method to get the Kubernetes Deployment of the tool for a user.

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-aad7b41260663fb400ff942a02fd3d14R158)

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+    @property
+    def idled(self):
+        return self.deployment.metadata.labels.get(IDLED) == "true"
+
+    def restart(self):
+        return client.AppsV1Api().delete_collection_namespaced_replica_set(
+            namespace=self.user.k8s_namespaces,
+            label_selector=f"app={self.name}",
+            watch=True,
+        )
```
andyhd: #### 1e. Refactor Tools code
>Lines 158-206

The `ToolDeployment` class is a wrapper around the Kubernetes Deployment for a specified tool.

It provides a class method to get a list of all a user's tool Deployments, a property to list a specific Deployment's Pods, and a method to restart a tool (by deleting all the pods, prompting Kubernetes to spin up new ones).

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-aad7b41260663fb400ff942a02fd3d14R209)

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+                        reason = f"ExitCode:{state.terminated.exit_code}"
+                return {"phase": "terminated", "reason": reason}
+
+        return {"phase": self.pod.status.phase}
+
+    @property
+    def display_status(self):
+        status = self.status
+        reason = f":{status['reason']}" if status.get("reason") else ""
+        return f"{status['phase']}{reason}"
```
andyhd: #### 1e. Refactor Tools code
>Lines 209 - 237

The `Pod` class has been copied from the frontend Node webapp. It provides a property to calculate the status of a Kubernetes Pod, which is used to determine the status of a tool Deployment.

[Next →]()

controlpanel/kubeapi/views.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Proxy requests to the Kubernetes cluster API
+    """
+
+    base_url = client.Configuration().host
+
+    # Without this, we get SSL: CERTIFICATE_VERIFY_FAILED
+    verify_ssl = client.Configuration().ssl_ca_cert
+
+    proxy_middleware = ["controlpanel.kubeapi.views.KubeAPIAuthMiddleware"]
```
andyhd: ### 2. Move Kubernetes API proxy into its own app
> Lines 43 - 53

I junked the existing proxy view function we had in favour of [djproxy](https://github.com/thomasw/djproxy) on the principle that it is always better to use off-the-shelf code than maintain and worry about the correctness of our own implementation.

As a result, the Kubernetes API proxy is reduced to a few lines of code configuring the proxy View class, and a middleware object that inserts the request user's authentication token into proxied requests to the Kubernetes API.

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-8c49a844762f3e9fdeb511f2a34ca23eR77)

controlpanel/kubeapi/oidc_patch.py
```diff
@@ -0,0 +1,121 @@
...
+    else:
+        config.verify_ssl = False
+
+    client = kubernetes.client.ApiClient(configuration=config)
+
+    response = client.request(
+        method="GET",
+        url="%s/.well-known/openid-configuration"
+        % provider['config']['idp-issuer-url'].rstrip('/'),
+    )
```
andyhd: #### 2a. Monkey-patching the Kubernetes client
>Lines 74 - 78

I submitted this patch to the Kubernetes Python client project - we get a URL ending with a `/` from Auth0 for `idp-issuer-url` (the tenant base URL). But this code assumes there is no trailing slash and ends up creating a URL that is rejected by the Kubernetes API. 🙄 

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-8c49a844762f3e9fdeb511f2a34ca23eR113)

controlpanel/kubeapi/oidc_patch.py
```diff
@@ -0,0 +1,121 @@
...
+
+kubernetes.config.kube_config.KubeConfigLoader._refresh_oidc = refresh_oidc
+
+
+def get_yaml_config_loader(filename, **kwargs):
+    with open(filename) as f:
+        return kubernetes.config.kube_config.KubeConfigLoader(
+            config_dict=yaml.load(f, Loader=yaml.FullLoader),
+            config_base_path=os.path.abspath(os.path.dirname(filename)),
+            **kwargs)
```
andyhd: #### 2a. Monkey-patching the Kubernetes client
> Lines 113 - 118

This monkey-patch should be short-lived, as [the fix has already been accepted upstream](https://github.com/kubernetes-client/python-base/pull/111). The Kubernetes config YAML file was being loaded with `yaml.load`, which has an arbitrary code execution vulnerability.

[Next →]()

controlpanel/kubeapi/oidc_patch.py
```diff
@@ -0,0 +1,121 @@
...
+        log.debug(provider["config"]["idp-issuer-url"])
+
+        self._refresh_oidc(provider)
+
+        if self._config_persister:
+            self._config_persister(self._config.value)
+
+    self.token = f"Bearer {provider['config']['id-token']}"
+
+    return self.token
```
andyhd: #### 2a. Monkey-patching the Kubernetes Python client
>Lines 19 -46

The fix for this base64 padding issue still hasn't made it into the Python Kubernetes client.

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-8c49a844762f3e9fdeb511f2a34ca23eR52)

controlpanel/settings/common.py
```diff
@@ -0,0 +1,400 @@
...
+
+INSTALLED_APPS = [
+    # Django channels for asynchronous support
+    "channels",
+    # Django eventstream for SSE support
+    "django_eventstream",
+    # Django auth system
+    "django.contrib.auth",
+    # OIDC client
+    "mozilla_django_oidc",
```
andyhd: ### 3. Replace custom OIDC client code with `mozilla-django-oidc`
>Lines 62 - 63

I scrapped our custom authorization code in favour of Mozilla's Django OIDC module. There's this Django app and a middleware class.

[Next →](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-33610d8b51443bfb29fe9662b597d96eR128)

controlpanel/oidc.py
```diff
@@ -0,0 +1,44 @@
...
+    def filter_users_by_claims(self, claims):
+        sub = claims.get("sub")
+        if not sub:
+            return self.UserModel.objects.none()
+
+        try:
+            return User.objects.filter(pk=sub)
+
+        except User.DoesNotExist:
+            return self.UserModel.objects.none()
```
andyhd: #### 3a. Recognise users by their Auth0 `sub` claim
>Lines 128 - 128

We're using the user's Auth0 `sub` claim (eg: `github|1234`) as their user ID, so I subclassed the OIDC authentication backend to handle this.

[Next →]()

controlpanel/api/models.py
```diff
@@ -50,27 +57,54 @@ def aws_delete_role(self):
         services.delete_role(self.iam_role_name)
 
     def helm_create(self):
-        helm.init_user(self.username, self.email, self.get_full_name())
-        helm.config_user(self.username)
+        user_slug = sanitize_dns_label(self.username)
+        values = ",".join(f"{key}={value}" for key, value in {
```
xoen: we may want to be more explicit here about the fact that these are values for the `init-user` helm chart, maybe by renaming to `init_user_values`. Not a big deal anyway.

controlpanel/api/auth0.py
```diff
@@ -0,0 +1,211 @@
...
+
+class AuthorizationAPI(APIClient):
+    base_url = settings.AUTH0["authorization_extension_url"]
+    audience = "urn:auth0-authz-api"
+
+    def get_users(self):
+        return self.get_all("users", page=0, per_page=100)
+
+    def get_group(self, group_name):
+        groups = [
```
xoen: Why do you need this list? Isn't `get_all("groups")` returning a list already? Couldn't you assign that to `groups` and the rest of the logic stays the same? Probably there is a good reason why you're using list comprehension here.

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+    default_code = "unsupported_tool"
+
+
+class ToolRepository(UserDict):
+
+    def __getitem__(self, key):
+        if key not in self:
+            raise UnsupportedToolException(key)
+
+        if key not in self:
```
xoen: Isn't this `if` statement duplicated? Maybe you wanted something like `if key not in SUPPORTED_TOOL_NAMES` here as well? Not sure.

controlpanel/api/tools.py
```diff
@@ -0,0 +1,237 @@
...
+            *params,
+        )
+
+    def get_user_deployment(self, user):
+        deployments = client.AppsV1Api().list_namespaced_deployment(
+            namespace=user.k8s_namespace,
+            label_selector=f"app={self.name}"
+        )
+
+        if len(deployments.items) < 1:
```
xoen: Isn't a bit overkill to use exception just because a tool is not deployed for a user? Wouldn't return `None` be good enough?

controlpanel/api/migrations/0006_user_last_name_length.py
xoen: I'm a noob but does this also needs `replaces = [('control_panel_api', '0006_user_last_name_length')]`?

controlpanel/settings/common.py
```diff
@@ -0,0 +1,453 @@
...
+
+# Function called to logout of OIDC provider
+OIDC_OP_LOGOUT_URL_METHOD = "controlpanel.oidc.logout"
+
+# OIDC client secret
+OIDC_RP_CLIENT_ID = os.environ.get("OIDC_RP_CLIENT_ID")
+OIDC_RP_CLIENT_SECRET = os.environ.get("OIDC_RP_CLIENT_SECRET")
+
+# OIDC JWT signing algorithm
+OIDC_RP_SIGN_ALGO = os.environ.get("OIDC_RP_SIGN_ALGO", "HS256")
```
xoen: [This used to be `RS256`](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-8a1eae853f4aab578f972d4c4a41b747L97), I know it's a setting so it can be overridden but I thought it would be useful to point-out to avoid some pain in the future. We could change the default or be sure the config is `RS256` (unless we want for some reason to change the client but I'm not sure we can).

controlpanel/kubeapi/views.py
xoen: *Disclaimer*: I'm still wrapping my head around the whole thing.

Specifically, [I don't see the bit that we were doing where we were stripping the first few characters of the path](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-85ef3a34be3164c0b669e382d500904dL23) of the requests sent to k8s (because they were maid to `/k8s/`). 

But **the most important question is**: do we still need the proxy? I haven't look the whole thing but I see some places where k8s requests are made through the Python client. I can't really remember, sorry if this is really a stupid question.

controlpanel/settings/common.py
```diff
@@ -0,0 +1,453 @@
...
+# Store Server Sent Events for 24 hours
+EVENTSTREAM_STORAGE_CLASS = 'django_eventstream.storage.DjangoModelStorage'
+
+
+# -- Github
+
+# Allowed Github organizations
+GITHUB_ORGS = [
+    # 'ministryofjustice',
+    'moj-analytical-services',
```
xoen: We may want to have it an environment variable at some point in the future in case someone else wants to use this.

controlpanel/settings/common.py
```diff
@@ -0,0 +1,453 @@
...
+OIDC_OP_AUTHORIZATION_ENDPOINT = os.environ.get("OIDC_OP_AUTHORIZATION_ENDPOINT")
+OIDC_OP_JWKS_ENDPOINT = os.environ.get("OIDC_OP_JWKS_ENDPOINT", None)
+OIDC_OP_TOKEN_ENDPOINT = os.environ.get("OIDC_OP_TOKEN_ENDPOINT")
+OIDC_OP_USER_ENDPOINT = os.environ.get("OIDC_OP_USER_ENDPOINT")
+
+# Function called to logout of OIDC provider
+OIDC_OP_LOGOUT_URL_METHOD = "controlpanel.oidc.logout"
+
+# OIDC client secret
+OIDC_RP_CLIENT_ID = os.environ.get("OIDC_RP_CLIENT_ID")
```
xoen: I'm confused. Isn't this going to break the helm chart, etc? Unless I'm missing something this [used to be `OIDC_CLIENT_ID`](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-1fba42f0ae98b6a15ab28483ba2086b7L158). I see that internally the variable needs to be called `OIDC_RP_CLIENT_ID` because of the `mozilla_django_oidc` module, but do we need to break compatibility with the outside?

[Documentation](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-a8bc9331f35155a53c003183eef12992R15) still mentions the old variables.

Maybe I'm missing something here, apologies if that's the case.

doc/frontend.md
```diff
@@ -0,0 +1,21 @@
+# Frontend app
+
+The Control Panel Frontend is a OIDC login protected web application which
+allows users of the Analytical Platform to manage their hosted tools (including
+RStudion and Jupyter Lab), web apps and data sources. It also provides an admin
```
xoen: Typo: `RStudion` => `RStudio`.

Dockerfile
```diff
@@ -1,43 +1,64 @@
-FROM alpine:3.7
+FROM alpine:3.8 AS base
```
xoen: AFAICT this will run as `root`, ideally we'd want this to run as a normal user.

doc/environment.md
```diff
@@ -0,0 +1,24 @@
+# Environment Variables Reference
```
xoen: I think this documentation now needs love ❤️ - I'm trying to run this locally but some of these were renamed for some reason (as mentioned in another comment) and other variables seem new and it's not clear what's the correct value.

controlpanel/settings/common.py
davidread: Agreed. I'd suggest leaving that enhancement for another PR - minimize each PR and all that.

xoen: I agree on having smaller, more focused PRs 😁

r4vi:
trying to run it locally now :pray:

controlpanel/oidc.py
```diff
@@ -0,0 +1,44 @@
...
+
+class OIDCSubAuthenticationBackend(OIDCAuthenticationBackend):
+    """
+    Authentication backend which matches users by their `sub` claim
+    """
+
+    def create_user(self, claims):
+        user = User.objects.create(
+            pk=claims.get("sub"),
+            username=claims.get(settings.OIDC_FIELD_USERNAME),
```
r4vi: OIDC_FIELD_USERNAME isn't in claims when I try to log in causing this to fail. Not sure if it's something weird with my user or if this code path hasn't been tested

controlpanel/api/migrations/0006_user_last_name_length.py
andyhd: No, because there is no existing `0006_user_last_name_length` migration to replace.

controlpanel/api/models.py
andyhd: Actually, I want to refactor this a bit more, because the levels of abstraction are all wrong. I want to make it so that we can replace Helm if we want to (and I do want to).

controlpanel/api/tools.py
andyhd: This is a mistake - not sure what I meant to do here, but this duplication is wrong.

controlpanel/settings/common.py
andyhd: That's a mistake - I was copying code from the Mozilla Django OIDC module documentation and missed this.

controlpanel/settings/common.py
andyhd: Yes, I didn't mean to break the helm chart or rename these env vars - the names are from the Mozilla Django OIDC module and I must have forgotten to match them up.

doc/frontend.md
andyhd: 👀

Dockerfile
andyhd: 👍

doc/environment.md
andyhd: Yep, I was getting a bit tired towards the end of this PR and didn't keep the docs up-to-date - I'll fix them up.

Dockerfile
```diff
@@ -1,43 +1,64 @@
...
 
-MAINTAINER Andy Driver <andy.driver@digital.justice.gov.uk>
+LABEL maintainer="andy.driver@digital.justice.gov.uk"
 
-ENV HELM_VERSION 2.9.1
+ARG HELM_VERSION=2.13.0
```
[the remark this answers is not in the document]
andyhd: 👍

Makefile
```diff
@@ -1,48 +1,78 @@
...
+	@npm install
+	@cp -R node_modules/accessible-autocomplete/dist/ static/accessible-autocomplete
+	@cp -R node_modules/govuk-frontend/ static/govuk-frontend
+	@cp -R node_modules/@hmcts/frontend/ static/hmcts-frontend
+	@cp -R node_modules/html5shiv/dist/ static/html5-shiv
+	@cp -R node_modules/jquery/dist/ static/jquery
+
+transpile:
+	@echo
+	@echo "> Transpiling ES6..."
```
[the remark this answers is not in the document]
andyhd: I haven't tested without transpiling - we are doing it in the Node app, so I just copied it over. I'll give it a try.

controlpanel/api/helm.py
```diff
@@ -0,0 +1,150 @@
...
+def parse_upgrade_output(output):
+    section = None
+    columns = None
+    last_deployed = None
+    namespace = ''
+    resource_type = None
+    resources = {}
+    notes = []
+
+    for line in output.split('\n'):
```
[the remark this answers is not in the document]
andyhd: Yep, it really would be good to get JSON output from Helm

controlpanel/api/tools.py
andyhd: It's only a little one. But I think it's a good idea for the reasons I've given. What are your reasons against it?

controlpanel/api/auth0.py
andyhd: Did you miss the `if group["name"] == group_name` in the list comprehension? This is filtering the list of groups from `get_all("groups")` to the one (or more if there are duplicates) named group. The Authorization API only allows you to get a single group by ID, not by name, so we have to fetch all the groups and filter like this.

controlpanel/frontend/consumers.py
```diff
@@ -0,0 +1,59 @@
...
+from controlpanel.api.models import User
+from controlpanel.api.tools import Tool
+
+
+log = logging.getLogger(__name__)
+
+
+class ToolConsumer(SyncConsumer):
+
+    def deploytool(self, message):
```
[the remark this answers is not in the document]
andyhd: Yeah, underscores in these consumer method names are significant - the type value of the received message is translated to get the method name, and `.`s are converted to `_`s. So `deploy_tool` means the message type needs to be `deploy.tool` - I suppose this is what we want anyway, so I'll change this.

controlpanel/frontend/static/app.scss
```diff
@@ -0,0 +1,37 @@
...
+@import "static/govuk-frontend/settings/compatibility";
+@import "static/govuk-frontend/settings/global-styles";
+@import "static/govuk-frontend/settings/ie8";
+@import "static/govuk-frontend/settings/media-queries";
+@import "static/govuk-frontend/settings/colours-palette";
+@import "static/govuk-frontend/settings/colours-organisations";
+@import "static/govuk-frontend/settings/colours-applied";
+@import "static/govuk-frontend/settings/spacing";
+@import "static/govuk-frontend/settings/measurements";
+// comment the line below out to avoid using Transport New font
```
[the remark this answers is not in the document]
andyhd: We should. It's a pity though, the Transport New font looks good.

controlpanel/frontend/static/javascripts/modules/get-access-logs.js
```diff
@@ -0,0 +1,73 @@
+//moj.Modules.accessLogs = {
```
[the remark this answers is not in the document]
andyhd: Yep

controlpanel/frontend/static/local/buttons.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  &:hover,
+  &:focus{
+    background-color: $cpanel-destructive-hover-colour;
+  }
+}
+
+
+//.button-secondary {
```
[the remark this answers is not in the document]
andyhd: Yep

controlpanel/frontend/static/local/layout.scss
```diff
@@ -0,0 +1,19 @@
+//.grid-row {
```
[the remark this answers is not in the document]
andyhd: Yep

controlpanel/oidc.py
```diff
@@ -0,0 +1,44 @@
...
+            email=claims.get(settings.OIDC_FIELD_EMAIL),
+            name=claims.get(settings.OIDC_FIELD_NAME),
+        )
+        user.save()
+        # user.aws_create_role()
```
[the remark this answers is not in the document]
andyhd: Oops, debugging

controlpanel/settings/common.py
```diff
@@ -0,0 +1,453 @@
...
+            ]
+        },
+    },
+]
+
+
+# -- Auth
+
+# List of classes used when attempting to authenticate a user
+AUTHENTICATION_BACKENDS = ["controlpanel.oidc.OIDCSubAuthenticationBackend"]
```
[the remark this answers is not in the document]
andyhd: I wasn't aware of other API clients, apart from Concourse. I'll add it back in.

controlpanel/api/auth0.py
xoen: Makes sense, I probably missed something when I had a look 👍

controlpanel/oidc.py
andyhd: I've tested this, and it works for me - it's possible your Github user doesn't have a `nickname` claim (the value of `OIDC_FIELD_USERNAME`), but it seems unlikely

controlpanel/kubeapi/views.py
andyhd: I've [changed the URLs for the k8s proxy](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/610/files#diff-0ad9842f86b8342a9266c007338496adR9). The base URL path for the `kubeapi` proxy app is set in `controlpanel/urls.py`:
```python
path("api/k8s/", include("controlpanel.kubeapi.urls"))
```
and the `kubeapi` URL config looks like
```python
path("<path:url>", views.KubeAPIProxy.as_view(), name="proxy")
```
which means anything after `/api/k8s/` in the URL is the proxied URL.

andyhd: Regarding whether we still need the K8s proxy - I'm not sure. I didn't want to remove it in this PR, in case it is being used - _but_ I changed the URLs, so that would probably break anything that is expecting `/k8s/` instead of `/api/k8s/`.

controlpanel/api/tools.py
andyhd: 👍

controlpanel/kubeapi/views.py
xoen: Right, so if I request `/api/k8s/foo` the request to k8s will be `/foo` 👍

xoen: Regarding the use of the proxy to k8s. I think we may have to use it for all requests to k8s from the UI.
The UI now makes requests using the Python client but that uses the CP service account.

Is this what we want? Before this PR the user was using the same JWT token (by design) to make requests to both API and k8s API which meant:
1. we know who performs what
2. if there is a bug in the API/UI and a user is able to perform an operation which they're not supposed to do the RBAC in the k8s API would stop it (user would get a `403`)

If UI uses Python client and CP service account these two things will not be true anymore.
The CP would perform things on behalf of the users but the CP has more permissions than normal users and this can potentially lead to problems.

CC: @r4vi

andyhd: Ah yes, I didn't mean to change that functionality. I will fix it.

Makefile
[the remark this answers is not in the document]
andyhd: We're not just transpiling, but also concatenating and compressing. But I guess we can do that without transpiling first.

controlpanel/kubeapi/oidc_patch.py
```diff
@@ -0,0 +1,121 @@
...
+        return
+
+    parts = provider["config"]["id-token"].split(".")
+
+    if len(parts) != 3:  # Not a valid JWT
+        return None
+
+    padding = (4 - len(parts[1]) % 4) * "="
+
+    jwt_attributes = json.loads(base64.b64decode(parts[1] + padding).decode("utf-8"))
```
r4vi: this should be:
```suggestion
    jwt_attributes = json.loads(base64.urlsafe_b64decode(parts[1] + padding).decode("utf-8"))
```

Makefile
```diff
@@ -1,60 +1,111 @@
 HOST=0.0.0.0
 PORT=8000
-PROJECT=control-panel
-MODULE=control_panel_api
+PROJECT=controlpanel
+MODULE=controlpanel
 VENV=venv
 BIN=${VENV}/bin
```
r4vi: to make this work better for people who want to opt out of this Makefile creating a virtual env, this works better
```suggestion
BIN=${VIRTUAL_ENV}/bin
```

andyhd:
Merging at last! 🎉

PR #629, opened 2019-04-05
Update helm to 2.13.0

s-block:
## What

Update helm to 2.13.0 as we are updating tiller

## How to review

1. Helm updated to 2.13.0


davidread:
@s-block what's the plan for upgrading?

Is helm 2.13.x compatible with tiller 2.9.1 (which we currently have on the cluster) or are you deploying in step? We also have helm in concourse to think of.

PR #658, opened 2019-04-30
Add airflow to helm tools

s-block:
## What

This PR adds the functionality to deploy a loc al version of the helm chart `airflow-sqlite`.

It is deployed in the users namespace and uses the `auth-proxy` for authentication.

## How to review

1. Go to tools
2. Click to deploy airflow
3. Click the link to see if airflow is accessible

https://trello.com/c/utqhVsoG


control_panel_api/tools.py
```diff
@@ -119,6 +120,32 @@ def deploy_params(self, user):
         ]
 
 
+class Airflow(BaseTool):
+    name = 'airflow-sqlite'
+
+    def release_name(self, username):
+        return f'{username}-{self.name}'
```
andyhd: This override does nothing - you can remove it

control_panel_api/tools.py
```diff
@@ -119,6 +120,32 @@ def deploy_params(self, user):
...
+
+    @property
+    def chart_name(self):
+        return f'mojanalytics/{self.name}'
```
andyhd: This is also overriding the base class unnecessarily

control_panel_api/tools.py
s-block: removed

control_panel_api/tools.py
s-block: removed

control_panel_api/settings/base.py
```diff
@@ -177,6 +177,12 @@ def is_enabled(value):
 JUPYTER_LAB_AUTH_CLIENT_ID = os.environ.get('JUPYTER_LAB_AUTH_CLIENT_ID')
 JUPYTER_LAB_AUTH_CLIENT_SECRET = os.environ.get('JUPYTER_LAB_AUTH_CLIENT_SECRET')
 
+AIRFLOW_AUTH_CLIENT_DOMAIN = os.environ.get('AIRFLOW_AUTH_CLIENT_DOMAIN', OIDC_DOMAIN)
```
xoen: nice 👍

PR #695, opened 2019-06-11
Use node-sass to compile SCSS and remove django-sass-processor

andyhd:
* django-sass-processor has bugs in the management command for precompiling SCSS files to CSS
* we don't need to compile SCSS on request/server run, as we deploy a Docker image, so django-sass-processor is not needed
* we already use Node.JS to compile JS files when building the Docker image, so we can use it to compile SCSS too

### DIFF
--- a/controlpanel/oidc.py
+++ b/controlpanel/oidc.py
@@ -13,15 +13,12 @@ class OIDCSubAuthenticationBackend(OIDCAuthenticationBackend):
     """
 
     def create_user(self, claims):
-        user = User.objects.create(
+        return User.objects.create(
             pk=claims.get("sub"),
             username=claims.get(settings.OIDC_FIELD_USERNAME),
             email=claims.get(settings.OIDC_FIELD_EMAIL),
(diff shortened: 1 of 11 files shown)


controlpanel/oidc.py
```diff
@@ -13,15 +13,12 @@ class OIDCSubAuthenticationBackend(OIDCAuthenticationBackend):
...
-        user = User.objects.create(
+        return User.objects.create(
             pk=claims.get("sub"),
             username=claims.get(settings.OIDC_FIELD_USERNAME),
             email=claims.get(settings.OIDC_FIELD_EMAIL),
             name=claims.get(settings.OIDC_FIELD_NAME),
         )
-        user.aws_create_role()
-        user.helm_create()
-        return user
```
andyhd: This was missed in the big PR - aws role creation and helm deployment is handled in `User.save()`

xoen: Is this still happening somewhere else? (e.g. in the view)

andyhd: Yes, in `User.save()`

PR #696, opened 2019-06-11
Bugfixes

andyhd:
* Fix 500 error on Analytical Tools page
* Fix user details page - was showing logged-in user's details regardless of request

### DIFF
--- a/controlpanel/api/tools.py
+++ b/controlpanel/api/tools.py
@@ -6,7 +6,7 @@ from django.utils.functional import cached_property
 from rest_framework.exceptions import APIException
 
 from controlpanel.api.helm import Helm
-from controlpanel.kubeapi.views import kubernetes
+from controlpanel.kubeapi.views import load_kube_config, kubernetes
 
 
 IDLED = "mojanalytics.xyz/idled"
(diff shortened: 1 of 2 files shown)


controlpanel/api/tools.py
```diff
@@ -112,6 +112,7 @@ def deploy_for(self, user):
         )
 
     def get_user_deployment(self, user):
+        load_kube_config()
```
xoen: Hang on a second, maybe I'm missing something here. 

**Who is making these requests to the k8s API?** I thought all requests through the CP would be proxied to the k8s API but using the actual user auth token and restricted to the actual user's permission (e.g. enforced by k8s). I only skimmed through `load_kube_config()` but it seems it's loading the config from the cluster, e.g. using the CP's role to perform these actions. This means that if we have a bug a normal user could potentially do things they're not supposed to do.

Maybe I'm missing something. Requests to k8s used to be made as the actual user in the previous version.

CC: @r4vi

andyhd: You're right - this isn't the correct solution and it's causing the problem that the user is actually unable to do something that they should.

PR #697, opened 2019-06-20
Fix k8s auth

andyhd:
The Control Panel accesses the Kubernetes API in 2 different ways

1. URLs starting with `/api/k8s` are proxied directly to the Kubernetes API, passing through an `Authorization` header containing a OIDC ID token JWT as a `Bearer` token

2. via the official Kubernetes Python client, which makes HTTP requests using credentials loaded from a local `.kube/config` or in-cluster metadata API.

We want to use the currently logged-in user's credentials (for tracking and permissions purposes), so we need to override the default client configuration to use the current user's OIDC ID token

The problems fixed are:
* The OIDC ID token was not stored during the OIDC authorization flow
* The scopes requested were not compatible with the Kubernetes API


controlpanel/api/tools.py
```diff
@@ -111,9 +111,9 @@ def deploy_for(self, user):
             *params,
         )
 
-    def get_user_deployment(self, user):
-        load_kube_config()
-        deployments = kubernetes.client.AppsV1Api().list_namespaced_deployment(
+    def get_user_deployment(self, request):
+        client = RequestUserKubernetesClient()
```
r4vi: :+1:

controlpanel/frontend/jinja2/tools.html
```diff
@@ -33,10 +33,13 @@ <h1 class="govuk-heading-xl">Your tools</h1>
             {% endfor %}
           {% endif %}
         {% else %}
-          {% if True or is_deploying[tool.name] %}
-          Deploying
+          {% if True %}
```
r4vi: won't this always be `True`?

controlpanel/api/tools.py
andyhd: ugh, need to change `request` arg back to `user`

controlpanel/frontend/jinja2/tools.html
andyhd: Yes, this is just temporary while I work on the functionality behind it

PR #698, opened 2019-06-24
Allow non-superuser app admins to manage app access

andyhd:
## What

Tweaked the frontend templates to allow app admins to:
  * Add and remove customers
  * See who other admins are but not access their user page
  * Add or remove read/write access for existing app data sources

Some bugfixes

https://trello.com/c/08PIr8SK


controlpanel/api/cluster.py
```diff
@@ -218,12 +222,12 @@ def save(self):
         )
 
     def _update_statements(self):
-        self["Statement"] = [
+        self["Statement"] = list(filter(None, [
```
xoen: When did this class started inheriting from dictionary and became all meta and functional 😱😱😱? What was wrong with it? AFAIK the IAM policy manipulation bit is always been working quite well, why risking introducing bugs and breaking it with unnecessary refactoring?

Isn't changing this class out of scope for this PR? Isn't changing the permission independent than changing the IAM policy manipulation?

If that's the case, I'd revert these changes and keep this PR focused on the change to the permissions required to allow non-admin users to manage their apps.

(I'm only now started looking at this so I may miss something fundamental tho!)

andyhd: It's not meta or functional - the code to handle policies has barely changed from here https://github.com/ministryofjustice/analytics-platform-control-panel/blob/b1aebfabeb1f13e2d763e1a7516b71b497af347b/control_panel_api/aws.py

andyhd: I can move this bugfix to a separate PR, but it is a tiny fix and didn't seem worth it

xoen: But why did that class had to change? Was it broken? What happened to "if it ain't broke don't fix it"?

It appears more complicated to me (what's `__enter__()` and `__exit__`, etc..?). I'm sorry. I just didn't spot this change before. If this is a bugfix, I guess leave it here 🤷‍♂.

controlpanel/frontend/jinja2/includes/app-list.html
```diff
@@ -31,12 +31,10 @@
         {{ yes_no(user_is_app_admin) }}
       </td>
       <td class="govuk-table__cell">
-      {% if user_is_app_admin or request.user.is_superuser %}
```
xoen: Does this mean that all users see the "Manage app" button for all apps?

controlpanel/api/cluster.py
andyhd: The refactor was part of the big PR - I guess you didn't notice it amongst all the other stuff. The changes are only to improve the API by putting the context manager functionality into the S3AccessPolicy class. For example:
```python
def _access_to_bucket_deleted(sender, **kwargs):
    access_to_bucket = kwargs['instance']

    with S3AccessPolicy.load(access_to_bucket.aws_role_name()) as policy:
        policy.revoke_access(access_to_bucket.s3bucket.arn)
```
versus the previous code
```python
@contextmanager
def s3_access_policy(role_name):
    policy_document = aws.get_inline_policy_document(
        role_name=role_name,
        policy_name=S3_POLICY_NAME,
    )
    policy = S3AccessPolicy(document=policy_document)

    yield policy

    aws.put_role_policy(
        role_name=role_name,
        policy_name=S3_POLICY_NAME,
        policy_document=policy.document,
    )


def revoke_bucket_access(bucket_arn, role_name):
    with s3_access_policy(role_name) as policy:
        policy.revoke_access(bucket_arn)


def _access_to_bucket_deleted(sender, **kwargs):
    access_to_bucket = kwargs['instance']

    services.revoke_bucket_access(
        access_to_bucket.s3bucket.arn,
        access_to_bucket.aws_role_name(),
    )
```

controlpanel/frontend/jinja2/includes/app-list.html
andyhd: Yes, they can click through to the app details page, but non-admin users will not be able to edit anything. I did this for consistency with other lists of resources.

controlpanel/api/cluster.py
xoen: Yeah, I completely missed this change.

Thanks for explaining the rationale around this change. I avoided adding that reading/wringing functionality to the `S3AccessPolicy` class by design as I wanted it to have one single responsability: manipulate the access policy.

I think we're missing another class which is responsible for the reading/writing of the policy into the role which uses the `S3AccessPolicy` class.

But as this has already changed I wouldn't change it further as part of this PR anyway :)

andyhd:
Yeah, the access controls at this point are purely in the frontend code. It's enough to satisfy the ticket, but I want to introduce proper access control both in the frontend and the API with something like Django-Guardian in a later change.

PR #699, opened 2019-06-26
Enable deploying a tool without push status updates

andyhd:
In the interest of expediting the deployment of the new Control Panel frontend to alpha, this drops any attempts to use SSE or websockets to update tool status in favour of simply executing the tool deployment in the background and returning immediately.

### DIFF
--- a/controlpanel/frontend/jinja2/tools.html
+++ b/controlpanel/frontend/jinja2/tools.html
@@ -7,13 +7,14 @@
 
 <h1 class="govuk-heading-xl">Your tools</h1>
 
-<p class="govuk-body">The status of your tools will update automatically.</p>
+{# <p class="govuk-body">The status of your tools will update automatically.</p>
+#}
 
 <table class="govuk-table">
   <thead class="govuk-table__head">
     <tr class="govuk-table__row">
       <th class="govuk-table__header">Name</th>
-      <th class="govuk-table__header">Status</th>
+      <th class="govuk-table__header">{# Status #}</th>
       <th class="govuk-table__header"><span class="govuk-visually-hidden">Actions</span></th>
     </tr>
   </thead>
@@ -23,6 +24,7 @@
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">{{ tool.name }}</td>
       <td class="govuk-table__cell">
+        {#
         <div class="sse-listener tool-status" data-tool-name="{{ tool.name }}">
         {% if deployed %}
           {% if tool.idled %}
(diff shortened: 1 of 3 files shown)


controlpanel/frontend/jinja2/tools.html
```diff
@@ -7,13 +7,14 @@
 
 <h1 class="govuk-heading-xl">Your tools</h1>
 
-<p class="govuk-body">The status of your tools will update automatically.</p>
+{# <p class="govuk-body">The status of your tools will update automatically.</p>
```
xoen: Sorry for the silly jinja question: What do the hashes (`#`) do? Comment the stuff between them?

controlpanel/frontend/jinja2/tools.html
```diff
@@ -23,6 +24,7 @@ <h1 class="govuk-heading-xl">Your tools</h1>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">{{ tool.name }}</td>
       <td class="govuk-table__cell">
+        {#
```
xoen: Do you also need to close this `#` block? 🤷‍♂

xoen: Uh, sorry ignore me: I missed the other `#` after the `</div>` 🤦‍♂ .

controlpanel/frontend/views/tool.py
```diff
@@ -51,21 +53,18 @@ class DeployTool(LoginRequiredMixin, RedirectView):
...
-            'test', 'toolStatusChange', {'toolName': name, 'status': 'Deploying...'},
-        )
+        try:
+            Tool.create(name).deploy_for(self.request.user)
 
-        messages.success(
-            self.request, f"Deploying {name}... this may take up to 5 minutes",
-        )
+        except ToolDeploymentError as error:
+            messages.error(f"Failed deploying {name}")
```
xoen: So this are flash messages, are they actually shown 😸? 🏆

xoen:
LGTM 👍

controlpanel/frontend/jinja2/tools.html
andyhd: Yep

controlpanel/frontend/views/tool.py
andyhd: Yep
Screenshot 2019-06-26 at 16 35 35

andyhd: Wrong example, but you get the idea

xoen: Awesome, I didn't remember we used this messages. Obviously there is something in the layout/template showing them 😆

PR #700, opened 2019-06-26
Show time since last login on user list page

andyhd:
https://trello.com/c/nDJX3Kpf

### DIFF
--- a/controlpanel/frontend/views/user.py
+++ b/controlpanel/frontend/views/user.py
@@ -1,3 +1,4 @@
+import dateutil.parser
 from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.urls import reverse_lazy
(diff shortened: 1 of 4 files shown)


controlpanel/frontend/views/user.py
```diff
@@ -6,12 +7,37 @@
...
+
+            if auth0_id == '':
+                continue
+
+            try:
+                last_login = dateutil.parser.parse(user.get('last_login'))
+            except (TypeError, ValueError):
+                last_login = None
+
+            last_logins.append((auth0_id, last_login))
```
xoen: Python noob question: What's the benefit of appending tuples compared to just have `last_logins` as a dictionary, e.g.

```python
last_logins = {}

for user in users:
   # ...
   last_logins[auth0_id] = last_login
```

or even doing that directly on `context['last_login']`?

andyhd: There's no difference really. I suppose it might be faster to avoid those dictionary lookups when assigning, but I don't know if `dict(list_of_tuples)` will do that anyway. I started with a dict-comprehension, but had to catch the exceptions.

PR #701, opened 2019-07-02
Add nginx reverse proxy for deployment in production

andyhd:
⚠️ **REWRITTEN**

* Running Django with `DEBUG=False` means static files are not served, so we need to come up with alternative arrangements
* I've added a configuration for a simple Nginx reverse proxy which sits in front of the Django app and serves static files

**AFTER REWRITE**
* I've used the official nginx docker image and mount the configuration file and static assets as volumes - see the [Helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/356)

### DIFF
--- a/controlpanel/oidc.py
+++ b/controlpanel/oidc.py
@@ -16,7 +16,7 @@ class OIDCSubAuthenticationBackend(OIDCAuthenticationBackend):
         return User.objects.create(
             pk=claims.get("sub"),
             username=claims.get(settings.OIDC_FIELD_USERNAME),
-            email=claims.get(settings.OIDC_FIELD_EMAIL),
+            email=claims.get(settings.OIDC_FIELD_EMAIL, ''),
             name=claims.get(settings.OIDC_FIELD_NAME),
         )
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,13 +34,13 @@ jobs:
       - run:
           name: Build application Docker image
           command: |
-            docker build --cache-from=app -t app .
+            docker build --cache-from=controlpanel -t controlpanel .
 
       - run:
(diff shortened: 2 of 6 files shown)


controlpanel/oidc.py
```diff
@@ -16,7 +16,7 @@ def create_user(self, claims):
         return User.objects.create(
             pk=claims.get("sub"),
             username=claims.get(settings.OIDC_FIELD_USERNAME),
-            email=claims.get(settings.OIDC_FIELD_EMAIL),
+            email=claims.get(settings.OIDC_FIELD_EMAIL, ''),
```
andyhd: The email address should always be present, but it is not very important to us, so just silently handle a missing value.

docker-compose.prod.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - db
+    links:
+      - db
+    env_file: .env
+    environment:
+      DB_HOST: "db"
+      DB_NAME: "controlpanel"
+      DB_USER: "controlpanel"
+      DEBUG: "False"
+      ALLOWED_HOSTS: "*"
```
andyhd: This is just for testing locally

docker/nginx.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM nginx:stable-alpine AS base
+
+LABEL maintainer="andy.driver@digital.justice.gov.uk"
+
+FROM controlpanel:latest AS cpanel
```
andyhd: This probably isn't the best way to get the static files into the Docker image.

r4vi: `:latest` is the :fish:y bit

andyhd:
Not sure waitress will work with ASGI?

PR #705, opened 2019-07-05
Add parameter store

s-block:
Tests coming ...

Adds the ability to add/delete parameters that can be access in webapps and airflow tasks

## What

Added Parameter model. The secrets are not stored in this model but the name is stored so it can be accessed and deleted from the control panel

## How to review

1. Go to the Parameters tab on the main nav
2. Click on create a Parameter link
3. fill out form to create a parameter
4. view the Parameter on the list
5. Delete the parameter
6. Thew creation/deletion of the parameters can be viewed in AWS console - https://eu-west-1.console.aws.amazon.com/systems-manager/parameters

Provide [http://example.com](links) to relevant tickets, articles or other
resources.

https://eu-west-1.console.aws.amazon.com/systems-manager/parameters


controlpanel/frontend/jinja2/parameter-create.html
```diff
@@ -0,0 +1,108 @@
...
+      }}
+      {{
+        govukInput({
+          "label": {
+            "text": "Value",
+            "classes": "govuk-label--m",
+          },
+          "classes": "govuk-!-width-two-thirds",
+          "hint": {
+            "text": 'String. Any characters'
```
andyhd: Not sure this hint is very friendly

controlpanel/frontend/jinja2/parameter-create.html
```diff
@@ -0,0 +1,108 @@
...
+          "value": form.value.value(),
+          "errorMessage": { "html": form.value.errors|join("<br>") } if form.value.errors else {}
+        })
+      }}
+      {{
+        govukInput({
+          "label": {
+            "text": "Role name",
+            "classes": "govuk-label--m",
+          },
```
andyhd: I don't think you should make this a user-completed form field. The form should just be `key` and `value`.

controlpanel/frontend/jinja2/parameter-create.html
```diff
@@ -0,0 +1,108 @@
...
+            "maxlength": "60",
+          },
+          "value": form.role_name.value(),
+          "errorMessage": { "html": form.role_name.errors|join("<br>") } if form.role_name.errors else {}
+        })
+      }}
+      {{
+        govukRadios({
+          "label": {
+            "text": "Airflow or Web app",
```
andyhd: This field is bad UX. Better to have the form on the app page and populate this field automatically.

controlpanel/frontend/jinja2/parameter-create.html
```diff
@@ -0,0 +1,108 @@
...
+          "attributes": {},
+          "items": [
+            {
+              "value": "airflow",
+              "text": "Airflow",
+            },
+            {
+              "value": "webapp",
+              "text": "Web app",
+            }
```
andyhd: Hardcoded choices again. These should at least be passed as template context variables.

controlpanel/frontend/jinja2/parameter-list.html
```diff
@@ -0,0 +1,23 @@
...
+{% endset %}
+
+{% block content %}
+  <h1 class="govuk-heading-xl">{{ page_title }}</h1>
+
+  {{ parameter_list(object_list) }}
+
+  <p class="govuk-body">
+  <a class="govuk-button" href="{{ url('create-parameter') }}">
+      Create new parameter
```
andyhd: Who has permission to create parameters? You should hide this button from users who don't

controlpanel/api/models/parameter.py
```diff
@@ -0,0 +1,74 @@
...
+    key = models.CharField(
+        max_length=50,
+        validators=[RegexValidator(r'[a-zA-Z0-9_]{1,50}')]
+    )
+    description = models.CharField(max_length=600)
+    used_by = models.CharField(
+        max_length=8,
+        choices=(
+            ('airflow', 'Airflow'),
+            ('webapp', 'Web app'),
```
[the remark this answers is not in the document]
andyhd: What is this field for?

controlpanel/api/filters.py
```diff
@@ -62,3 +62,20 @@ def filter_queryset(self, request, queryset, view):
...
+    - Normal users see only Parameters they created
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        queryset = super().filter_queryset(request, queryset, view)
+
+        if is_superuser(request.user):
+            return queryset
+
+        return queryset.filter(created_by=request.user)
```
xoen: Nice 👍

controlpanel/api/models/parameter.py
```diff
@@ -0,0 +1,74 @@
...
+        null=True,
+    )
+
+
+@receiver(pre_save, sender=Parameter)
+def save_aws_parameter(sender, **kwargs):
+    parameter = kwargs['instance']
+    is_create = not parameter.pk
+
+    if is_create:
```
xoen: What about the update?

controlpanel/api/permissions.py
```diff
@@ -114,3 +114,26 @@ def has_object_permission(self, request, view, obj):
 class ToolDeploymentPermissions(BasePermission):
     def has_permission(self, request, view):
         return not request.user.is_anonymous
+
+
+class ParameterPermissions(BasePermission):
+    """
+    Superusers can do anything.
+    Normal users can list all, retrieve/update only themselves.
```
xoen: I know it's an artifact of a Copy&Paste but probably worth changing this to

> Normal users can list all, retrieve/update only they parameters.

to avoid any possible confusion.

controlpanel/frontend/forms.py
```diff
@@ -125,3 +125,16 @@ def __init__(self, *args, **kwargs):
...
+    key = forms.CharField(
+        validators=[RegexValidator(r'[a-zA-Z0-9_]{1,50}')]
+    )
+    role_name = forms.CharField(
+        validators=[RegexValidator(r'[a-zA-Z0-9_-]{1,60}')]
+    )
+    value = forms.CharField(widget=forms.PasswordInput)
+    used_by = forms.ChoiceField(choices=[
+        ('airflow', 'Airflow'),
+        ('webapp', 'Web app'),
```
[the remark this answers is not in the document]
xoen: Or maybe to begin with in a constant in the `Parameter` model.

I'd probably maintain the migration as it is even if it's a duplication as they shouldn't be coupled to the models as that may risk breaking them as the code changes.

Regarding this field, does it mean that a parameter can only be used a single thing only? Worth asking ourselves this question.

Pro of single parameter client:
- simpler
- peace of mind when deleting a parameter (you don't risk to accidentally break a client)

Cons:
- users may duplicate a parameter to use it for several things and updates may result in inconsistent values

controlpanel/api/permissions.py
xoen: Also, can all users see all Parameters really? If I remember correctly [the filtering](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/705/files#diff-8bd403b7253dd1da943b0ca74d79d7d8R67) is done before the permissions are checked.

That should mean that users will only see the params they created (which sounds sensible for normal users). If that's the case this comment should really be:

> Normal users can list/retrieve/update only parameters they created.

(or something like that)

If that's the case it means that the UI don't need to do anything fancy to show/hide buttons to delete parameters as the list will only include parameters they can mess with (but worth double checking).

controlpanel/frontend/jinja2/base.html
```diff
@@ -80,6 +80,11 @@
         "text": "Webapps",
         "href": url("list-apps"),
         "active": page_name == "webapps"
+      },
+      {
+        "text": "Parameters",
+        "href": url("list-parameters"),
+        "active": page_name == "parameters"
```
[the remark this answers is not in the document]
xoen: Sounds like a good point. What about airflow ones? (genuine question, not trolling)

controlpanel/frontend/jinja2/includes/parameter-list.html
```diff
@@ -0,0 +1,44 @@
...
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+  {%- for parameter in parameters %}
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell">{{ parameter.key }}</td>
+      <td class="govuk-table__cell">{{ parameter.arn }}</td>
+      <td class="govuk-table__cell">{{ parameter.role_name }}</td>
+      <td class="govuk-table__cell">
+        <form method="POST" action="{{ url("delete-parameter", kwargs={ "pk": parameter.id }) }}">
```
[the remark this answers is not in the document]
xoen: Good point, but isn't [the filtering](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/705/files#diff-8bd403b7253dd1da943b0ca74d79d7d8R67) only returning params users created? If I understand/remember correctly this should mean users will only see parameters they can mess.

controlpanel/api/permissions.py
andyhd: This is why comments are rarely useful

controlpanel/frontend/jinja2/parameter-create.html
xoen: Maybe something like "Value for the parameter. All characters allowed"? (or something like that)

controlpanel/frontend/jinja2/parameter-create.html
```diff
@@ -0,0 +1,108 @@
...
+    {% call govukFieldset({ "legend": { "html": legend|safe } }) %}
+      {{
+        govukInput({
+          "label": {
+            "text": "Key",
+            "classes": "govuk-label--m",
+          },
+          "classes": "govuk-!-width-two-thirds",
+          "hint": {
+            "text": '60 chars max, only letters, numbers and underscores'
```
xoen: Do users know what a parameter key is? Don't know if the label should be more explicit (maybe "Key" is good), or maybe explain in the hit that this is the parameter identifier?

The thing is, as far as I understand this the key is only appended to the rest of the name. Isn't the name or ARN the thing they're using to actually access the value?

controlpanel/frontend/jinja2/parameter-create.html
xoen: How can the app know where the param is going to be used tho? At least it's a radio button so it shouldn't allow them to enter weird stuff here.

controlpanel/frontend/jinja2/parameter-list.html
xoen: As far as I can tell all users can create params, is that right?

controlpanel/frontend/jinja2/includes/parameter-list.html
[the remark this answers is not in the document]
andyhd: You linked to the wrong filter (https://github.com/ministryofjustice/analytics-platform-control-panel/pull/705/files#diff-ccfeac00f568c4f2f555f7f0aa3e260bR20), but good point - the user can only see Parameters they created 👍

controlpanel/frontend/jinja2/parameter-create.html
andyhd: If there's no restriction, I think you can omit the "Any characters" or "All characters allowed".

controlpanel/api/permissions.py
andyhd: Also, I think if you are setting Parameters to be used as environment variables for a webapp, then the app admins should all be able to see and edit the list of parameters.

controlpanel/frontend/views/parameter.py
```diff
@@ -0,0 +1,63 @@
...
+from controlpanel.frontend.forms import CreateParameterForm
+
+
+class ParameterList(LoginRequiredMixin, ListView):
+    model = Parameter
+    template_name = "parameter-list.html"
+
+    def get_queryset(self):
+        queryset = Parameter.objects.all()
+        if is_superuser(self.request.user):
```
xoen: Hang on a second, isn't this logic already in the filter? Shouldn't be enough to tell this view to use that filter and the queryset would already be filter according to that? This duplication can lead to confusion/bugs.

controlpanel/frontend/jinja2/parameter-create.html
andyhd: This field (if it is actually necessary), should be populated automatically depending on whether the user is creating a Parameter from a webapp page or a (future) Airflow role/dag page.

controlpanel/frontend/views/parameter.py
andyhd: The filters you are referring to are for the REST API. I agree that the duplication should be avoided, but I think you'll have to do it via a model manager.

xoen: I'm so confused. What's the difference between the views in the API and the views in the frontend? I know it must sound like a very stupid question...

xoen: Right, so the views in `api` are gone and moved to `frontend`. Fair enough...but what happened to the filtering @andyhd?

[They used to be used](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/85273c703da7caa0ed348aa1d4c8b2c96d8fef37/control_panel_api/views.py#L239)...right, looks like some of the views are now in `controlpanel/api/views/models.py`...that's why I couldn't find them.

Sorry, got confused by the new structure.

xoen: Mmmm 🤔 but if this is reading from the DB, can't it just say "use the filter" and django would filter the records automatically according to the filter class? I haven't touched this in a long time so maybe I'm missing something obvious here.

andyhd: The old API views are still in place, just separated out into smaller files. They handle REST API URLs like `/api/cpanel/v1/apps/`, the new frontend views are for the web interface at URLs like `/webapps/`. It should be possible to share the filtering code between them, though.

andyhd: The filters you are talking about are Django REST Framework filters and are not designed to be used outside of the framework. But the `filter_queryset` method that they have is just Django ORM code, so it should be possible to share that.

andyhd: The DRF provides its own `View` classes which use the filters. But this frontend view is a different type of `View` class and doesn't have all the setup to use those filters. But, as I said, it shouldn't be too difficult to factor out the common code and share it.

xoen: @andyhd I see what you mean.

That's interesting...aren't the views in the `frontend` getting the data using the models/DB directly tho? If that's the case, is the `api` even used? And if not, are the permissions even used anymore? Does it mean that the user can just URL browse and see/do whatever they want? 😱😱

For example, this is the [view to delete a user](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/controlpanel/frontend/views/user.py#L43) it says to use the `User` model but where are [the permission](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/controlpanel/api/permissions.py#L61) enforced 🙀?

andyhd: That is an excellent point. I'll fix the security in the existing views now.

controlpanel/api/models/parameter.py
```diff
@@ -0,0 +1,74 @@
...
+    def value(self, val):
+        self._value = val
+
+    @property
+    def arn(self):
+        return f"arn:aws:ssm:::parameter{self.name}"
+
+    @property
+    def name(self):
+        return f"/alpha/{self.used_by}/{self.role_name}/secrets/{self.key}"
```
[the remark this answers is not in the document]
s-block: yep, sorted that

controlpanel/api/models/parameter.py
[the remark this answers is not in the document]
s-block: Parameters can be used in either airflow or webapps. Main reason is that in airflow tasks the developer can give themselves access to any parameter resource via the iam_policy.json file. I have got an explicit deny on all of the airflow policies for any parameter not for airflow and not for the specific role.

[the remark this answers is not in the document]
s-block: Are there any other places parameters are needed?

controlpanel/api/models/parameter.py
```diff
@@ -0,0 +1,74 @@
...
+    def value(self):
+        return self._value
+
+    @value.setter
+    def value(self, val):
+        self._value = val
+
+    @property
+    def arn(self):
+        return f"arn:aws:ssm:::parameter{self.name}"
```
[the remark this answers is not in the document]
s-block: Yeah, used this now

controlpanel/api/models/parameter.py
```diff
@@ -0,0 +1,74 @@
...
+        validators=[RegexValidator(r'[a-zA-Z0-9_]{1,63}')]
+    )
+    created_by = models.ForeignKey(
+        "User",
+        on_delete=models.SET_NULL,
+        null=True,
+    )
+
+
+@receiver(pre_save, sender=Parameter)
```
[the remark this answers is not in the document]
s-block: I had done this for save but overriding delete only works when the `object.delete()` is called. Also, I think it is pretty clear when the signals are in the sam e file. Happy to change this though.

controlpanel/frontend/jinja2/parameter-create.html
s-block: Removed

controlpanel/frontend/jinja2/parameter-create.html
s-block: For now we still need to know if it is airflow/webapp for restricting access to airflow `iam_policy.json` files. But yes, I agree that if/when we have an airflow page we should set it automatically.

controlpanel/frontend/jinja2/parameter-list.html
s-block: Anyone can create a Parameter. Users can only see their own.

controlpanel/api/models/parameter.py
s-block: At the moment there is no update. just create/delete.

controlpanel/api/permissions.py
s-block: Yeah, maybe something to do with Django Guardian

controlpanel/frontend/views/parameter.py
s-block: This restricts users from doing any action on parameters they haven't created - including delete - as the queryset only contains their own parameters. Unless you are a superuser of course.

controlpanel/frontend/jinja2/parameter-create.html
s-block: Yeah, until there is a way of listing airflow roles then it will need to be selected. I have made it an autocomplete field for now

controlpanel/frontend/jinja2/parameter-create.html
s-block: Have added to template context

controlpanel/frontend/forms.py
```diff
@@ -125,3 +125,15 @@ def __init__(self, *args, **kwargs):
...
+class CreateParameterForm(forms.Form):
+    key = forms.CharField(
+        validators=[RegexValidator(r'[a-zA-Z0-9_]{1,50}')]
+    )
+    role_name = forms.CharField(
+        validators=[RegexValidator(r'[a-zA-Z0-9_-]{1,60}')]
+    )
+    value = forms.CharField(widget=forms.PasswordInput)
+    app_type = forms.ChoiceField(
+        choices=Parameter._meta.get_field('app_type').choices
```
andyhd: Is this the recommended way to get the field choices? Accessing a "private" property seems wrong.

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
+moj.Modules.roleNames = {
+  id: "role_name",
+  selectId: "role_name-select",
+  formClass: "appRoles",
+  roleListEndpoint: "/parameters/form/role-list.js",
+  roles: null,
+
+  init() {
+    if (document.querySelectorAll("." + this.formClass).length) {
```
andyhd: If you just set `formClass: ".appRoles"`, you don't need to do this ugly string concatenation.

controlpanel/frontend/views/parameter.py
```diff
@@ -0,0 +1,86 @@
...
+
+class ParameterFormRoleList(LoginRequiredMixin, View):
+
+    def get(self, *args, **kwargs):
+        roles = aws.list_role_names()
+        data = {
+            "airflow": [r for r in roles if r.startswith(f"airflow")],
+            "webapp": [r for r in roles if r.startswith(f"{settings.ENV}_app")]
+        }
+        return JsonResponse(data)
```
andyhd: This view should be a DRF view in the `api` module, at a URL like `/api/cpanel/v1/parameters/`

controlpanel/frontend/jinja2/parameter-list.html
```diff
@@ -0,0 +1,23 @@
+{% from "modal-dialog/macro.html" import modal_dialog %}
```
andyhd: This macro is unused

controlpanel/frontend/jinja2/parameter-list.html
```diff
@@ -0,0 +1,23 @@
...
+{% from "includes/parameter-list.html" import parameter_list with context %}
+
+{% extends "base.html" %}
+
+{% set page_name = "parameters" %}
+{% set page_title = "Your parameters" %}
+
+{% set access_levels_html %}
+{% include "modals/user_data_access_levels.html" %}
+{% endset %}
```
andyhd: This variable is unused

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+      this.getRoles().done(() => {
+        this.loadRolesToSelect();
+        this.bindEvents();
+        console.log(this.roles);
```
andyhd: Remove this `console.log`

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+    return $.get(
+      this.roleListEndpoint,
+      (data) => {
+        this.roles = data;
+      }
+    )
+  },
+
+  loadRolesToSelect() {
+    $("#" + this.selectId).find('option:not([value=""])').remove();
```
andyhd: If you set `selectId` to `#role_name-select`, you don't need to do this concatenation.

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+    if (self.roles !== null) {
+      const appType = $('input[name=app_type]:checked').val();
+      if (appType) {
+        roles = this.roles[appType];
+      } else {
+        roles = this.roles['airflow'].concat(this.roles['webapp']);
+      }
+    }
+    $('.autocomplete__wrapper').remove();
+    $('#role_name-select').attr('id', 'role_name');
```
andyhd: If you assign `$(this.selectId)` to a variable (eg `selectField`), you can avoid this second DOM search. Also, you hard-coded `#role_name-select` here.

andyhd: Why are you modifying the `id` of the element?

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
+moj.Modules.roleNames = {
+  id: "role_name",
+  selectId: "role_name-select",
+  formClass: "appRoles",
+  roleListEndpoint: "/parameters/form/role-list.js",
```
andyhd: I dont' think you should hard-code this URL path

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+      (data) => {
+        this.roles = data;
+      }
+    )
+  },
+
+  loadRolesToSelect() {
+    $("#" + this.selectId).find('option:not([value=""])').remove();
+    let roles = [];
+    if (self.roles !== null) {
```
andyhd: `self.roles` should be `this.roles`?

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+
+  init() {
+    if (document.querySelectorAll("." + this.formClass).length) {
+      this.getRoles().done(() => {
+        this.loadRolesToSelect();
+        this.bindEvents();
+        console.log(this.roles);
+      });
+    }
+    this.$repoSlugTypeahead = $('#' + this.id)
```
andyhd: Avoid using JQuery unnecessarily. This could be `this.repoSlugTypeahead = document.getElementById(this.id);`

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+  },
+
+  bindEvents() {
+    $('input[name=app_type]').change(() => {
```
andyhd: You don't need JQuery for this either
```javascript
document.querySelector('input[name=app_type]').forEach(input => {
  input.addEventListener('change', (event) => {
    this.loadRolesToSelect();
  });
});
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+  loadRolesToSelect() {
+    $("#" + this.selectId).find('option:not([value=""])').remove();
+    let roles = [];
+    if (self.roles !== null) {
+      const appType = $('input[name=app_type]:checked').val();
+      if (appType) {
+        roles = this.roles[appType];
+      } else {
+        roles = this.roles['airflow'].concat(this.roles['webapp']);
+      }
```
andyhd: We are transpiling ES6, so if you changed the `parameters-list-roles` view to return a simple list instead of the object with two lists, you could do
```javascript
// this.roles = ['airflow-role-1', 'webapp-role-1', 'airflow-role-2', ...]
let roles = this.roles || [];
if (roles && appType) {
  roles = this.roles.filter(name => name.startsWith(appType));
}
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+    $('.autocomplete__wrapper').remove();
```
andyhd: Why remove this?

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,53 @@
...
+    $('#role_name-select').attr('id', 'role_name');
+    accessibleAutocomplete.enhanceSelectElement({
+      id: this.id,
+      selectElement: document.querySelector('#role_name'),
```
andyhd: Hard-coded query selector

requirements.lock
```diff
@@ -1,4 +1,4 @@
-adal==1.2.1
+adal==1.2.2
```
andyhd: Since you haven't updated `requirements.txt`, I'd remove this file from the PR - we can update dependencies in a separate PR.

andyhd: Same with `package-lock.json`

controlpanel/frontend/views/parameter.py
```diff
@@ -0,0 +1,63 @@
...
+    def get_form_kwargs(self):
+        return FormMixin.get_form_kwargs(self)
+
+    def get_success_url(self):
+        return reverse_lazy("list-parameters")
+
+    def form_valid(self, form):
+        self.object = Parameter(
+            key=form.cleaned_data['key'],
+            role_name=form.cleaned_data['role_name'],
```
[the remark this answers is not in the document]
s-block: Yeah that's right. I guess that isn't a problem as long as a user can't delete params another user has created

controlpanel/frontend/forms.py
s-block: I have set as a constant in the models file. That is fairly standard

controlpanel/frontend/static/javascripts/modules/role-names.js
s-block: Have to filter the roles in the api because there are a lot of roles that are not related to this. Also webapp roles are prefixed with `alpha_app` - does the frontend js have access to the environment (dev|alpha)?

s-block: Yep, had to use regex as webapp start with `(dev|alpha)_app_`

controlpanel/frontend/static/javascripts/modules/role-names.js
s-block: Yeah, autocomplete needs to be reloaded so this can be removed which refreshes the source

controlpanel/frontend/static/javascripts/modules/role-names.js
s-block: sorted

controlpanel/frontend/static/javascripts/modules/role-names.js
s-block: changed

controlpanel/frontend/static/javascripts/modules/role-names.js
s-block: So only airflow roles show when airflow is selected

controlpanel/api/models/parameter.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__(*args, **kwargs)
+        self._value = None
+
+    @property
+    def value(self):
+        return self._value
+
+    @value.setter
+    def value(self, val):
+        self._value = val
```
andyhd: Why is `value` a `@property`?

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,56 @@
...
+  formClass: ".appRoles",
+  autocompleteWrapperClass: ".autocomplete__wrapper",
+  roleFilters: {
+    airflow: new RegExp('^airflow_'),
+    webapp: new RegExp('^(alpha|dev)_app_'),
+  },
+  roles: null,
+
+  init() {
+    this.$selectField = $(this.selectId);
```
andyhd: This should be
```javascript
this.selectField = document.getElementById(this.selectId);
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,56 @@
...
+    this.roleListEndpoint = this.$selectField.attr(this.roleEndpointAttr);
```
andyhd: This should be
```javascript
this.roleListEndpoint = this.selectField.dataset.roleEndpoint;
```
See [Using data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes)

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,56 @@
...
+    if (document.querySelectorAll(this.formClass).length) {
```
andyhd: This can use `document.querySelector` instead, as you only have one selector

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,56 @@
...
+      this.getRoles().done(() => {
+        this.loadRolesToSelect();
+        this.bindEvents();
+      });
+    }
+  },
+
+  bindEvents() {
+    document.querySelectorAll('input[name=app_type]').forEach(input => {
```
andyhd: This can also be `document.querySelector`

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,56 @@
...
+    return $.get(
+      this.roleListEndpoint,
+      (data) => {
+        this.roles = data;
+      }
+    )
+  },
+
+  loadRolesToSelect() {
+    this.$selectField.find('option:not([value=""])').remove();
```
andyhd: This should be
```javascript
this.selectField.querySelectorAll('option:not([value=""])').forEach(option => {
  this.selectField.removeChild(option);
});
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,61 @@
...
+  formClass: ".appRoles",
+  autocompleteWrapperClass: ".autocomplete__wrapper",
+  roleFilters: {
+    airflow: new RegExp('^airflow_'),
+    webapp: new RegExp('^(alpha|dev)_app_'),
+  },
+  roles: null,
+
+  init() {
+    this.$selectField = $(this.selectSelector);
```
andyhd: This should be `this.selectField = document.querySelector(this.selectSelector)`

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,61 @@
...
+    return $.get(
+      this.roleListEndpoint,
+      (data) => {
+        this.roles = data;
+      }
+    )
+  },
+
+  loadRolesToSelect() {
+    const appType = $('input[name=app_type]:checked').val();
```
andyhd: This should be
```javascript
const appType = document.querySelector('input[name=app_type]:checked').value;
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,61 @@
...
+    let roles = this.roles || [];
+    if (roles && appType) {
+      roles = this.roles.filter(name => this.roleFilters[appType].test(name));
+    }
+    this.$selectField.find('option:not([value=""])').remove();
```
andyhd: This should be
```javascript
this.selectField.querySelectorAll('option:not([value=""])').forEach(option => {
  this.selectField.removeChild(option);
});
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,61 @@
...
+    this.$selectField.attr('id', this.id);
+    $(this.autocompleteWrapperClass).remove();
```
andyhd: This should be
```javascript
document.querySelector(this.autocompleteWrapperClass).remove();
```

controlpanel/frontend/static/javascripts/modules/role-names.js
```diff
@@ -0,0 +1,61 @@
...
+  bindEvents() {
+    document.querySelectorAll('input[name=app_type]').forEach(input => {
+      input.addEventListener('change', (event) => {
+        this.loadRolesToSelect();
+      });
+    });
+  },
+
+  getRoles() {
+    return $.get(
```
andyhd: You could get rid of JQuery altogether by replacing this with [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Examples), but it's not urgent.

s-block:
Ahhhhh. OK. Yeah, I have removed the jQuery dependency

controlpanel/api/models/parameter.py
s-block: We don't want to save it. Also we want to be able to show and set it in the API then send the value to AWS on save.

andyhd: Yeah, but the property doesn't do anything - you could just have
```python
def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self.value = None
```
couldn't you?

PR #706, opened 2019-07-09
Access Github API with user's access token

andyhd:
`cpfrontend` accesses the Github API directly, and uses the access token from the user's Auth0 identity
Translated this into Python


andyhd:
This should fix https://github.com/ministryofjustice/analytics-platform/issues/103

PR #708, opened 2019-07-11
Add missing serializers for Tools and ToolDeployments

andyhd:
Accessing `/api/cpanel/v1/deployments/` [raises an exception](https://sentry.service.dsd.io/mojds/control-panel-api/issues/36532/)

This also restores the output of `/api/cpanel/v1/tools/` to the format that the old frontend expects.


andyhd:
They are already hooked up, but I never tested them

PR #715, opened 2019-08-01
Unify API and frontend permissions model

andyhd:
* Added Django Rules to create object permissions for actions on models, eg: User X has permission to update App Y, rather than User X has permission to update Apps.
* Refactored API Django REST Framework permissions classes to check object permissions
* Added object permissions checks in frontend views
* Updated tests and added first frontend tests!


controlpanel/api/cluster.py
```diff
@@ -329,3 +330,20 @@ def delete_parameter(name):
...
             raise e
 
+
+def get_repositories(user):
+    repos = []
+    github = Github(user.github_api_token)
+    for name in settings.GITHUB_ORGS:
+        org = github.get_organization(name)
+        repos.extend(org.get_repos())
+    return list(repos)
```
r4vi: isn't repos already a `list`?

controlpanel/api/rules.py
```diff
@@ -0,0 +1,198 @@
...
+    :param user User: The user to check
+    :param obj App|AppS3Bucket|UserApp: The object to check
+    """
+    if obj is None:
+        return True
+
+    if isinstance(obj, App):
+        try:
+            UserApp.objects.get(user=user, app=obj, is_admin=True)
+        except UserApp.DoesNotExist:
```
r4vi: `user in obj.admins` might be faster/easier to read?

controlpanel/api/migrations/0008_remove_unused_role_team_models.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 2.2.3 on 2019-07-18 11:14
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
r4vi: shouldn't there be a migration to remove `Group` too?

r4vi:
I would still like @s-block so :eyes: this because he's used `rules` before

controlpanel/frontend/views/user.py
```diff
@@ -6,20 +6,26 @@
...
-    template_name = "user-list.html"
+    permission_required = 'api.list_user'
     queryset = User.objects.exclude(auth0_id='')
+    template_name = "user-list.html"
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        users = auth0.ManagementAPI().list_users()
+        users = auth0.ManagementAPI().list_users(
```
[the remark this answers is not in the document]
andyhd: That and `email` users - we're only interested in users who can login to Control Panel

controlpanel/frontend/views/app.py
```diff
@@ -31,38 +32,41 @@
...
+    permission_required = 'api.list_app'
     template_name = "webapp-list.html"
-    all_apps = False
+
+    def get_queryset(self):
+        qs = App.objects.all().prefetch_related('userapps')
+        return qs.filter(userapps__user=self.request.user)
+
+
+class AdminAppList(AppList):
```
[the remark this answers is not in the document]
andyhd: Not sure I follow you - isn't that what I'm doing?

controlpanel/api/cluster.py
andyhd: Oops, a leftover from when I was using `set`s to get rid of dupes. There can't be any anyway.

controlpanel/frontend/views/app.py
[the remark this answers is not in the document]
r4vi: yes ignore me I didn't read that AdminAppList was inheriting from AppList

controlpanel/api/migrations/0008_remove_unused_role_team_models.py
andyhd: I'm not sure about messing with models from `contrib.auth` - and I think we may want to use them in the future.

r4vi: oh right, I thought it was one of our models.

controlpanel/frontend/views/parameter.py
```diff
@@ -14,20 +15,26 @@
...
     model = Parameter
+    permission_required = 'api.list_parameter'
     template_name = "parameter-list.html"
 
     def get_queryset(self):
-        queryset = Parameter.objects.all()
-        if is_superuser(self.request.user):
-            return queryset
-        return queryset.filter(created_by=self.request.user)
+        return Parameter.objects.filter(created_by=self.request.user)
```
s-block: Makes sense to separate views :+1:

controlpanel/api/rules.py
```diff
@@ -0,0 +1,194 @@
+"""
+Rules and permissions (and object permissions) using django-rules
+"""
+
+from rules import add_perm, is_active, is_authenticated, is_superuser, predicate
```
xoen: Probably a stupid question, where are these funtions defined?

controlpanel/api/rules.py
```diff
@@ -0,0 +1,194 @@
...
+
+    if isinstance(obj, UserApp):
+        return obj.is_admin
+
+    # XXX raise exception?
+    return False
+
+
+add_perm('api.list_app', is_authenticated)
+add_perm('api.create_app', is_authenticated & is_superuser)
```
xoen: I imagine the permissions records associated with a superuser are deleted or updated if for whatever reason he stops being a superuser.

controlpanel/api/rules.py
```diff
@@ -0,0 +1,194 @@
...
+
+    :param user User: The user to check
+    :param obj AppS3Bucket|S3Bucket|UserS3Bucket: The object to check
+    :param kwargs dict: Additional search criteria
+    """
+    if obj is None:
+        return True
+
+    if isinstance(obj, S3Bucket):
+        try:
```
xoen: You could change to not use the exceptions to something like:

```python
# not tested/tried!
return UserS3Bucket.objects.filter(user=user, s3bucket=obj, **kwargs).exists()
```

Assuming here we just want to check if there is at least a record for the user.

(same results but without the exceptions which probably are a bit drastic in this case)

controlpanel/api/rules.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+    if other is None:
+        return True
+
+    return user == other
+
+
+add_perm('api.list_user', is_authenticated)
+add_perm('api.create_user', is_authenticated & is_superuser)
+add_perm('api.retrieve_user', is_authenticated & is_self)
```
xoen: How does `is_self` works here?

Is `add_perm()` passing `other` somehow or `other` is always `None` and `is_self()` always return `True`?

controlpanel/api/rules.py
andyhd: These function are imported from `rules`, which is the [Django Rules library](https://github.com/dfunckt/django-rules) (eg: [is_superuser](https://github.com/dfunckt/django-rules/blob/46c594ec2abb605647af49bfbbca17326c3f9df3/rules/predicates.py#L275))

controlpanel/api/rules.py
andyhd: This `is_superuser` predicate checks the `User.is_superuser` field. Django Rules doesn't use the `django.contrib.auth.models.Permission` class, it is not database backed.

controlpanel/api/rules.py
andyhd: Ah! I knew there must be a simpler way 👍

controlpanel/api/rules.py
andyhd: When we call `user.has_perm('api.retrieve_user', other_user)`, the auth backend first checks `is_self(request.user)` and if the result would be `False` (which it never is, because it doesn't make sense), we would get `PermissionDenied`. Otherwise, it then calls `is_self(request.user, other_user)`.

requirements.lock
```diff
@@ -78,6 +78,7 @@ raven==6.10.0
 requests==2.22.0
 requests-oauthlib==1.2.0
 rsa==4.0
+rules==2.0.1
```
xoen: Ah OK, this is where `add_perm()`, etc...comes from, and this is what Ravi meant by "there is no database":

> rules is a tiny but powerful app providing object-level permissions to Django, without requiring a database. [...]

tests/api/permissions/test_users3bucket_permissions.py
```diff
@@ -1,165 +1,114 @@
...
+        json.dumps(data),
+        content_type="application/json",
+    )
+
+
+def delete(client, users3bucket, *args):
+    return client.delete(reverse('users3bucket-detail', (users3bucket.id,)))
+
+
+@pytest.mark.parametrize(
```
xoen: I like how you factored out the tests to check the response status codes 👍

tests/frontend/views/test_app.py
```diff
@@ -0,0 +1,224 @@
...
+
+        (list_all, 'superuser', status.HTTP_200_OK),
+        (list_all, 'app_admin', status.HTTP_403_FORBIDDEN),
+        (list_all, 'normal_user', status.HTTP_403_FORBIDDEN),
+
+        (detail, 'superuser', status.HTTP_200_OK),
+        (detail, 'app_admin', status.HTTP_200_OK),
+        (detail, 'normal_user', status.HTTP_403_FORBIDDEN),
+
+        (create, 'superuser', status.HTTP_302_FOUND),
```
xoen: `302 FOUND`, weird...a redirect? 

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) seems to suggest it should only be used for `GET` or `HEAD` 🤷‍♂ :

> Even if the specification requires the method (and the body) not to be altered when the redirection is performed, not all user-agents conform here [...]. It is therefore recommended to set the 302 code only as a response for GET or HEAD methods

Also, the description says:

> `302 Found` redirect status response code indicates that the resource requested has been temporarily moved [...]

Maybe plain old `201 Created`/`204 No Content` would be more appropriate for this stuff, but I only had a quick look.

andyhd: These are frontend views, not a REST API. 302 is the status code that Django uses for `RedirectView` class-based views. Arguably, it should be 303.

PR #718, opened 2019-08-08
Ensure bucket names are correctly prefixed

andyhd:
The Javascript to automatically prefix a new bucket name with the `env` name was not updated.

This change fixes the Javascript and adds a Django form field validator to raise an error if the prefix is missing (if Javascript is disabled).


andyhd:
Fixes https://github.com/ministryofjustice/analytics-platform/issues/104

PR #719, opened 2019-08-08
Gracefully handle state mismatch

andyhd:
Users sometimes get a 400 error on OIDC callback. Not sure of the root cause, but this patch gracefully handles the exception raised and redirects the user to the login page.

Fixes https://github.com/ministryofjustice/analytics-platform/issues/105

### DIFF
--- a/controlpanel/oidc.py
+++ b/controlpanel/oidc.py
@@ -1,10 +1,17 @@
+import logging
 from urllib.parse import urlencode
 
 from django.conf import settings
+from django.core.exceptions import SuspiciousOperation
+from django.http import HttpResponseRedirect
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


r4vi:
where do we store the state? does cpanel have a redis session store? I'm guessing this might be because the 'state' is stored in memory and then the user is redirected to another replica

andyhd:
No redis, so must be in memory. I figured it might be a sessionAffinity thing, but the Service is configured for ClientIP sessionAffinity with a timeout of 3 hours. /shrug

andyhd:
Or have I misunderstood how sessionAffinity works?

andyhd:
Adding a Redis pod would be easy enough

r4vi:
no, you understand it right... but maybe the replica they were affined  to died, so they'd end up on another one

Issue #717, opened 2019-08-07
Our login rules for passwordless users expect `nickname` field


r4vi:
https://github.com/ministryofjustice/analytics-platform-control-panel/blob/ee6159cf8855c3a6da319bcabced18d73fc168e5/controlpanel/api/auth0.py#L193

When we create a user, most of the time it seems that auth0 creates a `nickname` on the auth0 user object using the first part of their email address. This doesn't happen in a small minority of cases (<5%). To avoid this breaking for those unfortunate users we should explicitly set `nickname` when creating a user.
Related to:
https://github.com/ministryofjustice/analytics-platform/issues/98

PR #720, opened 2019-08-08
Fallback to email username if no nickname

andyhd:
If the user has no nickname from userinfo, default to the username portion of their email address.

See https://github.com/ministryofjustice/analytics-platform/issues/98
Also https://github.com/ministryofjustice/analytics-platform-control-panel/issues/717

### DIFF
--- a/controlpanel/api/auth0.py
+++ b/controlpanel/api/auth0.py
@@ -187,6 +187,9 @@ class ManagementAPI(APIClient):
     audience = base_url
 
     def create_user(self, email, email_verified=False, **kwargs):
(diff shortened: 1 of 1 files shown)


controlpanel/api/auth0.py
```diff
@@ -187,6 +187,9 @@ class ManagementAPI(APIClient):
     audience = base_url
 
     def create_user(self, email, email_verified=False, **kwargs):
+        if "nickname" not in kwargs:
```
r4vi: :pray:

PR #722, opened 2019-08-13
Hide disallowed UI based on object permissions

andyhd:
Check `request.user.has_perm('...', object)` for whether to show certain UI elements, eg Delete buttons.

This should fix some bugs including:
* https://github.com/ministryofjustice/analytics-platform/issues/106
* https://github.com/ministryofjustice/analytics-platform/issues/107
* https://github.com/ministryofjustice/analytics-platform/issues/108


controlpanel/frontend/jinja2/datasource-detail.html
```diff
@@ -159,7 +156,7 @@ <h2 class="govuk-heading-m">Data access group</h2>
   {% endif %}
 </section>
 
-{% if user_is_admin %}
+{% if request.user.has_perm('api.destroy_s3bucket', object) %}
```
xoen: Maybe we should have a permission viewing for the logs.
I know it's the same result but semantically these are different operations.

Potentially we may have different permissions to view logs.

A part from this YAGNI the other benefit would be that if someone tweaks the permissions required to delete an S3 bucket we don't end up also accidentally changing the permissions to view logs.

andyhd: yeah, i was being lazy - we have a permission for viewing app logs, but not for viewing s3 bucket logs

controlpanel/frontend/jinja2/webapp-detail.html
```diff
@@ -176,9 +172,9 @@ <h2 class="govuk-heading-m">
     </tfoot>
   </table>
 
-{% if request.user.is_superuser %}
+{% if request.user.has_perm('api.add_app_admin', object) %}
   {% if admin_options|length %}
-  <form action="{{ url('add-app-admin', kwargs={ "pk": app.id }) }}" method="post" class="govuk-form-group">
+  <form action="{{ url('add-app-admin', kwargs={ "pk": object.id }) }}" method="post" class="govuk-form-group">
```
xoen: is object still an `app`? `object` is quite generic and may easily lead to bugs next time we tweak this view.

andyhd: `object` is the standard context variable set by the generic view class, but you are correct, [the documentation recommends changing this](https://docs.djangoproject.com/en/2.2/topics/class-based-views/generic-display/#making-friendly-template-contexts) - i'll fix it

PR #724, opened 2019-08-21
Add Server Sent Events consumer for push updates

andyhd:
## What

* Removed `django_eventstream` as I just couldn't get it to work
* Added `controlpanel.frontend.consumers.SSEConsumer` which sends a SSE event-stream response and pushes events to the client asynchronously
* Updated `controlpanel.frontend.consumers.BackendTaskConsumer` to send events via `SSEConsumer` on start and end of tool deployment
* Bumped a few versions, most significantly Python -> 3.7.3

## How to run

You need a Redis server running - put the host and port into `REDIS_HOST` and `REDIS_PORT` env vars respectively - or `make run-worker` will run it for you
Run the webserver as normal with `python manage.py runserver` (or `make run`)
You also need to run a worker process with `python manage.py runworker background_tasks` (or `make run-worker`)

See [Django Channels](https://channels.readthedocs.io/en/latest/introduction.html)

Tests coming as soon as I figure out how to write them 😩


andyhd:
Not merging yet as I need to update the Helm chart to deploy a Redis instance

controlpanel/frontend/consumers.py
```diff
@@ -1,59 +1,156 @@
...
 
-        out = deploy.stdout.read()
-        send_tool_status_change(tool.name, out)
-        log.debug(out)
+        send_tool_status_update(user.auth0_id, tool.name, "Ready")
 
-    def restart_tool(self, message):
+    def tool_restart(self, message):
         tool = Tool.create(message['tool_name'])
         user = User.objects.get(auth0_id=message['user_id'])
```
xoen: Apologies if I missed it, but how does authorisation works? We trust the incoming message and take the user ID from the message. I'm not familiar with this Python Channel thing so possibly I'm missing something.

xoen: I suppose to send a message to the channel you need to talk with Redis.

andyhd: The channels router has authentication middleware wrapping the whole thing here:
https://github.com/ministryofjustice/analytics-platform-control-panel/pull/724/files#diff-310d0aa3904afc2adf8ab19949784881R14

andyhd: and yes, you can't talk directly to Redis from outside the cluster

xoen: Isn't the authentication middleware only running for HTTP requests? I'm not familiar with Python Channels stuff (I'm having a read at their docs), maybe you don't need authorisation? 🤷‍♂

xoen: I suppose the worst case scenario is an unauthorised user restart or deploy someone's tool, not the end of the world.

controlpanel/frontend/views/tool.py
```diff
@@ -54,18 +56,20 @@ class DeployTool(LoginRequiredMixin, RedirectView):
     def get_redirect_url(self, *args, **kwargs):
         name = self.kwargs["name"]
 
-        try:
-            Tool.create(name).deploy_for(self.request.user)
+        async_to_sync(channel_layer.send)(
+            'background_tasks',
+            {
+                'type': 'tool.deploy',
```
xoen: I imagine the type `tool.deploy` will somehow trigger `BackgroundTaskConsumer#tool_deploy()`, is this just a naming convention or we define this mapping somewhere? Didn't see anything in `control-panel/controlpanel/routing.py` - not a problem, I'm just curious.

xoen: Oh yes, it's a [naming convention](https://channels.readthedocs.io/en/latest/topics/consumers.html#basic-layout) 👍

controlpanel/frontend/consumers.py
andyhd: good point, the authentication middleware is only for http, but there is no exposed endpoint/port for the background-tasks channel - as i understand it, you have to write to Redis to send a message on the channel, and Redis is not exposed outside the cluster

xoen: Yes that's the conclusion I got to as well by reading this and some of the Django Channels documentation. I guess you're locking down Redis and that should be enough 🤷‍♂

PR #726, opened 2019-08-29
Raise an exception when required configuration is missing

xoen:
Replaced `os.environ.get()` call with `os.environ[]` when there is
no sensible default and the user should configure a setting.

**NOTE**: Tests pass and `make run` seems to work but I'm not 100% this
is not breaking something because I need to set these variable in some
other places.

### DIFF
--- a/controlpanel/settings/common.py
+++ b/controlpanel/settings/common.py
@@ -171,20 +171,20 @@ OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = 15 * 60
 OIDC_CALLBACK_CLASS = 'controlpanel.oidc.StateMismatchHandler'
 
 # Hostname of the OIDC provider
-OIDC_DOMAIN = os.environ.get("OIDC_DOMAIN")
+OIDC_DOMAIN = os.environ["OIDC_DOMAIN"]
 
 # OIDC endpoints
-OIDC_OP_AUTHORIZATION_ENDPOINT = os.environ.get("OIDC_OP_AUTHORIZATION_ENDPOINT")
-OIDC_OP_JWKS_ENDPOINT = os.environ.get("OIDC_OP_JWKS_ENDPOINT")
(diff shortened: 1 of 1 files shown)


controlpanel/settings/common.py
```diff
@@ -252,9 +252,9 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
-        "NAME": os.environ.get("DB_NAME", PROJECT_NAME),
```
andyhd: This one has a default

controlpanel/settings/common.py
```diff
@@ -252,9 +252,9 @@
...
-        "USER": os.environ.get("DB_USER", ""),
-        "PASSWORD": os.environ.get("DB_PASSWORD", ""),
```
andyhd: These should probably have defaults of `PROJECT_NAME`

xoen:
@andyhd I've changed this but I'm not 100% sure it doesn't break anything unfortunately.

`make test` runs and pass locally and AFAIK all these environment variables involved are set by the helm chart - which is the good news. The bad news is that probably devs will have to add more values to their `.env` file.

Could you have a look and possibly check this branch and see if you see anything wrong with it? Thanks 😌

andyhd:
Ah, now I remember why they have defaults - `collectstatic` loads the settings (but it doesn't need them). Maybe there should be some overrides just for collectstatic.

controlpanel/settings/common.py
xoen: I know but we pass them anyway and I think we should try and be explicit as we have to pass the rest anyway.

xoen:
I see the `docker build` fails (on CircleCI), if `collectstatic` doesn't use them, maybe I can just pass an emtpy string, it would be a bit 💩 - but I'll play with it and see how it goes.

xoen:
Right. I'm going to abort this as it was more challenging than imagined especially for the benefits of doing this 😭

PR #727, opened 2019-08-29
Remove clear SECRET_KEY (rotated) from source code

xoen:
We don't use [`SECRET_KEY`](https://docs.djangoproject.com/en/dev/ref/settings/#secret-key) for anything special (at the moment) but nevertheless this shouldn't be in clear in the source code.

**NOTE**: I've added/changed this in the [config files](https://github.com/ministryofjustice/analytics-platform-config/pull/157) and updated the [helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/375).

### DIFF
--- a/controlpanel/settings/common.py
+++ b/controlpanel/settings/common.py
@@ -207,9 +207,7 @@ OIDC_DRF_AUTH_BACKEND = "controlpanel.oidc.OIDCSubAuthenticationBackend"
 
 # -- Security
 
-SECRET_KEY = os.environ.get(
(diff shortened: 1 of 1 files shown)


xoen:
As discussed with @r4vi  and from what I can see [in the Django documentation for `SECRET_KEY`](https://docs.djangoproject.com/en/dev/ref/settings/#secret-key) rotate/change this shouldn't cause any problems (famous last words 😁 ...)

PR #729, opened 2019-08-30
Deny access to eventstream unless logged in

andyhd:
If an unauthenticated user tries to `GET /events/`, return a 403

### DIFF
--- a/controlpanel/frontend/consumers.py
+++ b/controlpanel/frontend/consumers.py
@@ -38,6 +38,11 @@ class SSEConsumer(PatchedAsyncHttpConsumer):
         """
         self.streaming = True
 
+        user = self.scope.get('user')
+        if not user or not user.is_authenticated:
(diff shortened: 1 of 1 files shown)


controlpanel/frontend/consumers.py
```diff
@@ -38,6 +38,11 @@ def __init__(self, *args, **kwargs):
         """
         self.streaming = True
 
+        user = self.scope.get('user')
```
r4vi: if the events are filtered by the users id then is this required?

andyhd: It is, because `AnonymousUser` doesn't have a `auth0_id` field

andyhd: Of course, there are several ways to handle this - I could have changed `user.auth0_id` to `getattr(user, 'auth0_id', None)` I suppose.

PR #730, opened 2019-08-30
Set public access block on created buckets

andyhd:
Turn on public access blocking on S3 buckets created in the Control Panel by default.

We have an AWS Lambda function which checks buckets block public access every day at 9AM UTC as a backup, but it's better to do it up front.

### DIFF
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -78,6 +78,24 @@ class AWSClient(object):
                 'TagSet': [{'Key': str(k), 'Value': str(v)} for k, v in tags.items()]
             })
 
+    def put_public_access_block(
+        self,
+        bucket_name,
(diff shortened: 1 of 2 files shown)


andyhd:
I know the `AWSClient` class is largely redundant - I've been working on a refactor (natch) but it's not ready yet.

xoen:
This is to prevent someone accidentally change the access of an S3 bucket right?

I'm asking as I see the buckets are already created with `acl='private'`, so they should be private unless I'm missing something.

andyhd:
Yes. The buckets are created with `Objects can be public` access type, but with the `private` ACL, as you say. Then a Lambda function runs every day at 9am and sets them to `Bucket and objects not public`. This change just creates the buckets with that access type in the first place.

PR #732, opened 2019-09-04
Tool refactor

andyhd:
* Make `Tool` class (representing an app available in the Software Catalog) into a Django model, to allow adding tools easily in the database. Tool configuration is also in the database, rather than in the Django settings. Also removes the need for the weird metaclass and singleton ToolRegistry.
* Make `ToolDeployment` class emulate a Django model, but wrapping the Kubernetes API to represent a user's Kubernetes Deployment of a tool. Abstracts the use of Helm and the Kubernetes API to install, delete and get the status of the tool, so that we can easily replace Helm if we want to in future (I do).
* Simplify tool status messages sent in SSEs to `Not deployed`, `Deploying`, `Ready`, `Failed`, `Idled` or `Unknown`
* Update UI on receiving SSEs to show relevant buttons


controlpanel/api/migrations/0010_DATA_add_tools.py
```diff
@@ -0,0 +1,80 @@
...
+import secrets
+
+from django.conf import settings
+from django.db import migrations
+
+
+TOOLS = [
+    {
+        'chart_name': 'rstudio',
+        'version': '2.1.1',
```
xoen: One thing I'm thinking is that this schema would mean to add more than one "rstudio" entry for different versions (as it's not technically normalised). I suppose adding more tables would make things more complicated.

I like the values being version-specific in case we'll somehow to break compatibility.

The drawback of having this in the DB is that we'll have to add an entry to allow a release of a newer version of a chart. But I suppose this is a feature as we want to "whitelist" the versions we support (although as far as I can tell from quick look we still deploy latest version right?).

Alternative is we get this info "life" from the helm repo, but not sure it's a good idea as it would not be a curated list of tools we support anymore.

(I'm starting to have a look so this is just a comment with few random thoughts)

Makefile
```diff
@@ -72,7 +72,8 @@ transpile:
 redis: /usr/local/var/run/redis.pid
 	@echo
 	@echo "> Running Redis server..."
-	@redis-server &>/tmp/redis.log &
+	@if [ -z "$$REDIS_PASSWORD" ]; then REQUIREPASS="--requirepass=$$REDIS_PASSWORD"; fi
```
xoen: nice ☺️  - I also had to remove the pidfile dependency in my machine as for some reason it was complaining - probably the pidfile location is very machine-dependant. Although I like the idea of starting redis only if it's not already running.

controlpanel/frontend/jinja2/tools.html
```diff
@@ -12,56 +12,45 @@ <h1 class="govuk-heading-xl">Your tools</h1>
 <table class="govuk-table">
```
xoen: FYI: I tried to run this locally. Not sure why yet but I can't see anything in the tools table (it says `0 tools`) - helm releases are there and pods are running 🤷‍♂

controlpanel/api/migrations/0010_DATA_add_tools.py
andyhd: I wasn't particularly intending to offer users multiple versions of a tool, but when we release a new version of a tool, existing deployments will be different and we should be able to show that difference.

I have changed the Helm command to pass `--version={version}`, so we will need to deal with this as soon as we have a new version of a tool chart.

I'm not keen to rely on the Helm repo or Tiller as a source of truth - Tiller has been removed in Helm 3 and I'm not even sure Helm is worth using. It would be better to add more metadata to the Kubernetes deployments/services/etc to identify the tool version - I think this is the direction Helm is going too.

Makefile
andyhd: Yeah, I didn't change that because I knew you had already changed it.

controlpanel/frontend/jinja2/tools.html
andyhd: Did you run the migrations?

controlpanel/api/migrations/0010_DATA_add_tools.py
xoen: Makes sense, so we have one "good"/recommended version of a tool in the DB and when a user has an old/different version we offer them the option to "Upgrade" 👍

controlpanel/frontend/jinja2/tools.html
xoen: I didn't before. After running them it's all good ✨  😍

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+
+
+class ToolDeploymentError(Exception):
+    pass
+
+
+def deploy_tool(tool, user, **kwargs):
+    values = {
+        'username': user.username.lower(),
+        'Username': user.username.lower(),  # XXX backwards compatibility
```
xoen: I know the comment says "backward compatibility" but why do we pass both?

Oh is it because some of the tools have `Username` as value instead of `username`?

andyhd: Yes, exactly that. I thought it was easier to just pass both variants and standardise the Helm charts later.

controlpanel/api/migrations/0010_DATA_add_tools.py
```diff
@@ -0,0 +1,80 @@
...
+            'authProxy.auth0_client_id': '%(client_id)s',
+            'authProxy.auth0_client_secret': '%(client_secret)s',
+            'cookie_secret': '<SECRET_TOKEN>',
+        },
+    },
+    {
+        'chart_name': 'airflow-sqlite',
+        'version': '0.2.2',
+        'name': 'Airflow',
+        'description': '1.10.3',
```
xoen: I wonder if `description` in these `Tool` records should be the actual description (e.g. a pseudo-user friendly description of the tool, possibly different than [the one in the helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/rstudio/Chart.yaml#L2)) and there is room for a `appVersion` field with [the version of the tool](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/rstudio/Chart.yaml#L5) (e.g. what we currently have in the `description`).

Don't know if it makes sense.

andyhd: Yeah, I thought about that. We can put anything in the description field in the database - these initial data migrations are only to keep the Control Panel working as it is (although there may be many different versions of Airflow, RStudio and Jupyter-Lab already deployed, so I think the version stuff may need to be switched off for now).

controlpanel/api/migrations/0010_DATA_add_tools.py
```diff
@@ -0,0 +1,80 @@
...
+
+TOOLS = [
+    {
+        'chart_name': 'rstudio',
+        'version': '2.1.1',
+        'name': 'RStudio',
+        'description': 'RStudio: 1.2.1335+conda, R: 3.5.1, Python: 3.7.1, patch: 3',
+        'values': {
+            'authProxy.cookieSecret': '<SECRET_TOKEN>',
+            'authProxy.auth0.domain': '%(domain)s',
```
xoen: So, to be sure I understand what's going on. The `'%(name)s'` means something like this right?:

```python
template = 'my name is %(name)s'
name = "John Doe"
rendered = template % name # rendered == "my name is John Doe"
```

So it's basically like `%s` in `printf()` but using the variable name right?

andyhd: Yeah, it's the old Python string interpolation style - `'%(domain)s' % {'domain': 'some value'}` is equivalent to `'{domain}'.format({'domain': 'some value'})`. The `<SECRET_TOKEN>` string was previously `{{secret_token}}`, but you mentioned getting an error with the migration, so I changed it so that there was no chance of a syntax problem or reading comprehension fail. I could go back to the `format` style, but it's a data migration that is intended to be run only once, so no point changing it.

controlpanel/api/migrations/0010_DATA_add_tools.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def initialize_tools(apps, schema_editor):
+    Tool = apps.get_model('api', 'Tool')
+
+    for metadata in TOOLS:
+        conf = settings.TOOLS[metadata['chart_name']]
+        for key, value in metadata['values'].items():
+            if isinstance(value, str):
+                metadata['values'][key] = value % conf
```
xoen: Maybe it's me being paranoid, but isn't there a change that a password/value contains `%(` something and this breaks? I guess it's a migration and if we don't have anything like that now we're possibly fine...?

controlpanel/api/migrations/0010_DATA_add_tools.py
xoen: > although there may be many different versions of Airflow, RStudio and Jupyter-Lab already deployed, so I think the version stuff may need to be switched off for now

What do you mean by "switch off the version stuff"? I didn't get to the upgrade logic that but I would imagine we'll use the `Tool.version` field to determine if a user tool is out-of-date no?

controlpanel/api/migrations/0010_DATA_add_tools.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0009_tool'),
+    ]
+
+    operations = [
+        migrations.RunPython(initialize_tools, lambda *args: None),
```
xoen: Noob question: What is the `, lambda *args: None),` bit doing? I had a quick look at `RunPython` and the second argument seems to be `reverse_code` but probably I'm missing something.

controlpanel/api/cluster.py
```diff
@@ -64,6 +67,13 @@
     },
 }
 
+TOOL_DEPLOYING = 'Deploying'
```
xoen: Nice that we use constants.

There is something that feels odd about having these constants in the `api.cluster` module tho for some reason.
I wonder if a better place for them would be inside the `ToolDeployment` class, e.g. `ToolDeployment.FAILED` or `ToolDeployment.IDLED` 🤷‍♂

controlpanel/api/models/tool.py
```diff
@@ -0,0 +1,131 @@
...
+    chart_name = models.CharField(max_length=100, blank=False)
+    name = models.CharField(max_length=100, blank=False)
+    values = JSONField(default=dict)
+    version = models.CharField(max_length=100, blank=False)
+
+    class Meta(TimeStampedModel.Meta):
+        db_table = "control_panel_api_tool"
+        ordering = ('name',)
+
+    def __repr__(self):
```
xoen: How is this used? Do we override this just because it makes things easier to debug or it serves some other purpose?

controlpanel/api/models/tool.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def delete(self):
+        """
+        Remove the release from the cluster
+        """
+        cluster.delete_tool_deployment(self)
+
+    @property
+    def host(self):
+        return f'{self.user.slug}-{self.tool.chart_name}.{settings.TOOLS_DOMAIN}'
```
xoen: It seems to work in my simple case but I notice that this is slightly different compared to what we had before: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/732/files#diff-aad7b41260663fb400ff942a02fd3d14L200

```Python
    @property	
    def name(self):	
        return self.deployment.metadata.labels["app"]	

    @property	
    def username(self):	
        return self.user.username.lower()	

    @property	
    def url(self):	
        return f"https://{self.username}-{self.name}.{settings.TOOLS_DOMAIN}"
```

I wonder if there is some weird edge-case lurking in here? 

**Nevermind** it seems like the `app` label in the deployments is actually just `{{{ .Chart.Name }}`. I'm mainly wondering what would happen if we decide to allow 2 major/supported versions of a tool to run side-by-side. But it seems like a problem for another day which we may never have or what to support.

controlpanel/frontend/static/javascripts/modules/tool-deploy.js
```diff
@@ -1,5 +1,5 @@
 moj.Modules.tool_deploy = {
-  class: ".deploy-tool",
+  class: ".tool-deploy-form",
```
xoen: [What I did](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/4f5df1955ac37a1b103152b7ec027df965324d8f) in my branch was to change the CSS class to `form.background-submit` as I wanted pretty much the same behaviour for the "Upgrade" button.

Don't know if that makes sense but happy for you to get that if it does.

tests/api/views/test_tool_deployment.py
```diff
@@ -1,29 +1,32 @@
...
     response = client.post(reverse('deployment-list'), {"name": tool_name})
     assert response.status_code == status.HTTP_400_BAD_REQUEST
 
 
-def test_create_when_valid_tool_name(client, tool_deploy, users):
-    tool_name = 'rstudio'
-    user = users['normal_user']
-    client.force_login(user)
-    response = client.post(reverse('deployment-list'), {"name": tool_name})
+def test_create_when_valid_tool_name(client, deploy_tool, tool, users):
```
xoen: Probably stupid question: Who is calling this method and passing the `tool`/`deploy_tool`/etc arguments?

controlpanel/api/migrations/0010_DATA_add_tools.py
andyhd: It only runs once with the values currently in the settings - none of which have `%` in them.

controlpanel/api/migrations/0010_DATA_add_tools.py
andyhd: This code may break because users already have versions of Airflow, RStudio and Jupyter that are not found in the database.

controlpanel/api/migrations/0010_DATA_add_tools.py
andyhd: Yes, the second argument is the reverse migration. `lambda *args: None` is the noop function.

controlpanel/api/cluster.py
andyhd: Yeah, I agree, but if the constants are defined in `models/tool.py` and used in `cluster.py` you get a circular import. That's why I also defined the exceptions in `cluster`. Maybe I should move them into another module that both the model and `cluster` can import from.

controlpanel/api/models/tool.py
andyhd: Yes, it's just so that it prints something useful at the console.

controlpanel/api/models/tool.py
andyhd: Yeah, this is what we were talking about with having to put version numbers in the hostname or maybe using the randomly generated release names instead. I think we should standardise the hostnames - this particular pattern is not necessarily the right one, but we currently have two different ones (`username-rstudio` and `jupyter-lab-username`) which sucks.

controlpanel/frontend/static/javascripts/modules/tool-deploy.js
andyhd: Yeah, I like what you did, although I might just add `background-submit` rather than replace `tool-deploy-form`. But I'll let you submit that change in your own PR :)

tests/api/views/test_tool_deployment.py
andyhd: You should read the [pytest documentation](https://docs.pytest.org/en/latest/).
Pytest finds all functions that match a specified pattern (by default `test_*`), executes them and passes in the fixtures that match the argument names.
* `client` is a fixture defined in `pytest-django` that returns the Django test client
* `deploy_tool` is defined above and returns a mock patch of the `deploy_tool` function in `controlpanel.api.cluster`
* `tool` is also defined above and returns a model instance
* `users` is defined in `tests/conftest.py`. (pytest evaluates all the `conftest.py` files in the path to the current test function)

Pytest seems a little magical to start with, but it is the best and I think most popular Python test framework.

controlpanel/api/migrations/0010_DATA_add_tools.py
xoen: I wonder if we could use [`RunPython.noop()`](https://docs.djangoproject.com/en/2.2/ref/migration-operations/#django.db.migrations.operations.RunPython.noop) instead so that it's more apparent...not a big deal I suppose as it's a one off migration.

controlpanel/api/migrations/0010_DATA_add_tools.py
xoen: I plan to have a better look at the rest but by "may break" do you mean that if a user has a version of a tool which is not in the CP DB it will not be shown in the Tools page at all?

I will probably mess with the DB locally later and see how it goes but ideally we may want to show the users' tools regardless of our latest/supported version (so that they can upgrade).

controlpanel/api/cluster.py
xoen: I see what you mean about the cycle now 🤒- not the end of the world anyway.

tests/api/views/test_tool_deployment.py
xoen: > users is defined in tests/conftest.py. (pytest evaluates all the conftest.py files in the path to the current test function)

😮 now I see where that comes from. VS Code seems confused by pytest's magical fixtures but I can see the logic now.

Regarding the two fixtures defined above the tests... 🤦‍♂ my brain must have been fried yesterday afternoon 😆

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+def list_tool_deployments(user, search_name=None, search_version=None):
+    deployments = []
+    k8s = KubernetesClient()
+    results = k8s.AppsV1Api.list_namespaced_deployment(user.k8s_namespace)
+    for deployment in results.items:
+        app_name = deployment.metadata.labels["app"]
+        _, version = deployment.metadata.labels["chart"].rsplit("-", 1)
+        if search_name and search_name not in app_name:
+            continue
+        if search_version and not version.startswith(search_version):
```
xoen: Do we need to check the version? I'm wondering if this is causing the problem we were talking about [in the other comment](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/732/files#r321617248) where if the user is using a different version is not shown/can't restart.

I've tried to change the version of the RStudio `Tool` record in the DB locally and the list of tools got confused/confusing. I couldn't see my RStudio instance and it seemed as if the RStudio helm chart wasn't installed and offered me to "Deploy" that version I didn't have.

I tried to comment this `if` statement and it doesn't seem to help 🤔. I think completely hide an instance if it's older than current version would make our user very confused and unhappy. I'll play with it locally in case I have some good idea ☺️

controlpanel/frontend/static/javascripts/modules/tool-status.js
```diff
@@ -21,9 +24,32 @@ moj.Modules.toolStatus = {
   buildEventHandler(listener) {
     return event => {
       const data = JSON.parse(event.data);
-      if (data.toolName == listener.dataset.toolName) {
-        listener.innerText = data.status;
+      if (data.toolName != listener.dataset.toolName) {
+        return;
+      }
+      listener.querySelector(this.statusLabelClass).innerText = data.status;
+      switch (data.status.toUpperCase()) {
```
xoen: AFAICT `data.status` is already sent in uppercase so probably `toUpperCase()` could be removed.

controlpanel/frontend/static/javascripts/modules/tool-status.js
```diff
@@ -21,9 +24,32 @@ moj.Modules.toolStatus = {
...
+        case 'FAILED':
+          this.showActions(listener, ['reset', 'remove']);
+          break;
       }
     };
+  },
+
+  showActions(listener, action_names) {
+    listener.querySelectorAll(this.actionClass).forEach(action => {
+      action.classList.toggle(this.hidden, !action_names.includes(action.dataset.actionName));
```
xoen: Clever.

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+        log.warning(f"Multiple objects returned for {tool_deployment}")
+        return TOOL_STATUS_UNKNOWN
+
+    conditions = {
+        condition.type: condition
+        for condition in deployment.status.conditions
+    }
+
+    if 'Available' in conditions:
+        if conditions['Available'].status == 'True':
```
xoen: The status of k8s deployments statuses (no pun intended) is a bit of a mess.

If it helps, I've noticed that I got `Unknown` status while one of the two containers in RStudio was still starting, `type` was `"Available"` but `status` was `"False"` (what a mess!):

```YAML
status:
  conditions:
  - lastTransitionTime: "2019-09-06T14:11:01Z"
    lastUpdateTime: "2019-09-06T14:11:01Z"
    message: Deployment does not have minimum availability.
    reason: MinimumReplicasUnavailable
    status: "False"
    type: Available
```

(but bear in mind that I think my RStudio is currently broken so I don't know how typical that would be)

Maybe we could have a "Starting" status? I have no idea what the k8s condition would be if the pod doesn't start propertly then, would we stuck displaying "Starting"?

For instance, my RStudio's pod is now stuck in `CrashLoopBackOff` status but the `Deployment`'s condition is still as above 😕

xoen: I see that you got rid of the code which was checking the pods' `statuses`|`states` - I haven't looked at that properly yet so not sure if we're missing some edge case we had before but this seems so much cleaner 🌟 😍

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+    values = {
+        'username': user.username.lower(),
+        'Username': user.username.lower(),  # XXX backwards compatibility
+        'aws.iamRole': user.iam_role_name,
+        'toolsDomain': settings.TOOLS_DOMAIN,
+    }
+
+    # generate per-deployment secrets
+    for key, value in tool.values.items():
+        if value == '<SECRET_TOKEN>':
```
xoen: Nice one Andy ☺️

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+            tool.values[key] = secrets.token_hex(32)
+
+    values.update(tool.values)
+    values.update(kwargs)
+    values = ','.join(f'{key}={val}' for key, val in values.items())
+
+    try:
+        return helm.upgrade_release(
+            f'{tool.chart_name}-{user.slug}',
+            f'mojanalytics/{tool.chart_name}',  # XXX assumes repo name
```
xoen: I actually now realise that we use this `mojanalytics`/repo name a lot. Maybe worth moving to a config (happy to have that default to `mojanalytics` so we don't have to add config, etc...) or at least a constant?

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+            f'--version', tool.version,
```
xoen: The fact we now pass the version is quite nice as we don't have to worry about merging a version and having that released for new users (but of course that means we need to update the DB record when we're happy with a new chart version)

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,120 @@ def get_repository(user, repo_name):
...
+    except HelmError as error:
+        raise ToolDeploymentError(error)
+
+
+def list_tool_deployments(user, search_name=None, search_version=None):
+    deployments = []
+    k8s = KubernetesClient()
+    results = k8s.AppsV1Api.list_namespaced_deployment(user.k8s_namespace)
+    for deployment in results.items:
+        app_name = deployment.metadata.labels["app"]
```
xoen: Shame we don't have some label with the helm chart name + version in our helm charts 😞  - Helm documentation mentions this: [The Chart Best Practices Guide - Labels and Annotations](https://helm.sh/docs/chart_best_practices/#labels-and-annotations).

Actually, it seems that at least the latest versions of the tools have the `chart` label which is something like `chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"`. I just did a bit of git digging and obvs Jupyter and Airflow are copied from RStudio chart and RStudio had that label for 2 years so maybe it's safe to use the `chart` label after all - it may be less brittle than use the `app` label.

xoen: Oh hang on, you do use the `chart` label, sorry 🤦‍♂

controlpanel/api/models/tool.py
```diff
@@ -0,0 +1,131 @@
...
+        user = kwargs.get('user')
+        filter = Q(chart_name=None)  # Always False
+        for deployment in cluster.list_tool_deployments(user):
+            _, version = deployment.metadata.labels["chart"].rsplit("-", 1)
+            filter = filter | (
+                Q(chart_name=deployment.metadata.labels["app"]) &
+                Q(version=version)
+            )
+
+        tools = Tool.objects.filter(filter)
```
xoen: I think this is the tricky bit as we get the supported tools from the DB but them map them to the deployments.
When there isn't corresponding record for the version the user has we don't return it.

If I try to remove the `version=` from the filter and from `cluster.list_tool_deployments()` things gets weird.

I don't know what's the best solution to keep showing unsupported/old versions of a tool in the list (and allowing user to restart, upgrade, etc...). Maybe you have some ideas regarding this, probably worth chat as I can imagine my comments may sound a bit random, sorry.

xoen: One possibility would be have deployments in the list as they come from kubernetes (potentially augment with data coming from DB if the tool is of the same version) and use the DB mainly for deployment, e.g. as "lists of tools/versions" we support).

That would mean the `ToolDeployment` doesn't [incapsulate an instance of a `Tool`](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/732/files#diff-1b16cf3d380938c07fe469d8d7058f70R78) or possibly it's optional or something (but will have to be there when deploying).

controlpanel/api/migrations/0010_DATA_add_tools.py
andyhd: Ah, crap - I should have known there was a thing for that. I'll use `noop` 👍

controlpanel/frontend/static/javascripts/modules/tool-status.js
andyhd: It definitely isn't sent in uppercase.

controlpanel/api/cluster.py
andyhd: I don't think having a `Deploying` *and* `Starting` status is worthwhile - in the (hopefully) normal case, you would see `Starting` only very briefly between `Deploying` and `Ready`.

Are you saying that you want more code to figure out what is the reason for an `Unknown` status? I think that may not be worth the effort and I don't want to show the reason message to users, but maybe we should monitor the frequency of `Unknown` statuses to decide if it is worth spending more time on it - I'll add a `log.warning`.

controlpanel/api/models/tool.py
andyhd: Yeah, I think getting the metadata from Kubernetes is the right approach. I'll have a try at making Kubernetes the source of truth instead of the db.

controlpanel/api/cluster.py
xoen: Re "deploying vs. starting" these days some tools (at least RStudio) have slow init containers but we may get away with the simple "Deploying" and iterate if necessary 👍 

The less code the happier I am you know. Hopefully that "Unknown" status is an edge case and infrequent.

controlpanel/api/models/tool.py
xoen: Yeah, that may work. Sorry for find this and all the comments 😢 - feel free to Slack me if you need to discuss anything.

andyhd: No need to apologise, this is a good code review! The Tool version stuff isn't meant to be complete, but I realise now that it will actually break for some current users, so I'll disable for this PR and let you sort it out in yours :trollface:

xoen: Yeah, as said I like the high level idea and the way this is structured but unfortunately there will be always users with older versions and hiding their tools is not gonna work ☺️

Everything will collapse the moment I touch anything 😱 😋

controlpanel/api/models/tool.py
```diff
@@ -0,0 +1,131 @@
...
+    def _poll(self):
+        """
+        Poll the deployment subprocess for status
+        """
+
+        if self._subprocess.poll() is None:
+            return cluster.TOOL_DEPLOYING
+
+        if self._subprocess.returncode:
+            log.error(err_msg)
```
xoen: FYI: I'm getting `name 'err_msg' is not defined` from this as `err_msg` is not defined (apologies if you were already aware of this).

xoen: Played with this - I'm not familiar with running subprocess in Python at all, replace with `log.error(self._subprocess.returncode)` prints `ERROR Failed deploying RStudio for xoen` which seems very weird to me as I thought a process's return code was a number not a string 😁

controlpanel/api/migrations/0009_tool.py
```diff
@@ -0,0 +1,34 @@
...
+        migrations.CreateModel(
+            name='Tool',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created', django_extensions.db.fields.CreationDateTimeField(auto_now_add=True, verbose_name='created')),
+                ('modified', django_extensions.db.fields.ModificationDateTimeField(auto_now=True, verbose_name='modified')),
+                ('description', models.TextField(blank=True)),
+                ('chart_name', models.CharField(max_length=100)),
+                ('name', models.CharField(max_length=100)),
+                ('values', django.contrib.postgres.fields.jsonb.JSONField(default=dict)),
```
r4vi: nice

controlpanel/api/cluster.py
```diff
@@ -346,3 +356,123 @@ def get_repository(user, repo_name):
...
+    )
+
+    if not deployments:
+        raise ObjectDoesNotExist(tool_deployment)
+
+    if len(deployments) > 1:
+        log.warning(f"Multiple matches for {tool_deployment!r} found")
+        raise MultipleObjectsReturned(tool_deployment)
+
+    return deployments[0]
```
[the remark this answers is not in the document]
andyhd: Currently you can't have the same tool installed twice, and I wasn't really trying to change that behaviour in this PR

controlpanel/frontend/jinja2/tool-list.html
```diff
@@ -38,11 +38,14 @@ <h1 class="govuk-heading-xl">Your tools</h1>
...
-           href="{{ url('restart-tool', kwargs={'name': tool.chart_name}) }}">
-          Reset
-        </a>
+        <form action="{{ url('restart-tool', kwargs={'name': tool.chart_name}) }}"
+              class="govuk-button hmcts-button--secondary tool-action {% if not deployment %} govuk-visually-hidden {% endif %}"
+              data-action-name="restart"
+              method="post">
+          {{ csrf_input }}
+          {# <input type="hidden" value="{{ tool.version }}" name="version"> #}
+          <button class="govuk-button hmcts-button--secondary right">Reset</button>
```
xoen: I vote to keep this `Restart` or users will confuse this button with "rest the environment to factory" (e.g. delete conda envs etc...)

But nice one, this should fix the restart button `405` error 👍

andyhd: Sorry, copy-pasta here - class names need tidying up

controlpanel/api/cluster.py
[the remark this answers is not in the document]
xoen: Agreed. One instance of each tool is what we had and we still want for the time being 🙆‍♂

controlpanel/frontend/jinja2/tool-list.html
xoen: Oh, I see this is where the "Reset" caption comes from.

PR #735, opened 2019-09-10
Get deployment status with request user's id token

andyhd:
When deploying a tool, the deployment status updates are pushed to the client via SSE. The updates should go `Deploying` > `Ready` / `Failed`, but the `Ready` or `Failed` events are never sent in dev and alpha.
This is because the method of checking the status of the deployment in the background task queries the Kubernetes API with the credentials of the `cpanel` service account and it doesn't seem to have permission to read deployments in users' namespaces.
This change passes the request user's id token through to the background deployment task when it is started, to be used as a credential when accessing the Kubernetes API.

### DIFF
--- a/controlpanel/api/models/tool.py
+++ b/controlpanel/api/models/tool.py
@@ -74,6 +74,7 @@ class ToolDeployment:
     objects = ToolDeploymentManager()
 
     def __init__(self, tool, user):
+        self._id_token = None
         self._subprocess = None
         self.tool = tool
         self.user = user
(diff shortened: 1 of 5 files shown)


controlpanel/api/models/tool.py
```diff
@@ -95,6 +96,7 @@ def save(self, *args, **kwargs):
         """
         Deploy the tool to the cluster (asynchronous)
         """
+        self._id_token = kwargs.get('id_token')
```
xoen: Maybe I'm missing something: where is `kwargs` defined here?

xoen: Sorry, that was GitHub collapsed the method definition and it looked like this was in the body of the constructor 😆

PR #736, opened 2019-09-10
Fix links to app logs

andyhd:
Kibana app logs are labelled with `app_name: {helm_release_name}`, where the release name is [generated by the Github org Concourse resource](https://github.com/ministryofjustice/analytics-platform-concourse-github-org-resource/blob/fbbdd516384f048517d1cecca4cd30f8c54a3dba/resource/out#L89) but the Control Panel was linking into Kibana with `app_name: {github_repo_name}-webapp`.


andyhd:
By the way, you can't easily test this in dev, because there aren't any app logs. The `Secure_Geezer` app has an underscore in the repo name, which should be converted to a hyphen in the kibana deep link.

PR #739, opened 2019-09-13
Add email scope to get email claim

andyhd:
Missing `email` scope meant we didn't get the `email` and `email_verified` claims from Auth0 on user login, which in turn meant that new users were created with an empty email address, which causes the `config-user` chart job to fail to complete


controlpanel/settings/common.py
```diff
@@ -185,7 +185,7 @@
 # OIDC JWT signing algorithm
 OIDC_RP_SIGN_ALGO = os.environ.get("OIDC_RP_SIGN_ALGO", "RS256")
 
-OIDC_RP_SCOPES = "openid profile offline-access"
+OIDC_RP_SCOPES = "openid email profile offline-access"
```
xoen: 👌

PR #740, opened 2019-09-13
Protect granting admin access to bucket

andyhd:
Hide admin option if user does not have permission to create bucket admins

Currently, when granting or editing a user's access to a bucket, all three access levels are shown in the form (`read only`, `read write` and `admin`). If a user who does not have permission to create bucket admin users tries to set the access level to `admin`, they will get a 403 error page.

This change prevents them from selecting the `admin` access level option and from seeing the error page.


controlpanel/frontend/views/datasource.py
```diff
@@ -213,13 +213,28 @@ def get_success_url(self):
...
+    template_name = 'datasource-access-grant.html'
 
     def get_form_kwargs(self):
         return FormMixin.get_form_kwargs(self)
 
+    def get_context_data(self):
+        context = super().get_context_data()
+        bucket = get_object_or_404(S3Bucket, pk=self.kwargs['pk'])
+        context['bucket'] = bucket
+        access_group = bucket.users3buckets.all().select_related('user')
```
r4vi: could do: `list(bucket.users3buckets.all().select_related('user').values_list('user__auth0_id', flat=True))` ?
```suggestion
        access_group = list(bucket.users3buckets.all().select_related('user').values_list('user__auth0_id', flat=True))
```

controlpanel/frontend/jinja2/datasource-detail.html
```diff
@@ -113,46 +76,11 @@ <h2 class="govuk-heading-m">Data access group</h2>
...
-
-    <div class="govuk-form-group">
-      <button class="govuk-button">Grant access</button>
-    </div>
-  </form>
-  {% else %}
-  <p class="govuk-body">(All available users already have access to this data source.)</p>
-  {% endif %}
+  {% if request.user.has_perm('api.create_users3bucket', bucket) and users_options|length %}
+    <a href="{{ url('grant-datasource-access', kwargs={'pk': bucket.pk}) }}"
```
xoen: Wasn't this a `POST` request? Wouldn't this make it a `GET` request?

controlpanel/frontend/views/datasource.py
```diff
@@ -213,13 +213,28 @@ def get_success_url(self):
...
+        member_ids = [member.user.auth0_id for member in access_group]
+        context['access_group'] = access_group
+        context['users_options'] = User.objects.exclude(
+            auth0_id__isnull=True,
```
xoen: can `User.auth0_id` even be null? I thought it was the primary key for that table?

controlpanel/frontend/views/datasource.py
```diff
@@ -213,13 +213,28 @@ def get_success_url(self):
...
     def get_form_kwargs(self):
         return FormMixin.get_form_kwargs(self)
 
+    def get_context_data(self):
+        context = super().get_context_data()
+        bucket = get_object_or_404(S3Bucket, pk=self.kwargs['pk'])
+        context['bucket'] = bucket
+        access_group = bucket.users3buckets.all().select_related('user')
+        member_ids = [member.user.auth0_id for member in access_group]
+        context['access_group'] = access_group
```
xoen: Does `access_group` need to go in the `context`? Is it used by the template, etc...?

controlpanel/frontend/jinja2/datasource-detail.html
andyhd: Instead of submitting an inline form (which was indeed a POST request), this link takes you to a separate form page

controlpanel/frontend/views/datasource.py
andyhd: It can and is - the `concourse` user has no `auth0_id`. This code was added a while back to prevent users accidentally adding the `concourse` user to their data access group. `auth0_id` is the primary key, but it is not prevented from being `null` - an old issue. I think `auth0_id` should not be the primary key, as having more than one user that is not logging in via Auth0 may be useful, but probably YAGNI

controlpanel/frontend/views/datasource.py
andyhd: Good catch 👍

controlpanel/frontend/views/datasource.py
xoen: Interesting 👌

PR #741, opened 2019-09-17
Added ability to "Upgrade" a tool

xoen:
Similarly to how a user can deploy an instance of a tool, this will allow users to upgrade a tool (the `helm` command is the same).

We only show the "Upgrade" button when a user's deployed tool version differs from the version in the DB (which is the version we consider stable and supported).

Also:
- fixed "Restart" button style/layout
- fixed `jsConfirm` module to work with buttons too (e.g. the "delete data source" button was a `<button>` with `.js-confirm` class but it didn't ask for confirmation)
- removed unused `delete_button` jinja macro
- commented redis server pidfile dependency as it doesn't seem to work on my machine (and probably on other platforms too)

**Ticket**: https://trello.com/c/rz6o5bug/101-users-can-upgrade-their-tools

### DIFF
--- a/controlpanel/api/models/tool.py
+++ b/controlpanel/api/models/tool.py
@@ -49,17 +49,24 @@ class ToolDeploymentManager:
         return tool_deployment
 
     def filter(self, **kwargs):
+        deployed_versions = {}
         user = kwargs.get('user')
         filter = Q(chart_name=None)  # Always False
         for deployment in cluster.list_tool_deployments(user):
-            _, version = deployment.metadata.labels["chart"].rsplit("-", 1)
+            chart_name, version = deployment.metadata.labels["chart"].rsplit("-", 1)
+            deployed_versions[chart_name] = version
             filter = filter | (
-                Q(chart_name=deployment.metadata.labels["app"])
+                Q(chart_name=chart_name)
                 # & Q(version=version)
             )
 
         tools = Tool.objects.filter(filter)
-        return [ToolDeployment(tool, user) for tool in tools]
+        results = []
+        for tool in tools:
+            outdated = tool.version != deployed_versions[tool.chart_name]
+            tool_deployment = ToolDeployment(tool, user, outdated)
+            results.append(tool_deployment)
+        return results
 
 
 class ToolDeployment:
@@ -73,11 +80,12 @@ class ToolDeployment:
 
     objects = ToolDeploymentManager()
 
-    def __init__(self, tool, user):
+    def __init__(self, tool, user, outdated=False):
         self._id_token = None
         self._subprocess = None
         self.tool = tool
         self.user = user
+        self.outdated = outdated
 
     def __repr__(self):
         return f'<ToolDeployment: {self.tool!r} {self.user!r}>'
--- a/controlpanel/frontend/consumers.py
+++ b/controlpanel/frontend/consumers.py
@@ -15,7 +15,9 @@ from controlpanel.api import cluster
 from controlpanel.api.cluster import (
     TOOL_DEPLOYING,
     TOOL_DEPLOY_FAILED,
+    TOOL_IDLED,
     TOOL_READY,
+    TOOL_UPGRADED,
 )
 from controlpanel.api.models import Tool, ToolDeployment, User
 from controlpanel.utils import PatchedAsyncHttpConsumer, sanitize_dns_label
@@ -132,6 +134,9 @@ class BackgroundTaskConsumer(SyncConsumer):
         """
         self.tool_deploy(message)
 
+        tool, user = self.get_tool_and_user(message)
+        update_tool_status(user, tool, TOOL_UPGRADED)
+
     def tool_restart(self, message):
         """
         Restart the named tool for the specified user
@@ -197,7 +202,7 @@ def wait_for_deployment(user, tool, deployment):
     while True:
         status = deployment.status
         update_tool_status(user, tool, status)
-        if status in (TOOL_DEPLOY_FAILED, TOOL_READY):
+        if status in (TOOL_DEPLOY_FAILED, TOOL_READY, TOOL_IDLED):
             return status
         sleep(1)
 
--- a/controlpanel/frontend/consumers.py
+++ b/controlpanel/frontend/consumers.py
@@ -126,6 +126,12 @@ class BackgroundTaskConsumer(SyncConsumer):
         else:
             log.debug(f"Deployed {tool.name} for {user}")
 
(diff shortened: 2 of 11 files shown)


controlpanel/api/models/tool.py
```diff
@@ -49,17 +49,24 @@ def create(self, *args, **kwargs):
...
-                Q(chart_name=deployment.metadata.labels["app"])
+                Q(chart_name=chart_name)
                 # & Q(version=version)
             )
 
         tools = Tool.objects.filter(filter)
-        return [ToolDeployment(tool, user) for tool in tools]
+        results = []
+        for tool in tools:
+            outdated = tool.version != deployed_versions[tool.chart_name]
```
andyhd: Not too comfortable with `!=`, although I experimented with a couple of semver python modules and they're all a bit unpleasant, so stick with this (at least for now)

controlpanel/api/models/tool.py
```diff
@@ -73,11 +80,12 @@ class ToolDeployment:
 
     objects = ToolDeploymentManager()
 
-    def __init__(self, tool, user):
+    def __init__(self, tool, user, outdated=False):
         self._id_token = None
         self._subprocess = None
         self.tool = tool
         self.user = user
+        self.outdated = outdated
```
andyhd: I think `outdated` would be better as a property or `is_outdated` method on `ToolDeployment`. Maybe pass in the `deployed_version` to the constructor instead?

controlpanel/frontend/consumers.py
```diff
@@ -126,6 +128,15 @@ def tool_deploy(self, message):
...
             log.debug(f"Deployed {tool.name} for {user}")
 
+    def tool_upgrade(self, message):
+        """
+        Upgrade simply means re-installing the helm chart for the tool
+        """
+        self.tool_deploy(message)
+
+        tool, user = self.get_tool_and_user(message)
+        update_tool_status(user, tool, TOOL_UPGRADED)
```
andyhd: I'm not keen on introducing new statuses - I don't think theres much value in differentiating between `Upgraded` and `Ready`.

controlpanel/frontend/jinja2/includes/forms.html
```diff
@@ -56,8 +56,3 @@ <h2 class="heading-medium govuk-fieldset__heading">
     {{ caller() }}
   </fieldset>
 {%- endmacro %}
-
-{% macro delete_button(label, confirm="") -%}
-  <input type="submit" class="button button-warning right js-confirm" value="{{
-  label }}" data-confirm-message="{{ confirm | default("Are you sure?") }}">
-{%- endmacro %}
```
andyhd: I thought this whole file was already deleted - why is this change here?

controlpanel/frontend/static/javascripts/modules/confirm.js
```diff
@@ -16,7 +16,9 @@ moj.Modules.jsConfirm = {
       }
     });
 
-    $(document).on('click', `input[type="submit"].${this.confirmClass}`, (e) => {
+    // works on any children of a `<form>` with `confirmClass` but it's
+    // usually used on `<input type="submit">` or `<button>`
+    $(document).on('click', `form > .${this.confirmClass}`, (e) => {
```
andyhd: Not sure if this is meant to be in this PR?

controlpanel/frontend/static/javascripts/modules/tool-status.js
```diff
@@ -38,10 +38,13 @@ moj.Modules.toolStatus = {
           break;
         case 'READY':
         case 'IDLED':
-          this.showActions(listener, ['open', 'restart', 'remove']);
+          this.showActions(listener, ['open', 'restart', 'upgrade', 'remove']);
+          break;
+        case 'UPGRADED':
+          this.showActions(listener, ['open']);
```
andyhd: Why doesn't an upgraded tool have `restart`? (and `remove` - not that it exists yet)

controlpanel/frontend/consumers.py
xoen: One good reason for being specific is that we know that we need to hide the "Upgrade" button once the tool is upgraded: Bear in mind we don't want to re-trigger a refresh and if we just recycle "ready" or "upgraded" we can't tell the difference.

controlpanel/frontend/jinja2/includes/forms.html
xoen: Maybe was deleted recently? It wasn't deleted when I made the change 😆

controlpanel/frontend/static/javascripts/modules/confirm.js
xoen: I almost used this module and found out there was a little bug (not working with buttons) and fixed it as it was a small fix.

controlpanel/frontend/jinja2/includes/forms.html
andyhd: I must have it in a branch somewhere 😕

controlpanel/frontend/consumers.py
andyhd: Ah, good point

controlpanel/frontend/static/javascripts/modules/tool-status.js
xoen: no `restart`: I imagine no point to restart something which was just upgraded - but I can add if you disagree
no `remove`: because it doesn't exist - although I can add if you think it's good 
 to leave for the future

Speaking of `restart`: does it make sense when something is already idled? I was tempted to remove but then I left as it was a bit out of scope and I thought there may have been a reason for it.

controlpanel/api/models/tool.py
xoen: I'm not keen on passing the `deployed_version` as I don't want to make this mandatory and when I make it optional the semantic of this gets weird: Does `deployed_version=None` means not deployed or out of date? I don't want to get into that as can lead to confusion/bugs.

But I like the idea of just having a `is_outdated` property which we set in the `ToolDeploymentManager` - it may not be the best solution but it would work for the time being.

controlpanel/frontend/static/javascripts/modules/tool-status.js
andyhd: They may want to restart if the tool fails to unidle.

xoen: Wouldn't that be a separate state tho? It would be probably be "unknown" or "failed" no? (maybe I miss something).

PR #742, opened 2019-09-20
User app buckets

andyhd:
* [Allow users to see which webapps they have access to](https://trello.com/c/OPPqoDzZ)
* [Show which apps have access on webapp datasource page](https://trello.com/c/edG7MzAl)

### DIFF
--- a/controlpanel/frontend/jinja2/base.html
+++ b/controlpanel/frontend/jinja2/base.html
@@ -61,7 +61,7 @@
 {% endblock %}
 
 {% block beforeContent %}
-  {% if not hide_nav %}
+  {% if not hide_nav and request.user.is_authenticated %}
   {{ navbar({
     "items": [
       {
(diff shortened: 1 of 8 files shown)


controlpanel/frontend/jinja2/base.html
```diff
@@ -81,13 +81,13 @@
         "active": page_name == "warehouse-datasource-list"
       },
       {
-        "hide": not request.user or not request.user.userapps or not request.user.userapps.filter(is_admin=True).exists(),
+        "hide": not request.user.users3buckets.filter(s3bucket__is_data_warehouse=False).exists(),
```
xoen: Looks like quite a bit of low level/logic for a template. But don't worry too much, just an observation, we can improve later.

andyhd: Don't be afraid of logic in templates 😉

xoen: you know I'm afraid of my shadow Andy 😁

PR #744, opened 2019-09-24
Add logs from elasticsearch to app detail page

andyhd:
Shows the app logs from the last 24 hours

### DIFF
--- a/controlpanel/api/elasticsearch.py
+++ b/controlpanel/api/elasticsearch.py
@@ -1,3 +1,4 @@
+from dateutil.parser import parse
 from django.conf import settings
 from elasticsearch import Elasticsearch
 from elasticsearch_dsl import Q, Search
@@ -7,7 +8,7 @@ from elasticsearch_dsl.query import Range
 def bucket_hits_aggregation(bucket_name, num_days=None):
     conn = Elasticsearch(hosts=settings.ELASTICSEARCH['hosts'])
 
-    s = Search(using=conn, index=settings.ELASTICSEARCH['index'])
+    s = Search(using=conn, index=settings.ELASTICSEARCH['indices']['s3-logs'])
 
     q1 = Q('term', **{'bucket.keyword': bucket_name})
     q2 = Q('terms',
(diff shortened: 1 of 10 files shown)


controlpanel/frontend/static/components/app-logs/macro.html
```diff
@@ -1,4 +1,15 @@
...
+  </div>
+  <p class="govuk-body">
+    <a href="{{ kibana_link(app) }}">View logs in Kibana</a>
+  </p>
+{% endmacro %}
+
+{% macro kibana_link(app) -%}
 {% filter replace(" ", "") -%}
 {% filter replace("\n", "") -%}
 https://kibana.services.{{env}}.mojanalytics.xyz/app/kibana#/discover?_g=(
```
r4vi: maybe we should start parameterising the domain now because it's going to change soon

andyhd: Good point

controlpanel/api/elasticsearch.py
```diff
@@ -23,3 +24,32 @@ def bucket_hits_aggregation(bucket_name, num_days=None):
                   size=100)
 
     return s.execute().aggregations.bucket_hits
+
+
+def app_logs(app, num_hours=None):
+    if not num_hours:
```
xoen: maybe I'm missing something, I'm a Python noon after all, but what's the difference between this and having the `num_hours` parameter have a default of `1`: `def app_logs(app, num_hours=1):`?

controlpanel/frontend/static/components/app-logs/macro.html
```diff
@@ -1,7 +1,18 @@
-{% macro app_logs(app) -%}
+{% macro app_logs(app, kibana_base_url) %}
+  <div class="app-logs">
+    <pre>{% for entry in app.get_logs() -%}
```
xoen: nice one :)

controlpanel/api/elasticsearch.py
andyhd: Yep, dunno how I ended up writing this

PR #745, opened 2019-09-25
Add css / babel watch scripts to package.json

r4vi:
## What

This commit allows you to run `npm run watch` which will keep compiling assets
when files are changed so you get quicker feedback when developing front-end
features.

I've added `corejs@2` to the requirements to get rid of a babel warning, and
delegated some of the makefile tasks to just call into `npm run` for css / babel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,20 +54,12 @@ node_modules:
 compilescss:
 	@echo
 	@echo "> Compiling SCSS..."
-	@./node_modules/.bin/node-sass \
-		--include-path node_modules/ \
-		-o static/ \
-		--output-style compact \
-		controlpanel/frontend/static/app.scss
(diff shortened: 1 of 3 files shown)


r4vi:
@andyhd I guess you use the makefile so you don't find it annoying but I run it from my IDE and needed this, @xoen what do you do for live asset rebuilding? Are you running the makefile too?

In the future I want to add browsersync.io too so we can see live changes to scss

xoen:
@r4vi yeah I've been running the thing using the `Makefile` just because it's the way it's documented (I think the `Makefile` grew a bit out of control) but I haven't worked on anything touching assets recently so probably not affected by this anyway.

It sounds like a good idea tho 👍

PR #747, opened 2019-09-26
Add back the link to the deployed App

xoen:
### What
This used to be part of the frontend but got lost in translation when removed node app.

This is adding it back:

[Imagine gif of Bender saying "I'm back, baby" here]

### How
The URL will be the host in the `Ingress` with the `repo` label set to the app repo (repo name without the URL).

As the `Ingress` resources don't live in the users' namespaces, users will not have sufficient permissions to list them, therefore this request is performed using the CP credentials (which can list Ingresses in that namespace). 
In order to do that I've tweaked the `api.kubernetes.KubernetesClient` class to receive a `use_cpanel_creds` argument (default is `False` as use of CP creds is discouraged and we should always aim at making requests to the k8s API using the user ID token).

### Small refactorings, etc...
- following conversation with @andyhd I've started to group the app-related operations performed on the "cluster" (k8s cluster but also AWS account) into the `api.cluster.App` class which wraps the `api.models.App`. Future plans include splitting this big module into smaller files (maybe rename it `external` to avoid further confusion with the k8s cluster?)
- `api.kubernetes.KubernetesClient` class now blows up if:
  - `id_token` is not provided (it optional) but `user_cpanel_creds=False` as this would cause the  up using the client to use the host (in-cluster or kube config) credentials
  - both `id_token` and `user_cpanel_creds=True` are provided as this is ambiguous

### Future refactorings
In the spirit of trying to keep PRs small and as focused as possible (and following @josh always wise advice), I've refrained myself from further simplifying the `api.kubernetes.KubernetesClient` class (by potentially removing logic to read session `id_token` and legacy RBAC flag) as it turned out to be more involved than expected - so I may do that separately

### Ticket
https://trello.com/c/eQqJ1FRw

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -213,7 +213,7 @@ class App:
         ).items
 
         if len(ingresses) != 1:
-            return ""
+            return None
 
         return f"https://{ingresses[0].spec.rules[0].host}"
 
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -8,7 +8,7 @@ from github import Github, GithubException
 from controlpanel.api.aws import aws, iam_arn
 from controlpanel.api.helm import helm
 from controlpanel.api.kubernetes import KubernetesClient
-
+from controlpanel.utils import github_repository_name
 
 log = logging.getLogger(__name__)
 
@@ -182,6 +182,8 @@ class App:
     Responsible for the apps-related interactions with the k8s cluster and AWS
     """
 
+    APPS_NS = "apps-prod"
+
     def __init__(self, app):
         self.app = app
 
@@ -199,6 +201,20 @@ class App:
         except Exception as e:
             if not is_ignored_exception(e):
                 raise e
+
+    def url(self, id_token):
+        repo_name = github_repository_name(self.app.repo_url)
+
+        k8s = KubernetesClient(id_token=id_token)
+        ingresses = k8s.ExtensionsV1beta1Api.list_namespaced_ingress(
+            self.APPS_NS,
+            label_selector=f"repo={repo_name}",
+        ).items
+
+        if len(ingresses) != 1:
+            return ""
+
+        return f"https://{ingresses[0].spec.rules[0].host}"
 
 
 class IAMRole:
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -134,16 +134,6 @@ def purge_user(user):
     helm.delete(f"init-user-{user.slug}")
 
 
-def create_app_role(name):
-    role = IAMRole(name)
-    role.save()
-
-
-@ignore_unwanted_exception
-def delete_app_role(name):
-    aws.delete_role(name)
-
-
 @ignore_unwanted_exception
 def create_policy(policy_name, path="/"):
     policy_document = {
@@ -185,6 +175,30 @@ def update_policy_roles(policy_arn, stored_roles):
 
     for role in (live_roles - stored_roles):
         aws.attach_policy_to_role(policy_arn, role)
+
+
+class App:
+    """
+    Responsible for the apps-related interactions with the k8s cluster and AWS
+    """
+
+    def __init__(self, app):
+        self.app = app
+
+    @property
+    def iam_role_name(self):
+        return f"{settings.ENV}_app_{self.app.slug}"
+
+    def create_iam_role(self):
+        role = IAMRole(self.iam_role_name)
+        role.save()
+
+    def delete_iam_role(self):
+        try:
+            aws.delete_role(self.iam_role_name)
+        except Exception as e:
+            if not is_ignored_exception(e):
+                raise e
 
 
 class IAMRole:
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -202,10 +202,11 @@ class App:
             if not is_ignored_exception(e):
                 raise e
 
-    def url(self, id_token):
-        repo_name = github_repository_name(self.app.repo_url)
+    @property
+    def url(self):
+        k8s = KubernetesClient(use_cpanel_creds=True)
 
-        k8s = KubernetesClient(id_token=id_token)
+        repo_name = github_repository_name(self.app.repo_url)
         ingresses = k8s.ExtensionsV1beta1Api.list_namespaced_ingress(
             self.APPS_NS,
             label_selector=f"repo={repo_name}",
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -36,8 +36,7 @@ class KubernetesClient:
         load_kube_config()
         config = kubernetes.client.Configuration()
 
-        if settings.ENABLED['k8s_rbac']:
-
+        if settings.ENABLED["k8s_rbac"] and not use_cpanel_creds:
             if id_token is None:
                 request = CrequestMiddleware.get_request()
                 if request and request.user and request.user.is_authenticated:
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -9,10 +9,30 @@ from controlpanel.kubeapi.views import kubernetes, load_kube_config
 class KubernetesClient:
     """
     Wraps kubernetes.client default configuration with currently logged-in
-    user's credentials
+    user's credentials unless `use_cpanel_creds=True`, in that case the
+    in-cluster (or ~/.kube/config) credentials are used.
+
+    **IMPORTANT**: Do not pass `use_cpanel_creds=True` unless strictly
+    necessary. The CP `ServiceAccount` has more privileges than normal users.
+    For most operations performed by a user use their ID token to make requests
+    to the k8s API.
     """
 
-    def __init__(self, id_token=None):
+    def __init__(self, id_token=None, use_cpanel_creds=False):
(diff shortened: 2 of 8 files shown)


controlpanel/api/cluster.py
```diff
@@ -124,17 +124,45 @@ def purge_user(user):
...
-        aws.delete_role(name)
-    except Exception as e:
-        if not is_ignored_exception(e):
-            raise e
+    def __init__(self, app):
+        self.app = app
+
+    @property
+    def iam_role_name(self):
+        return f"{settings.ENV}_app_{self.app.slug}"
```
andyhd: 👍

controlpanel/api/cluster.py
```diff
@@ -124,17 +124,45 @@ def purge_user(user):
...
+        k8s = KubernetesClient(use_cpanel_creds=True)
+
+        repo_name = github_repository_name(self.app.repo_url)
+        ingresses = k8s.ExtensionsV1beta1Api.list_namespaced_ingress(
+            self.APPS_NS,
+            label_selector=f"repo={repo_name}",
+        ).items
+
+        if len(ingresses) != 1:
+            return ""
```
andyhd: Returning an empty string is fine as it evaluates to `False` in `{% if app_url %}`, but I think it makes more sense to return `None`

andyhd: Also, if there are more than one ingress, shouldn't we return one of them? It will be a valid URL, won't it?

controlpanel/api/kubernetes.py
```diff
@@ -9,10 +9,30 @@
...
+                "be careful when using it)"
+            )
+
+        if id_token and use_cpanel_creds:
+            raise ValueError(
+                "id_token and use_cpanel_creds can't be used together. "
+                "Remember: avoid using the Control Panel credentials to talk with "
+                "the k8s API unless stricly necessary."
+            )
+
```
andyhd: I think you've forgotten to push some code here

controlpanel/api/models/app.py
```diff
@@ -72,10 +72,10 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
         if is_create:
-            cluster.create_app_role(self.iam_role_name)
+            cluster.App(self).create_iam_role()
```
andyhd: I like this 👍

doc/running.md
```diff
@@ -113,3 +113,13 @@ Go to http://localhost:8000/
...
+
+or directly using `pytest`:
+
+```sh
+DJANGO_SETTINGS_MODULE=controlpanel.settings.test pytest --color=yes
+```
+
+**NOTE** Set the `DJANGO_SETTINGS_MODULE` is important or otherwise you
+may accidentally run the tests with the `development` settings with
+unpredictable results.
```
andyhd: The `DJANGO_SETTINGS_MODULE` used by `pytest` is configured in `/setup.cfg`, you shouldn't have to set it manually - is that not working?

andyhd: Also, you don't need to pass `--color=yes`, colours are automatically enabled if the terminal is capable of showing them

tests/api/cluster/test_app.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_app_url_when_no_ingresses(k8s_client, app, app_model):
+    k8s_client.ExtensionsV1beta1Api.list_namespaced_ingress.return_value.items = []
+
+    assert app.url == ""
+    k8s_client \
+        .ExtensionsV1beta1Api \
+        .list_namespaced_ingress \
+        .assert_called_once_with("apps-prod", label_selector="repo=test-repo")
```
andyhd: I'm pretty sure you can and should remove these `\`s - but use [Black](https://black.readthedocs.io/en/stable/index.html) as the source of truth on formatting

andyhd: You can break the line after `(` like:
```python
    k8s_client.ExtensionsV1beta1Api.list_namespaced_ingress.assert_called_once_with(
        "apps-prod",
        label_selector="repo=test-repo",
    )
```
The long line is <88 chars, which is the default max line-length in Black.

tests/api/cluster/test_app.py
```diff
@@ -0,0 +1,74 @@
...
+def test_app_url_when_single_ingress_found(k8s_client, app, app_model):
+    ingress = MagicMock(name="Ingress")
+    ingress.spec.rules = [MagicMock(name="Rule", host="test-app.example.com")]
+    k8s_client.ExtensionsV1beta1Api.list_namespaced_ingress.return_value.items = [ingress]
+
+    assert app.url == "https://test-app.example.com"
+    k8s_client \
+        .ExtensionsV1beta1Api \
+        .list_namespaced_ingress \
+        .assert_called_once_with("apps-prod", label_selector="repo=test-repo")
```
andyhd: You should replace these last three tests with a single parameterised test. Eg:
```python
mock_ingress = MagicMock(name="Ingress")
mock_ingress.spec.rules = [MagicMock(name="Rule", host="test-app.example.com")]

@pytest.mark.parameterize(
    'ingresses, expected_url',
    [
        ([], "",),
        (["ing_1", "ing_2"], ""),
        ([mock_ingress], "https://test-app.example.com"),
    ],
    ids=[
        "no-ingresses",
        "multiple-ingresses",
        "single-ingress",
    ]
)
def test_app_url(k8s_client, app, app_model, ingresses, expected_url):
    list_ingresses = k8s_client.Extensions.V1beta1Api.list_namespaced_ingress
    list_ingresses.return_value.items = ingresses
    assert app_url == expected_url
    list_ingresses.assert_called_once_with("apps-prod", label_selector="repo=test-repo")
```

controlpanel/api/cluster.py
xoen: I'll have a think about the `None` but it may be better as `None` conveys better the "no URL" meaning.

xoen: Regarding the multiple ingress I'd stick to return `""`/`None` as there should be only a single `Ingress` per app so if two are found we can't be 100% sure which URL is the right one.

I'd rather have the user come to us and ask why the URL is not shown and we then investigate than have shown a potentially wrong URL, the user freaks out, it comes to us anyway and we freak out as well.

It shouldn't happen anyway but I prefer to not pick a "random" URL when there is potential ambiguity.

controlpanel/api/kubernetes.py
xoen: Oh god, you're absolutely right! Superstar for catching this @andyhd 🌟❤️ 

I was tweaking this class and eventually decided to move the refactoring experiment into a separate PR but I thrown away the feature with the dirty water 🤦‍♂ 

The missing code is simply a `and not use_cpanel_creds` added to the `if`:

```python
        if settings.ENABLED['k8s_rbac'] and not use_cpanel_creds:
            if id_token is None:
                # ...
```

as in that case we don't want to use the `id_token`.

I'll dig this out or add again. Amazing, thanks again for spotting this Andy ❤️

doc/running.md
xoen: When I got this problem I found out that yes this is configured in `./setup.cfg` but in my case I also had that `DJANGO_SETTINGS_MODULE` set in my environment variable which unfortunately (in this case!) takes precedence against the `setup.cfg`.

This is by design and makes a lot of sense but I imagined it could happen to others following this documentation as one of the first things at the top of the page is to set `DJANGO_SETTINGS_MODULE` to `*.development`. I just thought having a simple note here could save someone else some time in the future.

xoen: Regarding `--color=yes` I'll remove, I just copied whatever was in the `Makefile`.

tests/api/cluster/test_app.py
xoen: @r4vi (who was there when I was writing this) told me to tell you I run this through black :trollface: 

I'll play with it. (no wonder Go's autoformatting is everyone's favourite feature 😎)

tests/api/cluster/test_app.py
xoen: Nice. I noticed at the time that there was some repetition in these tests but I got frustrated at pytest and when it worked and it was nice enough (you should have seen the previous 2 drafts of this test 😆) I left it.

Thanks for the hint ❤️ , I'll play with it 👍

controlpanel/api/cluster.py
andyhd: In my opinion, a working but "wrong" URL would be better than a missing one, but OK

tests/api/cluster/test_app.py
xoen:
```pytest
test_app_url has 'parameterize' mark, spelling should be 'parametrize'
```

blimey, "parametrize" that's a difficult word 🤣

controlpanel/api/kubernetes.py
```diff
@@ -9,15 +9,34 @@
...
         load_kube_config()
         config = kubernetes.client.Configuration()
 
-        if settings.ENABLED['k8s_rbac']:
-
+        if settings.ENABLED["k8s_rbac"] and not use_cpanel_creds:
```
xoen: @andyhd this is the bit that somehow got lost in previous commits set.

xoen:
@andyhd have another look if you wish, I've rebased and added back the missing code using the `use_cpanel_creds` argument.

Also:
- `cluster.App.url` returns `None` when can't determine URL
- refactored the tests to use parametrize 🏆 
- and formatted some of the code with black

PR #748, opened 2019-09-26
Support bulk-delete for app customers

r4vi:
## What
This commit updates the webapp-detail template to apply the selected rows module
to the customers table.

image

## How to review

1. add a bunch of users to an app
2. select a bunch of rows
3. press the `delete all selected` button

https://trello.com/c/2f2fMS0a/273-2-bulk-delete-for-app-users

### DIFF
--- a/controlpanel/frontend/jinja2/webapp-detail.html
+++ b/controlpanel/frontend/jinja2/webapp-detail.html
@@ -69,9 +69,17 @@
   </h2>
   <form method="post" action="{{ url("remove-app-customer", kwargs={"pk": app.id}) }}">
     {{ csrf_input }}
-    <table class="govuk-table">
+    <button class="govuk-button hmcts-button--secondary right selectable-rows__enable-on-selections"
+            type="submit"
+            name="submit"
+            disabled="disabled"
+            value="submit">
+      Remove all selected customers
+    </button>
+    <table class="govuk-table selectable-rows">
       <thead class="govuk-table__head">
         <tr class="govuk-table__row">
+          <th class="govuk-table__header"></th>
           <th class="govuk-table__header">Customer email</th>
           <th class="govuk-table__header">
             <span class="govuk-visually-hidden">Actions</span>
@@ -81,6 +89,7 @@
       <tbody class="govuk-table__body">
         {% for customer in customers %}
         <tr class="govuk-table__row">
+          <td class="govuk-table__cell">{{ row_selector('customer', customer.user_id) }}</td>
           <td class="govuk-table__cell">{{ customer.email }}</td>
           {% if request.user.has_perm('api.remove_app_customer', app) %}
           <td class="govuk-table__cell">
--- a/controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
+++ b/controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
@@ -1,24 +1,92 @@
+/**
+ * Module to allow selecting entire rows in a table instead of using a small checkbox.
+ *
+ * It will disable any in-row actions buttons once a selection is made, but will re-enable them once
+ * all the selections are cleared.
+ *
+ * It can optionally support enabling a (global)
+ * button(s) that lives outside the table.
+ * How to use:
+ *  1. Add the class "selectable-rows" to your <table>
+ *  2. For any button you want to enable when there are selections add the class "selectable-rows__enable-on-selections"
+ *  3. Use the macro `row_selector` to put insert an invisible input box in the first column of your table
+ * The module will automatically disable any action buttons that are in side the row.
+ *
+ * WARNING: This module doesn't support more than 1 table on the same page
+ * @type {{init(): void, onSelectionChanged(*): void, baseClass: string, bindEvents(): void}}
+ */
+
 moj.Modules.selectable_rows = {
-  selector: '.selectable-rows tr',
+  baseClass: "selectable-rows",
 
   init() {
-    if (document.querySelector(this.selector)) {
+    this.tableRowSelector = `.${this.baseClass} tr`;
+    this.tableRowButtonSelector = `${this.tableRowSelector} button`;
+    this.selectionClass = `${this.baseClass}__row-selected`;
+    this.enableOnSelectionSelector = `.${this.baseClass}__enable-on-selections`;
+    this.selectionChangedEventName = `${this.baseClass}:rowSelectionChanged`;
+    if (document.querySelector(this.tableRowSelector)) {
       this.bindEvents();
     }
   },
 
+  /**
+   * Runs when rows are selected/deselected
+   * @param event
+   */
+  onSelectionChanged(event) {
+    const rowButtons = document.querySelectorAll(this.tableRowButtonSelector);
+    const enableOnSelectionButtons = document.querySelectorAll(
+      this.enableOnSelectionSelector
+    );
+    if (event.detail.rowsSelected.length > 0) {
+      // some rows are selected
+      rowButtons.forEach(e => (e.disabled = true));
+      enableOnSelectionButtons.forEach(e => e.disabled = false);
+    } else {
+      rowButtons.forEach(e => (e.disabled = false));
+      enableOnSelectionButtons.forEach(e => e.disabled = true);
+    }
+  },
+
   bindEvents() {
-    document.querySelectorAll(this.selector).forEach(row => {
+    window.addEventListener(this.selectionChangedEventName, e =>
+      this.onSelectionChanged(e)
+    );
+    document.querySelectorAll(this.tableRowSelector).forEach(row => {
       row.addEventListener("click", event => {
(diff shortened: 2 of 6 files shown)


r4vi:
TODO: tests???

r4vi:
I've been thinking about how to setup tests for components and that's probably a whole new ticket rather than adding here. It will require adding a javascript test runner like `jest` - I think it's worth doing but shouldn't block this ticket

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
```diff
@@ -1,24 +1,92 @@
...
+ *
+ * It can optionally support enabling a (global)
+ * button(s) that live outside the table.
+ * How to use:
+ *  1. Add the class "selectable-rows" to your <table>
+ *  2. For any button you want to enable when there are selections add the class "selectable-rows__enable-on-selections"
+ *  3. Use the macro `row_selector` to put insert an invisible input box in the first column of your table
+ * The module will automatically disable any action buttons that are in side the row.
+ *
+ * WARNING: This module doesn't support more than 1 table on the same page
```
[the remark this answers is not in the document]
r4vi: it would disable / enable buttons on both tables. Just needs more work for it to support that

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
```diff
@@ -1,24 +1,92 @@
...
+   * @param event
+   */
+  onSelectionChanged(event) {
+    const rowButtons = document.querySelectorAll(this.tableRowButtonSelector);
+    const enableOnSelectionButtons = document.querySelectorAll(
+      this.enableOnSelectionSelector
+    );
+    if (event.detail.rowsSelected.length > 0) {
+      // some rows are selected
+      rowButtons.forEach(e => (e.disabled = true));
```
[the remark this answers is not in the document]
r4vi: brain-fart

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
```diff
@@ -1,24 +1,92 @@
...
-  selector: '.selectable-rows tr',
+  baseClass: "selectable-rows",
 
   init() {
-    if (document.querySelector(this.selector)) {
+    this.tableRowSelector = `.${this.baseClass} tr`;
+    this.tableRowButtonSelector = `${this.tableRowSelector} button`;
+    this.selectionClass = `${this.baseClass}__row-selected`;
+    this.enableOnSelectionSelector = `.${this.baseClass}__enable-on-selections`;
+    this.selectionChangedEventName = `${this.baseClass}:rowSelectionChanged`;
```
xoen: it doesn't matter and it's merged but maybe we could lose the `row` for brevity and consistency with the `onSelectionChanged` event handler :)

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
```diff
@@ -1,24 +1,92 @@
...
+      enableOnSelectionButtons.forEach(e => e.disabled = false);
+    } else {
+      rowButtons.forEach(e => (e.disabled = false));
+      enableOnSelectionButtons.forEach(e => e.disabled = true);
+    }
+  },
+
   bindEvents() {
-    document.querySelectorAll(this.selector).forEach(row => {
+    window.addEventListener(this.selectionChangedEventName, e =>
```
xoen: I may miss something but couldn't just pass `this.onSelectionChanged` as event listener instead of the anonymous function which calls it? the event loop would call it anyway passing the event right?

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
```diff
@@ -1,24 +1,92 @@
...
+      this.onSelectionChanged(e)
+    );
+    document.querySelectorAll(this.tableRowSelector).forEach(row => {
```
xoen: Probably not or not meaningful anyway but is there any performance hit by creating a bunch of anonymous functions in a loop here? Possibly fascinating but not important.

A more appropriate question would be, any reason to not have this as a named function (something like `onRowClicked()`)? Like we have `onSelectionChanged()`.

Just curious, nothing wrong with this of course :)

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
r4vi: I think I had it like that to begin with but then I think `this` was bound to something else so using the => arrow syntax to keep `this` the same

xoen: interesting, that is so sneaky!

r4vi: more info: https://maximdenisov.gitbooks.io/you-don-t-know-js/content/scope_&_closures/lexical-this.html

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
```diff
@@ -1,24 +1,92 @@
...
+          "input[type='checkbox'].row-selector"
+        );
+
+        // user clicks on the button, don't select the row
+        if (event.target.tagName === "BUTTON") {
+          return true;
+        }
         if (checkbox) {
+          // user clicks somewhere on the row that isn't the checkbox itself
+          if (event.target !== checkbox) {
```
xoen: I haven't tried it but it sounds like a good idea.
Is there something similar to `for` for `checkbox`es? it may be more predictable than handling this manually.

I was thinking about possible ways this could break if the raw contained links but in the case of customers that shouldn't be the case.

(what I mean by "break" is something like a user clicks on a link that opens a tab, the selection changes without him noticing, or something like that. But probably not a problem)

controlpanel/frontend/views/app.py
```diff
@@ -248,9 +247,9 @@ class RemoveCustomer(UpdateApp):
 
     def perform_update(self, **kwargs):
         app = self.get_object()
-        user_id = self.request.POST.get('customer')
-        app.delete_customers([user_id])
-        messages.success(self.request, "Successfully removed customer")
+        user_ids = self.request.POST.getlist('customer')
```
xoen: not sure what's the convention but maybe name this `customers` as it's a list really? 🤷‍♂

controlpanel/frontend/jinja2/webapp-detail.html
```diff
@@ -81,6 +89,7 @@ <h2 class="govuk-heading-m">
       <tbody class="govuk-table__body">
         {% for customer in customers %}
         <tr class="govuk-table__row">
+          <td class="govuk-table__cell">{{ row_selector('customer', customer.user_id) }}</td>
```
xoen: this `row_selector` was already there right? I don't see in the diff but I can see it in the repo.

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
xoen: I doubt this will make the cut in a newer edition of "Javascript: The Good Parts".

xoen: I've encountered this before but I still find it sneaky. "arrow functions as pretty much like normal functions...but WAIT, not quite".

In a way the weirdness factor is caused by the fact the behaviour of the old school functions is weird and the new ones are behaving in a better, more predictable way.

Oh well, it's JS 🤷‍♂

controlpanel/frontend/jinja2/webapp-detail.html
r4vi: yeah @andyhd already started on this stuff - I just finished it

controlpanel/frontend/views/app.py
r4vi: it's not a list, it's like the `customer` form param being submitted multiple times. `.get()` will just take the first one, `.getlist()` will give you every time it has been submitted in the current request. I think the name is right here.

xoen: happy for you to leave it as it 👍 (still looks like `customer` contains the customer IDs to bulk delete, submitted in the POST request being handled 🤷‍♂)

andyhd: @xoen `customer` is the name of the form field. There is more than one field, but each field's name is `customer`. It wouldn't make sense to call each field `customers`, as you only populate the field with a single customer user ID.

controlpanel/frontend/static/components/selectable-rows/selectable-rows.js
andyhd: There's nothing like `<label for="">` for a table row, this has to be done manually.

andyhd: Also, there must never be a link that opens a tab, or a selection that changes without user input - both definitely against UXD rules.

PR #749, opened 2019-09-27
Fix template error on Group page

s-block:
Fixes template error


andyhd:
You should add `context_object_name = 'policy'` to `IAMManagedPolicyDetail` instead

PR #750, opened 2019-09-30
Fix for KubernetesClient error when no id_token provided

xoen:
That check was in place for some refactoring I'm working on (which
should remove the code that reads the ID token from request when not
passed).

The problem is that currently some of the code don't pass the `id_token`
and let the class read from current request.

I'm commenting that check for now.

### DIFF
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -19,12 +19,12 @@ class KubernetesClient:
     """
 
     def __init__(self, id_token=None, use_cpanel_creds=False):
-        if not use_cpanel_creds and not id_token:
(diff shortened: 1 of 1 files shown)


xoen:
@s-block this is the problem, as some code at the moment relies on the `KubernetesClient` class to read the `id_token` from the request when not provided.

I plan to remove that logic and always pass the `id_token` from the outside of this class but I must have left this check here from the refactoring branch - I'm commenting for now 😞

xoen:
Thanks @s-block  🌟

andyhd:
You shouldn't commit commented-out code (I know I do it sometimes 😞). No need to delete this now, of course.

PR #754, opened 2019-10-02
Don't get App URL until we fix the `KubernetesClient` bug

xoen:
Sometimes users get a `401` from the Kubernetes API. This seems
to be caused by a bug/race condition in the way the `kubernetes` module
loads the configuration (this is our best theory so far)

It's annoying for users so avoiding this until we fix the problem.

Part of ticket: https://trello.com/c/D4bpabeD

### DIFF
--- a/controlpanel/frontend/views/app.py
+++ b/controlpanel/frontend/views/app.py
@@ -66,7 +66,8 @@ class AppDetail(LoginRequiredMixin, PermissionRequiredMixin, DetailView):
         context = super().get_context_data(**kwargs)
         app = self.get_object()
 
-        context["app_url"] = cluster.App(app).url
+        # Temporarily hiding App URL until we fix the `KubernetesClient` bug causing the 401
+        context["app_url"] = None # cluster.App(app).url
         context["admin_options"] = User.objects.filter(
             auth0_id__isnull=False,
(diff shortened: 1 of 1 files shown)


controlpanel/frontend/views/app.py
```diff
@@ -66,7 +66,8 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         app = self.get_object()
 
-        context["app_url"] = cluster.App(app).url
+        # Temporarily hiding App URL until we fix the `KubernetesClient` bug causing the 401
+        context["app_url"] = None # cluster.App(app).url
```
xoen: @andyhd not that familiar with Jinja but I'm assuming that setting `app_url` to `None` would mean the `if` in the template would evaluate `False` and don't show anything in the block, right? (sounds like a stupid question but worth double-checking)

```jinga
  {% if app_url %}
  <!--  DISPLAY URL HERE  -->
  {% endif %}
```

andyhd: Yes, the code in `{% ... %}` tags is evaluated as Python

andyhd: This looks fine

PR #756, opened 2019-10-03
Fix for kubernetes race condition on configuration

xoen:
### Problem
Users were getting some strange `401` when trying to get the App URL.
This wasn't happening all the time and not in all environments.

The logic to find the URL for an app involved using the CP `ServiceAccount` credentials instead of the user's `id_token`.

We think this weird behaviour was caused by some sort of race condition on the credentials in the config.

### Changes overview
1. [moved function](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/93d3eafea11a97dc8cc4674192b6954283daccc8) to load kubernetes config (`load_kube_config()`) from `controlpanel.kubeapi.views` to `controlpanel.api.kubernetes` module as it's where you'd expect this to be.
2. [changed](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/321272bb1bdc9476e0d695d9e9ee98bf8edec9e8) `load_kube_config()` to `get_config()`. Client code is really interested in getting a [`Configuration`](https://github.com/kubernetes-client/python/blob/2b3fe30d24ed2b74a9ba9830dc37943951860c01/kubernetes/client/configuration.py#L41) instance to create a useable k8s client. All `load_kube_config()` calls were followed by a "get config" anyway (`kubernetes.client.Configuration()`)
3. [fixed credentials race condition in k8s configuration](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/e5aeb0c83ed3edce6833549ee61f94d7eae57eff), this is achieved by making sure the `config.api_key` and `config.api_key_prefix` are not shared with other instances of `kubernetes.client.Configuration` (see below for reasoning)

### Reason
The `kubernetes` module way of retrieving a configuration is a bit strange. You call [`load_incluster_config()`](https://github.com/kubernetes-client/python-base/blob/474e9fb32293fa05098e920967bb0e0645182d5b/config/incluster_config.py#L89) (or `load_kube_config()` locally) which loads the config but it doesn't return it, [just sets it](https://github.com/kubernetes-client/python-base/blob/474e9fb32293fa05098e920967bb0e0645182d5b/config/incluster_config.py#L81) into a sort `Configuration` singleton.

This [makes a shallow copy of the configuration](https://github.com/kubernetes-client/python/blob/2b3fe30d24ed2b74a9ba9830dc37943951860c01/kubernetes/client/configuration.py#L38), but unfortunately this is not enough as it contains other dictionaries which would therefore be shared across all configuration instances. That is the problem.

Specifically the authentication settings are stored into the [`api_key`](https://github.com/kubernetes-client/python/blob/2b3fe30d24ed2b74a9ba9830dc37943951860c01/kubernetes/client/configuration.py#L57) and [`api_key_prefix`](https://github.com/kubernetes-client/python/blob/2b3fe30d24ed2b74a9ba9830dc37943951860c01/kubernetes/client/configuration.py#L60) fields are also dictionaries.

### Solution
The proposed solution is to copy these two dictionaries so that they're not shared.
This means that when we [set the `api_key` to the user `id_token`](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/5a6b6c06eb0ec3864f30326bbd135a38bf94bdbf/controlpanel/api/kubernetes.py#L46) this should not overwrite any other `api_key` in other instances of `Configuration`.

### Ticket/Resources
Ticket: https://trello.com/c/D4bpabeD
Issue: https://github.com/kubernetes-client/python/issues/932

### DIFF
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import inspect
 import kubernetes
 import os
@@ -22,6 +23,13 @@ def get_config():
         kubernetes.config.load_kube_config()
 
     config = kubernetes.client.Configuration()
+
+    # The deepcopy is to avoid a race conditions on the credentials keys
+    #
+    # See: https://github.com/kubernetes-client/python/issues/932
+    config.api_key_prefix = deepcopy(config.api_key_prefix)
+    config.api_key = deepcopy(config.api_key)
+
     return config
 
 
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -23,7 +23,9 @@ def get_config():
 
     config = kubernetes.client.Configuration()
 
-    # The deepcopy is to avoid a race conditions on the credentials keys
+    # A deepcopy of the configuration is used to avoid a race condition
+    # caused by subsequent calls to Configuration() reusing a singleton
+    # datastructure
     #
     # See: https://github.com/kubernetes-client/python/issues/932
     config.api_key_prefix = deepcopy(config.api_key_prefix)
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -1,9 +1,24 @@
 import inspect
+import kubernetes
+import os
 
 from crequest.middleware import CrequestMiddleware
 from django.conf import settings
 
-from controlpanel.kubeapi.views import kubernetes, load_kube_config
+# This patch fixes incorrect base64 padding in the Kubernetes Python client.
+# Hopefully it will be fixed in the next release.
+from controlpanel.kubeapi import oidc_patch
+
+
+def load_kube_config():
+    """
+    Load Kubernetes config. Avoid running at import time.
+    """
+
+    if 'KUBERNETES_SERVICE_HOST' in os.environ:
+        kubernetes.config.load_incluster_config()
+    else:
+        kubernetes.config.load_kube_config()
 
 
 class KubernetesClient:
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -11,13 +11,12 @@ from django.conf import settings
 from controlpanel.kubeapi import oidc_patch
 
 
-
 def get_config():
     """
     Load Kubernetes config. Avoid running at import time.
     """
 
-    if 'KUBERNETES_SERVICE_HOST' in os.environ:
+    if "KUBERNETES_SERVICE_HOST" in os.environ:
         kubernetes.config.load_incluster_config()
     else:
         kubernetes.config.load_kube_config()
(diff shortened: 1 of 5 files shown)


xoen:
FYI: I need to fix some of the tests that [(now) mock the wrong thing](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/5a6b6c06eb0ec3864f30326bbd135a38bf94bdbf/tests/kubeapi/test_permissions.py#L13) as I've moved the code into `controlpanel.api.kubernetes`.

(tests also currently fail in CircleCI but because of some weird docker build problem)

xoen:
This should fix the test failure: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/756/commits/7f820c018086c7f3618062065c4c29a004fc187c

I see CircleCI is still having some problem with the Dockerfile, something must have changed. I'll have a look in a bit.

xoen:
Hopefully bump `python3-dev=3.7.3-r0` => `python3-dev=3.7.4-r0` will fix it: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/756/commits/2f01759293c955d2a1457bac804b5bfc6677d5e5

controlpanel/api/kubernetes.py
```diff
@@ -1,9 +1,35 @@
...
+    """
+
+    if "KUBERNETES_SERVICE_HOST" in os.environ:
+        kubernetes.config.load_incluster_config()
+    else:
+        kubernetes.config.load_kube_config()
+
+    config = kubernetes.client.Configuration()
+
+    # The deepcopy is to avoid a race conditions on the credentials keys
```
andyhd: Can you expand this comment a bit? Something like `A deepcopy of the configuration is used to avoid a race condition caused by subsequent calls to Configuration() reusing a singleton datastructure`

controlpanel/kubeapi/views.py
```diff
@@ -1,30 +1,16 @@
...
-import os
 from urllib.parse import urljoin
 
 from django.conf import settings
 from django.core import exceptions
 from django.views.decorators.csrf import csrf_exempt
 from djproxy.views import HttpProxy
-import kubernetes
 
+from controlpanel import api
```
andyhd: Would this not be better as `from controlpanel.api import kubernetes`?

controlpanel/kubeapi/views.py
```diff
@@ -1,30 +1,16 @@
 from functools import wraps
```
andyhd: I know it's not connected to this PR, but this import seems unused

controlpanel/api/kubernetes.py
xoen: 👍

controlpanel/kubeapi/views.py
xoen: Nice catch 👍

controlpanel/kubeapi/views.py
xoen: As discussed, I've imported it this way to avoid ambiguity so that it's not confused with the official `kubernetes` module.

PR #755, opened 2019-10-02
Add `jest` for JavaScript unit testing

r4vi:
This allows us to unit test the components. I have added tests for the
`selectable-rows` component to serve as an example for future components

## How to test:
`npm run test`

https://trello.com/c/S0rQP0Kb/328-set-up-frontend-testing-for-js-components

PR #757, opened 2019-10-03
Bugfix: test files in app.js bundle

r4vi:
This commit excludes the test files from being bundled by babel. This bug was
introduced by https://github.com/ministryofjustice/analytics-platform-control-panel/pull/755


r4vi:
good spot @sh31

jest.config.js
```diff
@@ -176,9 +176,10 @@ module.exports = {
   // transform: null,
 
   // An array of regexp pattern strings that are matched against all source file paths, matched files will skip transformation
-  // transformIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  transformIgnorePatterns: [
+    "/node_modules/",
```
xoen: You are the JS expert and lover but is it OK to ignore transpiling `node_modules`? (genuine question, not trolling you)

jest.config.js
```diff
@@ -176,9 +176,10 @@ module.exports = {
...
+    "\\.test\\.js"
```
xoen: is that how you escape in this context? I would have thought that a `.` is escaped `\.` and a `\` would be escaped with `\\` 🤷‍♂

andyhd: these are regex patterns and the `.` is escaped because we want to match a literal `.`, not any single character. then the `\.` has to be escaped because `\` is an escape character in normal javascript strings.

andyhd: i thought you could write regex literals in javascript, so maybe this could be `/\.test\.js/` instead?

xoen: Apologies for more stupid question regarding his: If the `\` is also escaped doesn't this mean the regex wouldn't it be treated as literal `\`?

If we want to much any string containing the `.test.js` (my understanding) and because we need to escape the `.` (as in general is a wildchar meaning any single character) wouldn't the regex be `\.test\.js`? [[rubular link]](https://rubular.com/r/vhdhJbnoySTyhF)

I'm not saying this doesn't work, I'm just curious ☺️

r4vi: could be, but I pasted it and my IDE auto-escaped it for the language so :man_shrugging:

jest.config.js
r4vi: this is just for jest, and tests are running in node so nothing really needs transpiling. You do want this transpiled in the "babel" step so `node_modules` aren't excluded in the other file (package.json -> babel.ignore)

jest.config.js
r4vi: @xoen I did the same as you and figured out the regex i need is: `\.test\.js`, then I pasted it and my IDE found the the language of the file I was pasting it into was .js and escaped it properly for that to end up with `"\\.test\\.js"`

r4vi: life is too short to think about this stuff

xoen: Are you kidding me? Regular expressions conversations are my favourite past time 🤣

andyhd: @xoen, no. when the javascript file is parsed by node, it reads `\\.` and generates a string object containing the characters `\.` - the regular expression parser then interprets this as 'literal `.`', rather than 'wildcard single character'.
If the javascript string was `\.`, then javascript would interpret it as `.` and then the regular expression parser would match files named `xtestxjs`.

xoen: Thanks @andyhd - OMG that's strange. I didn't realise node was also parsing this before, hence the double escape. Thanks very useful.

PR #759, opened 2019-10-04
Removed legacy `ENABLE_K8S_RBAC` flag

xoen:
This was useful when we enabled RBAC in the k8s cluster and we
transitioned to it but it's not necessary anymore as this is always
enabled now.

It also works fine when running the app locally and tests pass so no
reason to keep this.

The main reason to remove this is that it will enable us to simplify
the `KubernetesClient` class and make it more testable (as we'd not
need to care about testing cases when this flag is enabled or disabled)

### DIFF
--- a/controlpanel/api/kubernetes.py
+++ b/controlpanel/api/kubernetes.py
@@ -63,7 +63,7 @@ class KubernetesClient:
 
         config = get_config()
 
-        if settings.ENABLED["k8s_rbac"] and not use_cpanel_creds:
+        if not use_cpanel_creds:
             if id_token is None:
                 request = CrequestMiddleware.get_request()
                 if request and request.user and request.user.is_authenticated:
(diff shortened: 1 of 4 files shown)


xoen:
Could you try to run this branch locally and see if it works for you?
- It works on my machine when run directly (I've clicked around and it didn't blow up)
- I run the tests with `make docker-test` and they pass
- I also run the tests with `DJANGO_SETTINGS_MODULE=controlpanel.settings.test pytest` and they pass

I tried to run with `make docker-run` but it blew up with weird gunicorn errors but I don't think that is related :dumpsterfire: 😁

r4vi:
works for me locally

xoen:
Awesome 🏆 , thanks Ravi ☺️

andyhd:
oops too slow!

PR #761, opened 2019-10-08
KubernetesClient related refactorings

xoen:
### What

- Simplified `KubernetesClient` [to not read the user `id_token` from HTTP request](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/5b8fd7b17c2c6975c405744a12d1332b0283d0ee) (this can't be always done, e.g. if in the context of a background worker)
- Above allowed me to write some tests [[1]](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/761/files#diff-7195327eb3862074d88addee3f580b5bR1) [[2]](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/761/commits/5b97d498f23d1d6120c5d6c385969f5404fa9208) for this class
- Remove the implicit read of the `id_token` required me to be sure that all places where a `KubernetesClient` is instantiated (and their client code) always pass the `id_token` when performing an operation against the kubernetes API. This is done in this two commits:
  - [change to `cluster. restart_tool_deployment()`](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/9a5921184c15d244ed80891a9e28101ccdeea2aa)
  - [change to the rest of `cluster` functions](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/e4f4c1cf3604cc73db7c46d1f714baef4ef28699) (functions using `cluster. list_tool_deployments()` directly or indirectly)
- also fixed use of [unimported `helm.HelmError` in `cluster` module](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/ed00f9be2d9bee0ec3fc17af5ae0146e93a1d61f)

### Design rationale
I've avoided moving too much around to keep this PR focused on simplifying/testing the `KubernetesClient` class.

One key design decision I've taken is to always explicitly passing the `id_token` when performing an operation against the k8s cluster. 
Potentially we could move this in the state (e.g. in an instance attribute) - I preferred to be explicit here. Plus this is not possible in the case of `list_tool_deployments()` or `ToolDeployment.objects.filter()` which would have to be "special" and require the client code to pass the `id_token` anyway.

Potential future changes could involve grouping the functions related to a tool deployment into a `cluster.ToolDeployment` class:

```python
class ToolDeployment():
  @classmethod
  def get_deployments()

  def deploy()
  def get_status()
  def get_deployment()
  def restart()
  def delete()
```

(or something like that)

Arguably the "fake" model `ToolDeployment` is really responsible for interacting with the k8s cluster and could be removed/moved into `cluster` but I've avoided this for the time being.

### How to review

1. Run tests, they should pass
2. Run CP locally (both HTTP server and worker)
3. You should be able to see the list of tools, and perform operations on them

### Ticket

### DIFF
--- a/controlpanel/api/views/tools.py
+++ b/controlpanel/api/views/tools.py
@@ -36,4 +36,9 @@ class ToolDeploymentViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
         return Response({}, status=status.HTTP_201_CREATED)
 
     def get_queryset(self):
-        return ToolDeployment.objects.filter(user=self.request.user)
+        user = self.request.user
+        id_token = user.get_id_token()
+        return ToolDeployment.objects.filter(
+            user=user,
+            id_token=id_token,
+        )
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -496,7 +496,7 @@ def deploy_tool(tool, user, **kwargs):
         raise ToolDeploymentError(error)
 
 
-def list_tool_deployments(user, id_token=None, search_name=None, search_version=None):
+def list_tool_deployments(user, id_token, search_name=None, search_version=None):
     deployments = []
     k8s = KubernetesClient(id_token=id_token)
     results = k8s.AppsV1Api.list_namespaced_deployment(user.k8s_namespace)
@@ -511,12 +511,12 @@ def list_tool_deployments(user, id_token=None, search_name=None, search_version=
     return deployments
 
 
-def get_tool_deployment(tool_deployment, **kwargs):
+def get_tool_deployment(tool_deployment, id_token):
     deployments = list_tool_deployments(
         tool_deployment.user,
+        id_token,
         search_name=tool_deployment.tool.chart_name,
         # search_version=tool_deployment.tool.version,
-        **kwargs,
     )
 
     if not deployments:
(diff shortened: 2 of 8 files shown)


controlpanel/api/views/tools.py
```diff
@@ -36,4 +36,9 @@ def create(self, request):
         return Response({}, status=status.HTTP_201_CREATED)
 
     def get_queryset(self):
-        return ToolDeployment.objects.filter(user=self.request.user)
+        user = self.request.user
+        id_token = user.get_id_token()
+        return ToolDeployment.objects.filter(
+            user=user,
+            id_token=id_token,
+        )
```
andyhd: This is a bit weird, because it reads like you're filtering based on an `id_token` field. I suppose that shows the problem with mimicking a Django model.

xoen: Thanks for the review @andyhd :)

That may be a next step in the refactoring. I haven't looked at the consequences of removing the model altogether, but I imagine there is something more done in the views/serialisers to make it look like a real model so I'm not touching at the moment.

I could still group these functions in a `cluster.ToolDeployment` class and use it from the `models.ToolDeployment` class as a stepping stone to future changes but I'll probably leave as a separate PR anyway as this is enough for a change set IMHO.

Another thing I'm wrapping my head around is the helm install subprocess polling which is currently distributed between `cluster`module (`deploy_tool()`), `models.ToolDeployment` (`get_status()`/`_poll()`) and `consumers` module (`wait_for_deployment()`). We could probably move this "waiting/checking" logic into the class instance but I didn't think too much about it yet.

PR #762, opened 2019-10-08
Moved `cluster` functions into `ToolDeployment` class

xoen:
This logically groups together the functions that operate on a tool deployment.

(also fixed `HelmError` import in `cluster` module)

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -6,7 +6,7 @@ from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from github import Github, GithubException
 
 from controlpanel.api.aws import aws, iam_arn
-from controlpanel.api.helm import helm
+from controlpanel.api.helm import helm, HelmError
 from controlpanel.api.kubernetes import KubernetesClient
 from controlpanel.utils import github_repository_name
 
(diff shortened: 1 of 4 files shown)


xoen:
@andyhd I've tested this locally and it seems to work. I'm having another look in the GH diff view which is nicer but feel free to double check if I've missed something when lifting these functions into the `cluster.ToolDeployment` class ☺️

controlpanel/api/cluster.py
```diff
@@ -463,121 +463,139 @@ class ToolDeploymentError(Exception):
...
+                *set_values,
+            )
+
+        except HelmError as error:
+            raise ToolDeploymentError(error)
+
+    def uninstall(self, id_token):
+        deployment = self.get_deployment(id_token)
+        helm.delete(
+            deployment.metadata.name,
```
xoen: @andyhd do you remember why when we install the tool charts we use ` f"{self.chart_name}-{self.user.slug}"` as release name (now moved in `ToolDeployment.release_name` property) but when we uninstall we get the `Deployment` name from k8s?

Not a big deal, I've left as it is for now (as this is just a refactoring) but potentially we could remove the need to make a request to the k8s API and just run the `helm delete`.

xoen: (plus if we don't talk with k8s API we don't need the `id_token` parameter)

andyhd: I think I was planning to use Kubernetes generated release names instead of specifying them. But yes, reuse `release_name` here and avoid the API call. 👍

controlpanel/api/cluster.py
```diff
@@ -463,121 +463,139 @@ class ToolDeploymentError(Exception):
...
+            deployments.append(deployment)
+        return deployments
+
+    def get_deployment(self, id_token):
+        deployments = self.__class__.get_deployments(
+            self.user,
+            id_token,
+            search_name=self.chart_name,
+            # search_version=tool_deployment.tool.version,
+        )
```
andyhd: Having `get_deployment` as an instance method feels a bit weird. Maybe this method should be renamed something like `load(self, id_token)` and store the result in `self._deployment`?

controlpanel/api/cluster.py
xoen: That would be nice, would that break with some of the installed releases you reckon?

controlpanel/api/cluster.py
xoen: I'm not that keen on state for this as it mud the water, we wouldn't use this in other methods anyway and if we do it also means that some client code may rely on this method being called before something else (and this instance attribute being set) which may lead to bugs.

As this is a refactoring PR focused on moving these function in the class I'm not going to change anything else - including the use of `release_name` which *may* change the actual behaviour. We can change that independently.

controlpanel/api/cluster.py
xoen: To be fair, we currently don't expose the "uninstall" tool functionality, still left for next round of refactorings as don't like to change too much at the same time ☺️ 

(I'm testing these changes in `dev` again and it looks like I didn't break anything so release to `alpha` will probably follow soon)

PR #763, opened 2019-10-14
Explicitly catch and gracefully handle errors

andyhd:
A lot of exceptions were being left to bubble up to the top-level Django handler, which simply displays a 500 error page with no further information for the user.
This change captures and handles several of those exceptions in a more graceful way, by showing useful warning messages or form error messages and allowing the user to retry their action.
There are still more exceptions left to handle, but this PR is already becoming too big, so I will continue in several smaller PRs focusing on different areas of the Control Panel.
This PR also includes a refactor of the AWS client code, with a view to making it simpler and removing a redundant layer of abstraction.


controlpanel/api/models/app.py
```diff
@@ -51,16 +51,23 @@ def customers(self):
...
-                user_options={"connection": "email"},
-            )
+            try:
+                auth0.AuthorizationAPI().add_group_members(
+                    group_name=self.slug,
+                    emails=emails,
+                    user_options={"connection": "email"},
+                )
+            except auth0.Auth0Error as e:
+                raise AddCustomerError(e)
```
xoen: Nice ☺️

andyhd:
Splitting AWS refactor into separate PR

PR #764, opened 2019-10-15
Refactor AWS client code to use boto3 Resource API

andyhd:
* Refactor AWS client code to use boto3 Resource API
* Remove redundant `AWSClient` class
* Remove redundant `IAMRole` class (replaced by boto3 `Role` resource class)
* Refactor `S3AccessPolicy` class to operate on boto3 `RolePolicy` objects - also fixes bug which removes all resources from policy
* Refactor `AppS3Bucket`, `PolicyS3Bucket` and `UserS3Bucket` to loosen coupling with `S3AccessPolicy` and `ManagedS3AccessPolicy`
* Refactor `cluster` module to follow @xoen's `cluster.App` example
* Test with moto
* Remove the unused `write_to_cluster` feature flag


controlpanel/api/aws.py
```diff
@@ -24,248 +22,415 @@ def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
...
+        'Action': [
+            's3:GetObject',
+            's3:GetObjectAcl',
+            's3:GetObjectVersion',
+        ],
+        'Effect': 'Allow',
+    },
+}
+
+BASE_S3_ACCESS_STATEMENT['readwrite'] = {
```
xoen: I've only started to have a look (so it may be a stupid observation) but I found the fact that we're changing a "constant" a bit confusing.
Also the fact that the `BASE_S3_ACCESS_STATEMENT` already contains the `readonly` and `readwrite` statements could lead to bugs if we're not careful.

Although I get that without resources it wouldn't actually grant any permission.

But let me have a look at the rest to see how this is used.

andyhd: This dict is just used as a lookup for constant values - perhaps it would be better as `BASE_LIST_STATEMENT`, `BASE_READONLY_STATEMENT` and `BASE_READWRITE_STATEMENT`. Having them in a dict lets me look up the base statement for the SID programmatically later on.

controlpanel/api/aws.py
```diff
@@ -24,248 +22,415 @@ def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
...
-        block_public_acls=True,
-        ignore_public_acls=True,
-        block_public_policy=True,
-        restrict_public_buckets=True,
-    ):
-        self._do(
-            "s3",
-            "put_public_access_block",
+
+def create_bucket(bucket_name, is_data_warehouse=False):
```
xoen: I'm not suggesting moving more things around as this was probably already there...

...but it feels strangely hi-level to have this `is_data_warehouse` argument in this function.
If the `aws.create_bucket()` function responsibility is to abstract the creation of an S3 bucket maybe it should receive a tags dictionary (and convert to the list of `{"Key": "...", "Value": "..."}`) instead of whether the S3 bucket is a data warehouse.

It makes me also wonder if this could belong in a `cluster.S3Bucket` class wrapping the `S3Bucket` model.
I notice that we also have a `cluster.create_bucket()` which is just calling this 🤷‍♂ 

Is `aws.create_bucket()` used somewhere else?

Maybe *long-term* if we really want to, we could move most of these AWS functions in the relevant methods in the `cluster` module and maybe just leave the class dealing with the IAM policy here...but that's for another refactoring.

controlpanel/api/cluster.py
```diff
@@ -218,230 +110,68 @@ def url(self):
...
-            "StringEquals": {
-                f"{settings.OIDC_DOMAIN}/:sub": sub,
-            }
-        }
-        self.policy["Statement"].append(stmt)
-
-
-class S3AccessPolicy(dict):
-    POLICY_NAME = "s3-access"
+class Bucket:
```
xoen: => `class S3Bucket` for consistency with pretty much everything else ☺️

controlpanel/api/aws.py
xoen: I had a look and it seems like `aws.create_bucket()` is only used in `cluster.Bucket.create()` which you just added so maybe we can bite the bullet and just move this logic there, but I'd still leave for a future refactoring.

controlpanel/api/aws.py
```diff
@@ -24,248 +22,415 @@ def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
...
+
+class S3AccessPolicy:
+    """Provides a convenience wrapper around a RolePolicy object"""
+
+    def __init__(self, policy):
+        self.policy = policy
+        self._policy_document = None
+        self.statements = {}
+
+        self._load()
```
xoen: This is a bit confusing but I haven't followed the various iterations closely so maybe I'm missing something...

The `policy` argument is always passed and it's coming from the IAM role inline policy (usually `role.Policy('s3-access')` or something like that).

So what's with all these weird instance attributes that we keep around?

- We store the policy passed to the constructor in `self.policy`
- `self. _policy_document ` is `None` but it would eventually contain the same thing after calling the `policy_document` property right?
- and we also have the `policy_document` property which is memoizing the same thing in `self._policy_document`

The `self.policy` attribute is a `boto` thing and we also use it to update the policy. Maybe we can get rid of some stuff and have something like:

```python
class S3AccessPolicy:

    def __init__(self, policy):
        self.policy = policy
        self.policy_document = policy.policy_document # do we need to make a copy?
        self.statements = {}
```

Would you still need the `self._policy_document` instance attribute and the `policy_document` property in that case?

controlpanel/api/aws.py
```diff
@@ -24,248 +22,415 @@ def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
...
+            if sid not in self.statements:
+                stmt = dict(BASE_S3_ACCESS_STATEMENT[sid])
+                self.statements[sid] = stmt
+                self.policy_document['Statement'].append(stmt)
+            return self.statements[sid]
+
+    def add_resource(self, arn, sid):
+        statement = self.statement(sid)
+        if statement:
+            resources = statement.setdefault('Resource', [])
```
xoen: This is weird, so this is to tell the dictionary to return an empty list when the key is not present right?

Wouldn't be easier to just add the `"Resource": []` key to the statements in the "constants"?

controlpanel/api/aws.py
andyhd: This `aws` module, is code for the platform to interact with AWS, so it definitely makes sense to have platform domain objects (eg data warehouse) referenced here. If we only presented a low-level interface to AWS here, there would be no need for this module - calling code could use `boto3` directly.

I think you are right that `cluster.create_bucket` should be moved into `cluster.Bucket.create`

controlpanel/api/cluster.py
andyhd: The name of `S3Bucket` is a mistake that I am slowly correcting

andyhd: Just kidding, I'll change the name

xoen: 😱 You know you freaked out for a moment 😝

controlpanel/api/helm.py
```diff
@@ -18,13 +18,7 @@ class HelmError(APIException):
 
 class Helm(object):
 
-    def __init__(self):
-        self.enabled = settings.ENABLED['write_to_cluster']
-
     def _execute(self, *args, check=True, **kwargs):
-        if not self.enabled:
```
xoen: Less code makes me happy ☺️

controlpanel/api/models/apps3bucket.py
```diff
@@ -24,9 +22,19 @@ class Meta:
...
         return f'<AppS3Bucket: {self.app!r} {self.s3bucket!r} {self.access_level}>'
 
+    def grant_bucket_access(self):
+        cluster.App(self.app).grant_bucket_access(
+            self.s3bucket.arn,
+            self.access_level,
+        )
+
+    def revoke_bucket_access(self):
+        cluster.App(self.app).revoke_bucket_access(self.s3bucket.arn)
```
xoen: nice ☺️

xoen: oh you added these to a `cluster.App` class instead of having something like `cluster.AppS3Bucket`.
Yeah makes sense as it's the subject anyway 👍

controlpanel/api/models/apps3bucket.py
```diff
@@ -24,9 +22,19 @@ class Meta:
         unique_together = ('app', 's3bucket')
         ordering = ('id',)
 
-    def aws_name(self):
+    @property
+    def iam_role_name(self):
```
xoen: Good idea to keep this consistent across classes, as I was thinking in the `App` model we could potentially removed these from the model altogether, especially if they're just used in `cluster.*` classes. (but we can do later on if we want)

controlpanel/api/models/policys3bucket.py
```diff
@@ -21,5 +18,16 @@ class Meta:
         unique_together = ("policy", "s3bucket")
         ordering = ("id",)
 
+    @property
```
xoen: I'll probably go as far as renaming this to `arn` as we have other places and we call these properties `arn` so it may be good to have the same here for consistency.

controlpanel/api/models/policys3bucket.py
```diff
@@ -21,5 +18,16 @@ class Meta:
...
     def aws_name(self):
         return self.policy.arn
+
+    def grant_bucket_access(self):
+        cluster.Group(self.policy).grant_bucket_access(
```
xoen: What's a group? I must have missed something in the last few iterations.

xoen: > Represents a group of cluster users who have access to datasources

controlpanel/api/models/s3bucket.py
```diff
@@ -65,7 +65,7 @@ def __repr__(self):
 
     @property
     def arn(self):
-        return s3_arn(self.name)
+        return cluster.Bucket(self).arn
```
xoen: Nicer, better 👍

controlpanel/api/models/s3bucket.py
```diff
@@ -98,9 +98,7 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
         if is_create:
-            bucket = cluster.create_bucket(self.name, self.is_data_warehouse)
-            if bucket:
-                self.location_url = bucket['Location']
```
xoen: So we *definitely* don't use this `location_url` field, I seem to remember you mentioned checking the DB and it's always blank or something 🤷‍♂

controlpanel/api/models/policys3bucket.py
andyhd: @s-block created the `IAMManagedPolicy` model and the `PolicyS3Bucket` which is essentially a group of users having access to an S3Bucket. We decided not to call the class `Group`, because it might be confusing with the `django.contrib.auth.models.Group`, but in retrospect, that was probably a mistake, and `IAMManagedPolicy` is a terrible name.

andyhd: I'm planning to fix that, but I wanted to avoid making any DB migrations in this PR.

controlpanel/api/models/s3bucket.py
andyhd: It is always blank, because we never saved the model after we set the field (ever since jmoz wrote it originally). But it isn't used anywhere anyway.

andyhd: Yes, it's never saved to the database, ever since jmoz first wrote it. And it's never used anyway.

controlpanel/settings/common.py
```diff
@@ -9,8 +9,6 @@
 # -- Feature flags
 
 ENABLED = {
-    # Enable writes to Kubernetes cluster
-    "write_to_cluster": is_truthy(os.environ.get("ENABLE_WRITE_TO_CLUSTER", True)),
```
xoen: 🎊

requirements.txt
```diff
@@ -19,6 +19,7 @@ Jinja2>=2.10.1,<3.0
 kubernetes==10.0.1
 misaka==2.1.1
 model_mommy==1.6.0
+moto==1.3.13
```
xoen: Isn't this more of a `requirements.dev.txt` dependency? Does the real thing rely on this to build/deploy?

doc/feature-flags.md
```diff
@@ -7,4 +7,3 @@ Current flags are:
 
 | Name | Description | Default |
 | ---- | ----------- | ------- |
```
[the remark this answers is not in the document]
andyhd: I did think that - we should be using more feature flags in future, though, so it doesn't hurt to keep it

requirements.txt
andyhd: Tests are in the docker image, because we run tests with docker-compose. I would like to remove the tests from the image, though...

controlpanel/api/models/access_to_s3bucket.py
```diff
@@ -45,27 +47,35 @@ class Meta:
...
         super().delete(*args, **kwargs)
 
     @property
     def resources(self):
         resources = [self.s3bucket.arn_from_path(p) for p in self.paths]
         return resources or [self.s3bucket.arn]
+
+
+# MUST use signals because cascade deletes do not call delete()
+@receiver(models.signals.pre_delete)
```
xoen: This called every time *any* record is deleted right?

controlpanel/api/aws.py
andyhd: Changing to your suggestion - it's much better.

controlpanel/api/aws.py
andyhd: We don't have `cluster.create_bucket` - maybe that was a typo.

controlpanel/api/models/policys3bucket.py
xoen: (but probably we can beef up the class comment to make it more clear why it's done, etc...)

xoen: oh hang on, I got confused, I thought the DB model was called `Group` but it's actually called `IAMManagedPolicy` so maybe we can rename the `cluster` class or maybe we stick with `cluster.IAMManagedPolicy` for consistency for the time being 🤷‍♂

controlpanel/api/cluster.py
```diff
@@ -218,230 +110,68 @@ def url(self):
...
-        )
 
-        if document is None:
-            return cls(role_name=role_name)
-
-        policy = cls(role_name=role_name, **document)
-        return policy
+class Group:
+    """
+    Represents a group of cluster users who have access to datasources
```
xoen: As mentioned in [other comment](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/764#discussion_r336036520) we should probably elaborate on what it is and why we did it, e.g.:

> A Group uses a managed policy as a way to group IAM roles that have access to same resources.
>
> This is done to overcome IAM limitation which doesn't allow adding IAM roles to IAM groups: https://stackoverflow.com/a/48087433/455642

or something like that, maybe even mentioned that each Analytical Platform user has an IAM role which is assumed by their tools - but that's probably overkill and potentially "basic knowledge".

controlpanel/api/models/policys3bucket.py
andyhd: This property isn't actually used by anything any more, so I'm going to remove it.

requirements.txt
andyhd: We don't run the tests during the Docker image build. I would like to, but the tests require a database to run against. We could build the image with tests included, run the tests and then build and push another image from the same codebase? Not in scope of this PR, though.

controlpanel/api/models/user.py
```diff
@@ -44,11 +44,11 @@ def get_id_token(self):
 
     @property
     def iam_role_name(self):
-        return f"{settings.ENV}_user_{self.username.lower()}"
+        return cluster.User(self).iam_role_name
```
xoen: Long term (as in, not as part of this), and assuming `iam_role_name` is mainly used in the `cluster` module we may well get rid of these from the models.

(maybe these properties are used in some of the models now to grant/revoke access by passing the target ARN but we may well pass the whole bucket to `(User|App).(grant|revoke)_access()` if necessary)

tests/api/cluster/test_group.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_arn(settings, iam_managed_policy):
+    assert cluster.RoleGroup(iam_managed_policy).arn == (
+        f'arn:aws:iam::{settings.AWS_ACCOUNT_ID}:policy/{settings.ENV}/group/test'
+    )
+
+
+def test_create(aws, iam_managed_policy):
+    cluster.RoleGroup(iam_managed_policy).create()
+    aws.create_group_policy.assert_called_with(
```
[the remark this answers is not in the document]
andyhd: maybe `aws.create_rolegroup`

controlpanel/api/models/access_to_s3bucket.py
```diff
@@ -45,27 +47,35 @@ class Meta:
...
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        self.grant_bucket_access()
         return self
 
     def delete(self, *args, **kwargs):
-        with self.policy_class.load(self.aws_name()) as policy:
-            for resource in self.resources:
-                policy.revoke_access(resource)
-
```
xoen: As discussed offline earlier (and for the record) the bit on the right makes more sense to me, but at the same time it would be also useful to understand what the original logic was doing and if we're missing something important here.

controlpanel/api/cluster.py
```diff
@@ -218,230 +114,72 @@ def url(self):
...
-        policy = cls(policy_arn=policy_arn, **document)
-        return policy
+    def revoke_bucket_access(self, bucket_arn):
+        aws.revoke_group_bucket_access(self.arn, bucket_arn)
 
-    def save(self):
-        aws.create_policy_version(
-            self.policy_arn,
-            policy_document=self,
-        )
```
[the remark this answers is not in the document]
andyhd: This is replaced by [a call to `boto3.resource('iam').Policy.create_version()`](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/764/files/5c6956d6a684a6f2a545433c4e8f68f31ff4542d#diff-655f3b2e4206c14d9b293048088f97f1R315)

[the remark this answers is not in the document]
xoen: 👍

controlpanel/api/models/policys3bucket.py
```diff
@@ -21,5 +19,12 @@ class Meta:
         unique_together = ("policy", "s3bucket")
         ordering = ("id",)
 
-    def aws_name(self):
-        return self.policy.arn
+    def grant_bucket_access(self):
+        cluster.RoleGroup(self.policy).grant_bucket_access(
```
s-block: `cluster` isn't imported

controlpanel/api/aws.py
```diff
@@ -24,248 +23,418 @@ def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
...
-    def delete_policy(self, policy_arn):
-        self._detach_policy_from_roles(policy_arn)
-        self._delete_policy_versions(policy_arn)
-        self._do("iam", "delete_policy", PolicyArn=policy_arn)
+def delete_group(group_arn):
+    policy = boto3.resource('iam').Policy(group_arn)
+    try:
+        policy.load()
+    except policy.meta.client.exceptions.NoSuchEntityException:
+        log.warning(f"Skipping deletion of policy {group_arn}: Does not exist")
```
s-block: Is `log` imported/created?

controlpanel/api/cluster.py
```diff
@@ -218,230 +114,72 @@ def url(self):
...
+    def path(self):
+        return f'/{settings.ENV}/group/'
 
-    def save(self):
-        aws.put_role_policy(
-            self.role_name,
-            policy_name=self.POLICY_NAME,
-            policy_document=self,
+    def create(self):
+        aws.create_group_policy(
```
s-block: `create_group_policy` does not exist in `aws`.

andyhd: Yep, I think @xoen already flagged this

controlpanel/api/aws.py
andyhd: oops 👍

controlpanel/api/aws.py
andyhd: `setdefault` saves a line. Without it we have to check if the statement dict has a `Resource` key and if it is missing (possible, but unlikely), create it, append the arn, then update the statement dict with the newly-created list. Eg:
```python
if statement:
    resources = statement.get('Resource', [])
    if arn not in resources:
        resources.append(arn)
        statement['Resource'] = resources
```

andyhd: actually, i think i'll do it without `setdefault`

controlpanel/api/aws.py
andyhd: This is a bit ugly, I agree. The `role.Policy.policy_document` is a lazy-loaded property, so if we assign it to `self.policy_document` in the constructor, that will trigger the API call. That would mean we have to move the `_load()` method code into the constructor to catch a possible `NoSuchEntityException`, and we would have to duplicate in the `ManagedS3AccessPolicy` class.
I'll try again.

controlpanel/api/aws.py
```diff
@@ -24,248 +23,418 @@ def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
...
-        )
+    for role in policy.attached_roles.all():
+        policy.detach_role(RoleName=role.name)
 
-    def _delete_policy_versions(self, policy_arn):
-        versions = self.list_policy_versions(policy_arn).get("Versions", [])
-        for version in versions:
-            if not version["IsDefaultVersion"]:
-                self._delete_policy_version(policy_arn, version["VersionId"])
+    # for group in policy.attached_groups.all():
```
s-block: Does this commented code need to be there?

controlpanel/api/models/access_to_s3bucket.py
```diff
@@ -45,27 +47,35 @@ class Meta:
...
+        self.revoke_bucket_access()
         super().delete(*args, **kwargs)
 
     @property
     def resources(self):
         resources = [self.s3bucket.arn_from_path(p) for p in self.paths]
         return resources or [self.s3bucket.arn]
+
+
+# MUST use signals because cascade deletes do not call delete()
```
s-block: :+1:

controlpanel/api/aws.py
andyhd: Also, because the `role.Policy.policy_document` is the return value of an API call, it is a different object every time, so I store it in `self._policy_document`.

controlpanel/api/aws.py
andyhd: No, I'll remove it

controlpanel/api/aws.py
```diff
@@ -221,32 +224,31 @@ class S3AccessPolicy:
...
                 self.statements[sid] = stmt
 
-    @property
-    def policy_document(self):
-        if self._policy_document is None:
-            self._policy_document = self.policy.policy_document
-        return self._policy_document
+        self.policy_document = policy_document
+
+    def load_policy_document(self):
```
xoen: Do we need to add a method? Is it used anywhere? If not I reckon could be simpler to inline in the constructor (but you know the code better)

controlpanel/api/aws.py
```diff
@@ -221,32 +224,31 @@ class S3AccessPolicy:
...
-        self._policy_document = None
         self.statements = {}
 
-        self._load()
-
-    def _load(self):
         try:
-            # ensure version is set
-            self.policy_document['Version'] = '2012-10-17'
+            policy_document = self.load_policy_document()
```
xoen: any benefit of using another variable instead of using the instance attribute?

controlpanel/api/aws.py
```diff
@@ -221,32 +224,31 @@ class S3AccessPolicy:
...
-            self.put(policy_document=deepcopy(BASE_S3_ACCESS_POLICY))
+            policy_document = deepcopy(BASE_S3_ACCESS_POLICY)
+            self.put(policy_document=policy_document)
+
+        # ensure version is set
+        policy_document['Version'] = '2012-10-17'
 
         # ensure statements are correctly configured and build a lookup table
-        for stmt in self.policy_document.setdefault('Statement', []):
+        for stmt in policy_document.setdefault('Statement', []):
```
xoen: You still need the `setdefault()`?

controlpanel/api/aws.py
```diff
@@ -259,9 +261,9 @@ def statement(self, sid):
     def add_resource(self, arn, sid):
         statement = self.statement(sid)
         if statement:
-            resources = statement.setdefault('Resource', [])
-            if arn not in resources:
-                resources.append(arn)
+            statement['Resource'] = statement.get('Resource', [])
```
xoen: I kind of prefer this to the `setdefault` to be honest, the "get or use default" is a more common pattern IMHO and it's probably less surprising as you see what you get if `Resource` is not there at the point of the `get()` call.

But is this still needed if you add the `Resource` keys?

controlpanel/api/aws.py
```diff
@@ -293,6 +295,9 @@ def put(self, policy_document=None):
             if stmt.get('Resource')
         ]
 
+        return self.save_policy_document(policy_document)
+
+    def save_policy_document(self, policy_document):
```
xoen: Considering we're within the context of `S3AccessPolicy` class we may even call this just "save" 🤷‍♂

controlpanel/api/aws.py
```diff
@@ -221,32 +224,31 @@ class S3AccessPolicy:
...
+            policy_document = self.load_policy_document()
 
         except self.policy.meta.client.exceptions.NoSuchEntityException:
             # create an empty s3 access policy
-            self.put(policy_document=deepcopy(BASE_S3_ACCESS_POLICY))
+            policy_document = deepcopy(BASE_S3_ACCESS_POLICY)
+            self.put(policy_document=policy_document)
+
+        # ensure version is set
+        policy_document['Version'] = '2012-10-17'
```
xoen: If we find that the logic to initialise the `self.policy_document` attribute is getting a bit hairy/low-level we may consider moving this logic into a "private" helper method. Maybe even in  `load_policy_document()`.

If you feel like assigning `self.policy_document` from within `load_policy_document` is a bit sneaky you may have something like:

```python
self.policy_document = self.get_policy_document()
```

tests/api/cluster/test_access_to_s3buckets.py
```diff
@@ -0,0 +1,78 @@
...
+)
+
+@pytest.fixture(autouse=True)
+def enable_db_for_all_tests(db):
+    pass
+
+
+@pytest.fixture
+def bucket():
+    return mommy.make('api.S3Bucket')
```
xoen: Do you need to use `model_mommy`? and also, do we need to persist the `S3Bucket` in the DB for this test? [`mommy.make()` will persist to DB](https://model-mommy.readthedocs.io/en/latest/basic_usage.html#non-persistent-objects) which is slow.

One test persisting wouldn't make much difference but in general we should avoid persisting to DB if not necessary as many of these tests cause the whole testsuite to be slow and increate the test feedback cycle.

xoen: AFAICT these tests "just" check that the correct AWS calls are made, so probably you don't need to persist the various records to the DB - but I didn't try to tweak/run the tests.

tests/api/cluster/test_app.py
```diff
@@ -2,35 +2,21 @@
...
-from controlpanel.api.cluster import BASE_ROLE_POLICY, App
-from controlpanel.api import models
-
-from django.conf import settings
+from controlpanel.api import cluster, models
 
 
 @pytest.fixture
-def app_model():
+def app():
```
xoen: reason why I called this `app_model` was as we're in the context of the `cluster` module which also contains an `App` class and I thought it may lead to confusion. But no need to change, let's see how it goes 👍

tests/api/cluster/test_bucket.py
```diff
@@ -0,0 +1,23 @@
+from model_mommy import mommy
+import pytest
+
+from controlpanel.api import cluster
+
+
+@pytest.fixture(autouse=True)
+def enable_db_for_all_tests(db):
+    pass
```
xoen: uh?

tests/api/cluster/test_bucket.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.fixture
+def bucket():
+    return mommy.make('api.S3Bucket', name='test-bucket')
```
xoen: as mentioned in other comment, probably persist the `S3Bucket` to the DB just to test the `arn` property and check the AWS calls are called is overkill and unnecessarily slow down the tests.

tests/api/cluster/test_bucket.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def test_arn(bucket):
+    assert cluster.S3Bucket(bucket).arn == 'arn:aws:s3:::test-bucket'
+
+
+def test_create(aws, bucket):
+    cluster.S3Bucket(bucket).create()
+    aws.create_bucket.assert_called_with(bucket.name, False)
```
xoen: Potentially I'd use the named argument as it helps with booleans and avoid confusion in case we add another argument. (no need to change unless you feel like it, just an observation as tests are also documentation and I had to look at the method definition to remember what was the second argument)

Also, potentially the `aws.create_bucket()` shouldn't care about high-level thing like whether a bucket is "data warehouse" or not, but happy to keep it as it is for the time being.

In my view, as mentioned in another comment, we could change the `aws.create_bucket()` API to receive tags so to be agnostic of how we use S3 buckets, e.g.

```python
aws.create_bucket(bucket_name, {"buckettype": "datawarehouse"})
```

but I do appreciate is a bit long-winded and it would need to transform the Python dictionary into the weird array of dictionaries with `Key`/`Value` internally.

But this is currently used only in `cluster.S3Bucket.create()` which could (should IMHO) pass the `{"buckettype": "datawarehouse"}` tags dictionary (as in my mind the `aws` module is more low-level).

But as I said, no need to change for now.

tests/api/cluster/test_group.py
```diff
@@ -0,0 +1,29 @@
+from model_mommy import mommy
+import pytest
+
+from controlpanel.api import cluster
+
+
+@pytest.fixture(autouse=True)
+def enable_db_for_all_tests(db):
+    pass
```
xoen: I've seen this in several tests files, is this a special method that once defined tells pytest to not write to the DB or something?

tests/api/cluster/test_user.py
```diff
@@ -0,0 +1,48 @@
...
+
+    aws.create_user_role.assert_called_with(user)
+    expected_calls = [
+        call(
+            f'init-user-{user.slug}',
+            'mojanalytics/init-user',
+            f"--set=NFSHostname={settings.NFS_HOSTNAME},"
+            f"Username={user.slug},"
+            f"Email={user.email},"
+            f"Fullname={user.get_full_name()},"
```
xoen: I see that `models.User.get_full_name()` returns `self.name`, this is few months old, I wonder what was the rationale for this method and if needed/nicer to have if it could be a property (as `joe.full_name` reads better than `joe.get_full_name()` IMHO).

But as this is not new, no need to change, it's just an observation.

tests/api/models/test_apps3bucket.py
```diff
@@ -56,7 +42,12 @@ def test_update_aws_permissions(app, bucket, aws):
 
     apps3bucket.save()
 
-    aws.put_role_policy.assert_called()
+    aws.grant_bucket_access.assert_called_with(
+        app.iam_role_name,
+        bucket.arn,
+        AppS3Bucket.READONLY,
+        apps3bucket.resources,
```
xoen: I wonder if at some point we should rename this to something less ambiguous/confusing than `resources`.
It took us a bit to understand that this was referring to S3 bucket paths.

I like much more the `path_arns` name used in the `aws.grant_bucket_access()` method definition.

tests/api/test_aws.py
```diff
@@ -1,144 +1,471 @@
...
 
-    def test_put_bucket_logging(self):
-        aws.put_bucket_logging('bucketname', 'target_bucket', 'target_prefix')
-        aws.client.return_value.put_bucket_logging.assert_called()
+@pytest.fixture(autouse=True)
+def aws_creds():
+    os.environ['AWS_ACCESS_KEY_ID'] = 'testing'
+    os.environ['AWS_SECRET_ACCESS_KEY'] = 'testing'
+    os.environ['AWS_SECURITY_TOKEN'] = 'testing'
+    os.environ['AWS_SESSION_TOKEN'] = 'testing'
```
xoen: not a biggie but probably better to use different test strings as it helps when tests fail (e.g. `test-key-id`, `test-secret-access-key`, etc...)

tests/api/test_aws.py
```diff
@@ -1,144 +1,471 @@
...
-        policy_name = 's3-access'
-        policy_document = {'test': 'policy document'}
 
-        aws.client.return_value.get_role_policy.return_value = {
-            'PolicyDocument': policy_document,
-        }
+@pytest.yield_fixture(autouse=True)
+def iam(aws_creds):
+    with moto.mock_iam():
+        yield boto3.resource('iam')
```
xoen: how does `autouse=True` works in this case? Probably a stupid question, would this mock IAM within this file right? not across all the tests right?

xoen:
had a look at the tests, I can't remember if we said we'll have the mocking of all boto calls when running tests, maybe it's left for another time, I'd be interested in see how it's done ☺️ 

You got Josh's bless. I left some comments but there is *a lot* touched by this PR so hard to tell if my review was good enough.

controlpanel/api/aws.py
andyhd: This is to allow the subclass to override it without having to copy out the rest of the constructor code.

controlpanel/api/aws.py
andyhd: It's better than `self.policy_document['Statement'] = self.policy_document.get('Statement', [])` - but it should be on a separate line.

controlpanel/api/aws.py
andyhd: Yes. This is defensive coding in case the statement with the specified Sid exists, but has no Resource key. Unlikely but possible.

andyhd: `setdefault` is part of the core language and shouldn't be surprising, but I agree that this is probably clearer.

controlpanel/api/aws.py
andyhd: I don't want to have both `save` and `put`. This method is for the subclass to override.

controlpanel/api/aws.py
andyhd: No, the `load_policy_document` method is deliberately as concise as possible to cover only the code that needs to be overridden in the subclass.

tests/api/cluster/test_access_to_s3buckets.py
andyhd: I benchmarked this and it makes no time difference to use `mommy.make` vs `mommy.prepare` vs just instantiating the model class and not persisting it. But I've switched to `mommy.prepare` to avoid persisting to the database anyway.

tests/api/cluster/test_bucket.py
andyhd: This fixture is run for every test in this file (`autouse=True`). It requires the `pytest-django` fixture `db`, which enables database access in tests - which is disabled by default. The fixture function itself does not have to do anything, as the `db` fixture is evaluated on calling this test function.

tests/api/cluster/test_bucket.py
andyhd: The `create_bucket` function is not called with `is_data_warehouse=False`, and this test fails if we check for that.

As I explained in a previous comment, the `controlpanel.api.aws` module is explicitly wrapping a higher-level interface around the boto3 API, and referring to AP domain objects such as "data warehouse" is deliberate. Passing the tagset to this wrapper layer breaks that level of abstraction. The point of this module is to abstract the use of lower-level AWS API calls to achieve problem domain goals, eg "Create a data warehouse bucket", and not "Create a bucket in S3 with these specific tags".

tests/api/cluster/test_group.py
andyhd: It tells `pytest-django` that the test method is allowed to access the database, which is disabled by default.

tests/api/iam_test_cases.py
```diff
@@ -1,38 +0,0 @@
-from unittest.case import TestCase
-
-
-class IAMPolicyTestCase(TestCase):
```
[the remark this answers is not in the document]
andyhd: These tests are made obsolete by the use of moto and the updated aws module and model test functions

tests/api/test_aws.py
andyhd: these are not really meant to be used in tests, but to prevent test code from actually connecting to AWS

tests/api/test_aws.py
andyhd: `autouse=True` means the fixture is evaluated for every test function in this file. If this file was `conftest.py`, then the fixture would be evaluated for every test function in this directory and subdirectories.

PR #765, opened 2019-10-15
Explicitly catch and gracefully handle errors

andyhd:
*Depends on #764*

* Add logging and error handling for Github repo listing
* Prevent creating App with already registered Github repo
* Improve create app form validation and errors
* Validate app customer email addresses
* Handle Auth0 errors adding and removing customers
* Set default choice on create app form

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -82,10 +82,10 @@ class App:
     def iam_role_name(self):
         return f"{settings.ENV}_app_{self.app.slug}"
 
-    def create_iam_role(self):
+    def create(self):
         aws.create_app_role(self.iam_role_name)
 
-    def delete_iam_role(self):
+    def delete(self):
         aws.delete_role(self.iam_role_name)
 
     def grant_bucket_access(self, bucket_arn, access_level):
@@ -176,8 +176,12 @@ def get_repositories(user):
     repos = []
     github = Github(user.github_api_token)
     for name in settings.GITHUB_ORGS:
-        org = github.get_organization(name)
-        repos.extend(org.get_repos())
+        try:
+            org = github.get_organization(name)
+            repos.extend(org.get_repos())
+        except GithubException as err:
+            log.warning(f'Failed getting {name} Github org repos for {user}: {err}')
+            raise err
     return repos
 
 
@@ -186,6 +190,7 @@ def get_repository(user, repo_name):
     try:
         return github.get_repo(repo_name)
     except GithubException.UnknownObjectException:
+        log.warning(f'Failed getting {repo_name} Github repo for {user}: {err}')
         return None
 
 
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -82,10 +82,10 @@ class App:
     def iam_role_name(self):
         return f"{settings.ENV}_app_{self.app.slug}"
 
-    def create(self):
+    def create_iam_role(self):
         aws.create_app_role(self.iam_role_name)
 
-    def delete(self):
+    def delete_iam_role(self):
         aws.delete_role(self.iam_role_name)
 
     def grant_bucket_access(self, bucket_arn, access_level):
--- a/controlpanel/api/models/app.py
+++ b/controlpanel/api/models/app.py
@@ -51,16 +51,23 @@ class App(TimeStampedModel):
     def add_customers(self, emails):
         emails = list(filter(None, emails))
         if emails:
-            auth0.AuthorizationAPI().add_group_members(
-                group_name=self.slug,
-                emails=emails,
-                user_options={"connection": "email"},
-            )
+            try:
+                auth0.AuthorizationAPI().add_group_members(
+                    group_name=self.slug,
+                    emails=emails,
+                    user_options={"connection": "email"},
+                )
+            except auth0.Auth0Error as e:
+                raise AddCustomerError(e)
 
     def delete_customers(self, user_ids):
-        auth0.AuthorizationAPI().delete_group_members(
-            group_name=self.slug, user_ids=user_ids
-        )
+        try:
+            auth0.AuthorizationAPI().delete_group_members(
+                group_name=self.slug,
+                user_ids=user_ids,
+            )
+        except auth0.Auth0Error as e:
+            raise DeleteCustomerError(e)
 
     @property
     def status(self):
@@ -72,10 +79,22 @@ class App(TimeStampedModel):
         super().save(*args, **kwargs)
 
         if is_create:
-            cluster.App(self).create_iam_role()
+            cluster.App(self).create()
 
         return self
 
     def delete(self, *args, **kwargs):
-        cluster.App(self).delete_iam_role()
+        cluster.App(self).delete()
         super().delete(*args, **kwargs)
+
+
+class AddCustomerError(Exception):
+    pass
+
+
+class DeleteCustomerError(Exception):
+    pass
+
+
+App.AddCustomerError = AddCustomerError
+App.DeleteCustomerError = DeleteCustomerError
--- a/controlpanel/api/models/app.py
+++ b/controlpanel/api/models/app.py
@@ -58,7 +58,7 @@ class App(TimeStampedModel):
                     user_options={"connection": "email"},
                 )
             except auth0.Auth0Error as e:
-                raise AddCustomerError(e)
+                raise AddCustomerError from e
 
     def delete_customers(self, user_ids):
         try:
@@ -67,7 +67,7 @@ class App(TimeStampedModel):
                 user_ids=user_ids,
             )
         except auth0.Auth0Error as e:
-            raise DeleteCustomerError(e)
+            raise DeleteCustomerError from e
 
     @property
     def status(self):
@@ -79,12 +79,12 @@ class App(TimeStampedModel):
         super().save(*args, **kwargs)
 
         if is_create:
-            cluster.App(self).create()
+            cluster.App(self).create_iam_role()
 
         return self
 
     def delete(self, *args, **kwargs):
-        cluster.App(self).delete()
+        cluster.App(self).delete_iam_role()
         super().delete(*args, **kwargs)
 
 
--- a/controlpanel/api/validators.py
+++ b/controlpanel/api/validators.py
@@ -5,12 +5,15 @@ from django.core.validators import RegexValidator
 
 def validate_env_prefix(value):
     """Validate the name has env-prefix, check current set ENV value"""
-    if not value.startswith("{}-".format(settings.ENV)):
+    if not value.startswith(f"{settings.ENV}-"):
         raise ValidationError("Name must have correct env prefix e.g. {}-bucketname".format(settings.ENV))
 
 
 # An S3 bucket name needs to be min 3 chars, max 63 chars long.
-validate_s3_bucket_length = RegexValidator(regex='^.{3,63}$', message="must be between 3 and 63 characters")
+validate_s3_bucket_length = RegexValidator(
+    regex='^.{3,63}$',
+    message="must be between 3 and 63 characters",
+)
 
 # See: AWS' Bucket Restrictions and Limitations
 # http://docs.aws.amazon.com/en_gb/AmazonS3/latest/dev/BucketRestrictions.html
@@ -21,6 +24,23 @@ validate_s3_bucket_length = RegexValidator(regex='^.{3,63}$', message="must be b
 #
 # An S3 bucket name starts with a label, it can have more than one label
 # separated by a dot.
-validate_s3_bucket_labels = RegexValidator(regex='^([a-z][a-z0-9-]*[a-z0-9])(.[a-z][a-z0-9-]*[a-z0-9])*$',
-                                           message="is invalid, check AWS S3 bucket names restrictions (for example, "
-                                            "can only contains letters, digits, dots and hyphens)")
+validate_s3_bucket_labels = RegexValidator(
+    regex='^([a-z][a-z0-9-]*[a-z0-9])(.[a-z][a-z0-9-]*[a-z0-9])*$',
+    message="is invalid, check AWS S3 bucket names restrictions (for example, ""can only contains letters, digits, dots and hyphens)",
+)
+
+
+def validate_github_repository_url(value):
+    github_base_url = "https://github.com/"
+
+    if not value.startswith(github_base_url):
+        raise ValidationError("Must be a Github hosted repository")
+
+    repo_name = value[len(github_base_url):]
+    org, _ = repo_name.split("/", 1)
+
+    if org not in settings.GITHUB_ORGS:
+        orgs = ", ".join(settings.GITHUB_ORGS)
+        raise ValidationError(
+            f"Unknown Github organization, must be one of {orgs}",
+        )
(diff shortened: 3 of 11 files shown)


controlpanel/frontend/forms.py
```diff
@@ -176,20 +193,64 @@ def __init__(self, *args, **kwargs):
...
+        if self.strip:
+            emails = [email.strip() for email in emails]
+        return emails
+
+    def clean(self, value):
+        value = self.to_python(value)
+        for email in value:
+            try:
+                validate_email(email)
+            except ValidationError:
```
r4vi: what's wrong with the message `validate_email` gives? I don't think you need to change it

controlpanel/api/models/app.py
```diff
@@ -72,10 +79,22 @@ def save(self, *args, **kwargs):
...
+
+class AddCustomerError(Exception):
+    pass
+
+
+class DeleteCustomerError(Exception):
+    pass
+
+
+App.AddCustomerError = AddCustomerError
```
r4vi: why? just to make it more djangothonic?

controlpanel/frontend/views/app.py
```diff
@@ -232,16 +245,28 @@ def post(self, request, *args, **kwargs):
...
+        return HttpResponseRedirect(
+            reverse_lazy("manage-app", kwargs={"pk": self.kwargs['pk']}),
+        )
+
+    def form_valid(self, form):
+        self.get_object().add_customers(form.cleaned_data['customer_email'])
+        return HttpResponseRedirect(self.get_success_url())
+
+    def get_form_kwargs(self):
+        kwargs = FormMixin.get_form_kwargs(self)
```
r4vi: are you doing FormMixin.get_form_kwargs because `self.get_form_kwargs` comes from ModelFormMixin?

controlpanel/frontend/forms.py
andyhd: `validate_email` just says `Enter a valid email address.` - but the user may have entered a long list of email addresses, and it's helpful to know (the first one) which is invalid.

controlpanel/api/models/app.py
andyhd: saves importing a list of names 🤷‍♂ i'm not wed to it, if you think it should go

r4vi: it's fine, no need to change

controlpanel/api/cluster.py
```diff
@@ -82,10 +82,10 @@ def __init__(self, app):
     def iam_role_name(self):
         return f"{settings.ENV}_app_{self.app.slug}"
 
-    def create_iam_role(self):
+    def create(self):
```
xoen: ❌ Could we please keep the old names for these two methods? `create()` is very vague and without reading the method body I can't tell what's supposed to happen. (is `App.create()` creating an app as in deploying an app helm chart or is it creating an app pipeline or is it creating an app IAM role or an app or an app group in Auth0 or something else? `create()` is too ambiguous, require thinking and ultimately will lead to confusion and bugs IMHO)

I'd keep `create_iam_role()`/`delete_iam_role()` as it's 100% explicit about the method responsibility and what's going to happen.

controlpanel/api/cluster.py
```diff
@@ -176,8 +176,12 @@ def get_repositories(user):
...
     github = Github(user.github_api_token)
     for name in settings.GITHUB_ORGS:
-        org = github.get_organization(name)
-        repos.extend(org.get_repos())
+        try:
+            org = github.get_organization(name)
+            repos.extend(org.get_repos())
+        except GithubException as err:
+            log.warning(f'Failed getting {name} Github org repos for {user}: {err}')
+            raise err
```
xoen: So we're happy with not wrapping this GH exception into another exception in this case 👍

controlpanel/api/models/app.py
```diff
@@ -51,16 +51,23 @@ def customers(self):
...
-                user_options={"connection": "email"},
-            )
+            try:
+                auth0.AuthorizationAPI().add_group_members(
+                    group_name=self.slug,
+                    emails=emails,
+                    user_options={"connection": "email"},
+                )
+            except auth0.Auth0Error as e:
+                raise AddCustomerError(e)
```
xoen: we could use `raise AddCustomerError  from e` which enrich the exception with the cause of the exception (as discussed yesterday)

controlpanel/api/models/app.py
```diff
@@ -51,16 +51,23 @@ def customers(self):
...
-        auth0.AuthorizationAPI().delete_group_members(
-            group_name=self.slug, user_ids=user_ids
-        )
+        try:
+            auth0.AuthorizationAPI().delete_group_members(
+                group_name=self.slug,
+                user_ids=user_ids,
+            )
+        except auth0.Auth0Error as e:
+            raise DeleteCustomerError(e)
```
xoen: same here, we could use `raise DeleteCustomerError  from e` so that the exception keeps track of the cause of the error.

controlpanel/api/models/app.py
```diff
@@ -72,10 +79,22 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
         if is_create:
-            cluster.App(self).create_iam_role()
+            cluster.App(self).create()
```
xoen: See comment above about keeping old method names.

controlpanel/api/models/app.py
```diff
@@ -72,10 +79,22 @@ def save(self, *args, **kwargs):
...
 
         return self
 
     def delete(self, *args, **kwargs):
-        cluster.App(self).delete_iam_role()
+        cluster.App(self).delete()
```
xoen: See comment above about keeping old method names.

controlpanel/api/models/app.py
```diff
@@ -51,16 +51,23 @@ def customers(self):
     def add_customers(self, emails):
```
xoen: Wouldn't you say that these belongs in `cluster.App` rather than in the model as they change state outside the CP itself.

controlpanel/api/models/app.py
```diff
@@ -51,16 +51,23 @@ def customers(self):
...
+            try:
+                auth0.AuthorizationAPI().add_group_members(
+                    group_name=self.slug,
+                    emails=emails,
+                    user_options={"connection": "email"},
+                )
+            except auth0.Auth0Error as e:
+                raise AddCustomerError(e)
 
     def delete_customers(self, user_ids):
```
xoen: Again, it feels this `delete_customers()` method belongs in `cluster.App` rather than in the model.

xoen: When I say "feels" I mean it seems logical to as it interacts with the cluster rather than with the DB.

controlpanel/api/validators.py
```diff
@@ -5,12 +5,15 @@
 
 def validate_env_prefix(value):
     """Validate the name has env-prefix, check current set ENV value"""
-    if not value.startswith("{}-".format(settings.ENV)):
+    if not value.startswith(f"{settings.ENV}-"):
```
xoen: better 👍

controlpanel/frontend/forms.py
```diff
@@ -1,17 +1,22 @@
...
 from controlpanel.api import validators
 from controlpanel.api.cluster import get_repository
-from controlpanel.api.models import S3Bucket
+from controlpanel.api.models import App, S3Bucket
 from controlpanel.api.models.access_to_s3bucket import S3BUCKET_PATH_REGEX
 from controlpanel.api.models.iam_managed_policy import POLICY_NAME_REGEX
 from controlpanel.api.models.parameter import APP_TYPE_CHOICES
 
 
+APP_CUSTOMERS_DELIMITERS = re.compile(r'[,; ]+')
```
xoen: So this is to strip `[`, `,`, `;`, ` `, and `]` chars right? Have you seen anyone pasting angle brackets (`<`,`>`)? Maybe Auth0 is angle-brackets-friendly 🤷‍♂ ?

controlpanel/frontend/forms.py
```diff
@@ -49,53 +64,53 @@ def __init__(self, *args, **kwargs):
 
     def clean(self):
         cleaned_data = super().clean()
-
-        if cleaned_data['connect_bucket'] == "new" and not cleaned_data.get('new_datasource_name'):
-            self.add_error('new_datasource_name', "This field is required.")
-
-        if cleaned_data['connect_bucket'] == "existing" and not cleaned_data.get('existing_datasource_id'):
+        connect = cleaned_data['connect_bucket']
```
xoen: Makes sense.

controlpanel/frontend/forms.py
```diff
@@ -49,53 +64,53 @@ def __init__(self, *args, **kwargs):
...
+        repo_name = value.replace("https://github.com/", "", 1)
         repo = get_repository(self.request.user, repo_name)
         if repo is None:
             raise ValidationError(
                 f"Github repository not found - it may be private",
             )
 
-        return value
+        try:
+            app = App.objects.get(repo_url=value)
```
xoen: You could use something like `if App.objects.filter(repo_url=value).exists()` instead of `get()` and remove the "record doesn't exist" exception handling.

```python
if App.objects.filter(repo_url=value).exists():
    raise ValidationError(f"App already exists for this repository URL")
```

See: [Django's `exists()`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#exists)

controlpanel/frontend/forms.py
```diff
@@ -176,20 +193,64 @@ def __init__(self, *args, **kwargs):
 
 class CreateParameterForm(forms.Form):
     key = forms.CharField(
-        validators=[RegexValidator(r'[a-zA-Z0-9_]{1,50}')]
+        max_length=50,
+        validators=[
+            RegexValidator(
+                r'[a-zA-Z0-9_]{1,50}',
+                message="Must contain only alphanumeric characters and underscores",
```
xoen: Nice ☺️ - I'd also mention the character limit to avoid any confusion (as in someone entering a 51 chars parameter which looks valid but it's not because of its length)

controlpanel/frontend/forms.py
```diff
@@ -176,20 +193,64 @@ def __init__(self, *args, **kwargs):
...
+            ),
+        ],
     )
     role_name = forms.CharField(
-        validators=[RegexValidator(r'[a-zA-Z0-9_-]{1,60}')]
+        max_length=60,
+        validators=[
+            RegexValidator(
+                r'[a-zA-Z0-9_-]{1,60}',
+                message="Must contain only alphanumeric characters, underscores and hyphens",
```
xoen: As said for the other validator, this is nicer. I'd mentioned the character limit as well here.

controlpanel/frontend/views/app.py
andyhd: Yes, because AddAppCustomersForm is not a subclass of ModelForm and making it one would be more complex.

controlpanel/api/cluster.py
andyhd: This method is going to have more things in it than just creating the role - but ok, I'll change the name when I add that extra functionality.

controlpanel/api/cluster.py
andyhd: There's nothing we can do to handle this exception, so wrapping it is pointless.

controlpanel/api/models/app.py
andyhd: Sure, but that wasn't in the scope of this PR

andyhd: I realize I often do things that are outside the scope of a single PR 😅

controlpanel/api/models/app.py
andyhd: I can't refactor everything at once! (or so I'm told)

controlpanel/frontend/forms.py
andyhd: I haven't seen it, but I agree that is quite a likely mistake to make. Those angle backets would not be delimiters, though, so a different regex would be needed to sanitize the inputs. I'm not sure how far we should go to be lenient in what we accept from user input. I think in this case, you are right and we can sanitize emails in the form `Firstname Lastname <username@domain>`.

controlpanel/frontend/forms.py
andyhd: True. Although it is supposedly more Pythonic to catch exceptions. In this case, I think you are correct - I'm not using the value of `app` after the exception is caught or not, so `.exists()` expresses intent more precisely.

controlpanel/frontend/views/app.py
```diff
@@ -250,8 +275,12 @@ class RemoveCustomer(UpdateApp):
     def perform_update(self, **kwargs):
         app = self.get_object()
         user_ids = self.request.POST.getlist('customer')
-        app.delete_customers(user_ids)
-        messages.success(self.request, f"Successfully removed customer{pluralize(user_ids)}")
+        try:
+            app.delete_customers(user_ids)
+        except App.DeleteCustomerError:
+            messages.error(self.request, f"Failed removing customer{pluralize(user_ids)}")
```
xoen: we probably want the nerdy details in Sentry or at least in the logs ☺️ (as long as we're confident the exception doesn't contain sensitive information)

controlpanel/frontend/views/app.py
```diff
@@ -250,8 +275,12 @@ class RemoveCustomer(UpdateApp):
     def perform_update(self, **kwargs):
         app = self.get_object()
         user_ids = self.request.POST.getlist('customer')
-        app.delete_customers(user_ids)
-        messages.success(self.request, f"Successfully removed customer{pluralize(user_ids)}")
+        try:
+            app.delete_customers(user_ids)
+        except App.DeleteCustomerError:
```
xoen: I like those ☺️

controlpanel/api/models/app.py
xoen: ahaha good point, it was one of my usual "thinking aloud" comments as it was on code you didn't touch 😆

controlpanel/api/models/app.py
xoen: yeah sorry, it was just an observation more than an incitation to move even more things around, my bad 😁

controlpanel/frontend/forms.py
xoen: Agreed, probably out of scope and if it's a problem we could look at that separately (or maybe even add something in the client code to explain the input is invalid/prevent the submit ... in the future)

tests/frontend/views/test_app.py
```diff
@@ -222,3 +225,64 @@ def test_list(client, app, users, view, user, expected_count):
...
+        'invalid-email',
+        'mixed-valid-invalid-emails',
+        'no-emails',
+    ],
+)
+def test_add_customers(client, app, users, emails, expected_response):
+    client.force_login(users['superuser'])
+    data = {'customer_email': emails}
+    response = client.post(reverse('add-app-customers', kwargs={'pk': app.id}), data)
+    assert expected_response(client, response)
```
xoen: clever 👍

controlpanel/frontend/views/app.py
andyhd: sentry strips sensitive information

xoen:
@andyhd couldn't this be merged/released? Does it really depends on #764?

andyhd:
You're right, it doesn't. 👍

andyhd:
No, it does depend on it

PR #769, opened 2019-10-23
Security fixes for OWASP

andyhd:
* Ensure CSRF and session cookies are HTTP only and sent over HTTPS
* Disable client-side caching (static files are served by Nginx with caching allowed)
* Enable XSS protection
* Disable content-type sniffing


obutterbach:
Good job, thank you 👍

PR #770, opened 2019-11-04
Notify team slack on creation of superuser

andyhd:
Notify the team via Slack when a user is granted superuser status, or is created as a superuser.

There may be a double notification when creating a superuser via the REST API, but this should be a very rare occurrence.

This will need an update to the `cpanel` helm chart to provide a Slack API token.


controlpanel/settings/common.py
```diff
@@ -555,3 +555,9 @@
 AIRFLOW_SECRET_KEY = os.environ.get('AIRFLOW_SECRET_KEY')
 AIRFLOW_FERNET_KEY = os.environ.get('AIRFLOW_FERNET_KEY')
 
+
+# -- Slack
+SLACK = {
+    "api_token": os.environ.get('SLACK_API_TOKEN'),
```
xoen: If this is required (it is) I'd use `[]` instead of `get()` so that we blow up early if misconfigured instead of having a broken CP running and raising exceptions.

andyhd:
The only way to avoid multiple calling places is to pass the request user to the model save method, but I can't think of a way to do it that is less cluttered. We can notify that the superuser was created or updated in the model if we don't require that the message include the username of the creator/grantor.

controlpanel/api/models/user.py
```diff
@@ -84,9 +84,14 @@ def reset_mfa(self):
...
     def save(self, *args, **kwargs):
         try:
-            User.objects.get(pk=self.pk)
+            existing = User.objects.get(pk=self.pk)
+
         except User.DoesNotExist:
             cluster.User(self).create()
+            if self.is_superuser:
+                slack.notify_team(
+                    f"`{self.username}` was created as a superuser"
```
xoen: Do we get a double notification when a user is created as superuser? (is it even possible to do so? I thought users created themselves when they logged in and then someone else could make them superuser if required)

xoen:
Oh that's annoying I see. Still not sure why this needs to happen in 3 places. Isn't adding this in one place only in the view enough?

You could still potentially get the user from the request as we did in other places as it feels better to do that at model level rather than at views level (e.g. if for whatever reason we add other ways of making someone a superuser we'll have to make sure the notification code is in the new views as well for example)

andyhd:
The reasons notifications are sent in 3 places are because:
1. The user can be created as a superuser or granted superuser access from the Python shell, or programmatically (for example by migrations) - in this case, there is no request and so no way to tell who performed the action. Arguably, this notification is unnecessary, because if someone has access to the shell, that they have granted superuser access to a user is a lesser concern.
2. The user can be created as a superuser or granted superuser access via the REST API. In reality, this will likely never happen. The REST API has very few clients and they do not create or modify users. Probably the correct thing to do here is remove the ability to create or modify users from the REST API.
3. The user can be granted superuser access via the frontend. This is the main use case, and probably the only place the notification really needs to be sent from.

andyhd:
If we add new ways of making someone a superuser (why would we?), then the person who implements that code will be responsible for sending a notification. Passing the request user to the model save method is not straightforward, as we use DRF ViewSets and other abstractions to save the model. Adding code to pass the request user would likely end up more complicated than what is presented here. YAGNI + KISS

xoen:
Thanks for the comment that confirm what I thought about the reasons to have this in 3 places.

I wouldn't care about 1. too much and we could also probably ignore 2. as "In reality, this will likely never happen" and that leaves case 3. which cover most of the cases.

So maybe we can be happy with just having the notification sent if someone use the CP to make someone else a superuser - IMHO of course. That would mean less code.

andyhd:
Following our conversation, I'll use the `CrequestMiddleware` to get the requesting user (if possible)

r4vi:
@andyhd I was kind of hoping we'd eventually get rid of `CrequestMiddleware` :cry:

controlpanel/api/slack.py
```diff
@@ -0,0 +1,20 @@
+from django.conf import settings
+import slack
+
+
+CREATE_SUPERUSER_MESSAGE = "`{username}` was created as a superuser"
+GRANT_SUPERUSER_ACCESS_MESSAGE = "`{username}` was granted superuser status"
+
+
+def notify_team(message, request_user=None):
```
xoen: This function doesn't need to depend on the whole thing, it just needs a username.

To be honest I preferred this before when it was simpler and it was just getting a string (as it's potentially more reusable).

controlpanel/api/slack.py
```diff
@@ -0,0 +1,20 @@
+from django.conf import settings
+import slack
+
+
+CREATE_SUPERUSER_MESSAGE = "`{username}` was created as a superuser"
+GRANT_SUPERUSER_ACCESS_MESSAGE = "`{username}` was granted superuser status"
```
xoen: I think the "constants" are overkill, if we really want to have these as "constants" I'd prefer to have them close to the place where they're used as at the end of the day this `slack` module is not concerned with the specific messages sent, it just needs a string to sent to the team channel.

controlpanel/api/slack.py
```diff
@@ -0,0 +1,20 @@
...
+    token = settings.SLACK['api_token']
+    if token:
+        if request_user:
+            message = f"{message} by `{request_user.username}`"
+        slack.WebClient(token=token).chat_postMessage(
+            as_user=False,
+            username=f"Control Panel [{settings.ENV}]",
+            channel=settings.SLACK["channel"],
+            text=message,
+        )
```
xoen: So potentially the Slack `token` could be blank/not set but we don't want to blow up.
Maybe send a message is still worth and it may also help to document this design decision in the best scenario...
...and worst case scenario if we were expecting a notification in Slack and it didn't happen the logs would tell us why.

controlpanel/api/models/user.py
```diff
@@ -83,10 +83,28 @@ def reset_mfa(self):
...
+
         try:
-            User.objects.get(pk=self.pk)
+            existing = User.objects.get(pk=self.pk)
+
         except User.DoesNotExist:
             cluster.User(self).create()
+            if self.is_superuser:
+                slack.notify_team(
+                    slack.CREATE_SUPERUSER_MESSAGE.format(username=self.username),
```
xoen: these "constants" from the other module are probably overkill, IMHO.

controlpanel/api/models/user.py
```diff
@@ -83,10 +83,28 @@ def reset_mfa(self):
...
+                    request_user=request_user,
```
xoen: Makes more sense to change to `username=request_user.username` IMHO as the `slack.notify_team()` doesn't really need the coupling and dependency with the `User` model (or whatever type it's the request user).

controlpanel/api/slack.py
andyhd: I disagree. This module is the interface between the Control Panel and the Slack client. The Slack client is only concerned with sending strings to the channel. The rest of the Control Panel code is only concerned that a notification is sent. This module is concerned with how that notification is sent (via Slack) and how it is formatted.
The strings should be constants, because we test that they are sent and if we change the contents of the message (but not the reason they are sent), then we would also have to update the tests.
The constants belong in this module because they are Slack notifications, and not used for logging or flash messages, for example. If we add more Slack notifications on other events and kept the message strings alongside their use, we would have notification strings spread throughout the codebase instead of gathered here in the relevant Slack module.

controlpanel/api/slack.py
andyhd: I agree, I think I changed it to pass the user for some intermediate change to do with the other calling places that were then removed.

controlpanel/api/slack.py
andyhd: The problem is that if the settings module raises an exception getting the token, then tests cannot run. I realise now that this does not work either - I have a better solution coming.

controlpanel/api/slack.py
xoen: I think we disagree then.

> This module is concerned with how that notification is sent (via Slack) and how it is formatted

Not really because I can `import slack; slack.notify_team("I like icecream")` and that's perfectly legit from the slack module perspective.

If you want this module to be more specific and have more control over the kind of messages sent to slack this should pull more weight, e.g.

```python
def notify_made_superuser(granted_by, username):
    # generate "{granted_by} made {username} a superuser"
    # use constants or whatever
    msg = "..."
    _notify_team(msg)

def _notify_team(msg):
    # just the logic to send messages to Slack
    pass
```

If we go this route (which may have sense if you think the responsibility of this module is to control which notifications are sent to Slack) this means the logic to add the `by John Doe` would go in here rather than in the model and I'd like that.

xoen: Before I forget, is it actually possible for a superuser to create another superuser before they even logged in? I thought a user was created only when they logged in the first time (sorry if I'm missing something), if that's the case you may not need the `CREATE_SUPERUSER_MESSAGE` and you could possibly simplify the logic a bit.

controlpanel/api/slack.py
xoen: Yeah I get that you may not want to send the notification going during tests etc...having a warning in the logs seems sensible.

Awesome about the improved solution ☺️

controlpanel/api/models/user.py
xoen: see new comment on potentially move this formatting inside the `slack` module if you want to enforce the kind of notifications sent and keep them all in the same place.

controlpanel/api/models/user.py
```diff
@@ -83,10 +83,28 @@ def reset_mfa(self):
...
             cluster.User(self).create()
+            if self.is_superuser:
+                slack.notify_team(
+                    slack.CREATE_SUPERUSER_MESSAGE.format(username=self.username),
+                    request_username=request_username,
+                )
+
+        else:
+            if self.is_superuser and not existing.is_superuser:
+                slack.notify_team(
```
xoen: If we go for having this like this I think would be more clear to have the use of the `existing` variable closer to the query rather than in this `else` block, e.g.:

```python
try:
    existing = User.objects.get(pk=self.pk)
    if self.is_superuser and not existing.is_superuser:
        # stuff in `else` block
except User.DoesNotExist:
    # ... same
```

andyhd: No, putting more code into the `try...except` block is worse, because we are not handling any exceptions raised in that code.

controlpanel/api/slack.py
andyhd: You are right about having a function like `notify_superuser_created(username, by_username)` - I will do that.
It is possible to create a new user as a superuser before they login programmatically or via the shell - this would also create their k8s namespace and IAM role, etc. You would have to know their Github ID.

controlpanel/api/models/user.py
xoen: Yeah but the if/sent notification in the `try` is the happy path, that's not exception handling, that's normal flow.

To be honest I personally don't like overusing exceptions for normal control flow like this, this could be written without having to use `try`/`catch` and it could be more readable, e.g.:

```python
if existing = User.objects.filter(pk=self.pk).first():
  # send notification about existing user becoming superuser
else:
  # send notification about new superuser user created
```

IMHO of course.

andyhd: Using exceptions for flow control is Pythonic.

andyhd: But I agree, your way is more readable.

PR #771, opened 2019-11-06
IAM role creation works in cross-account setup

xoen:
A quirk of AWS cross-account setup is that the IAM role trust relationship
(AKA `AssumeRolePolicyDocument`)'s `Principal` is not the full ARN of the
assuming IAM role but only the ID of the account where the assumimg IAM
role is.

Once we have both data and compute AWS account IDs we can then tell
if everything is in the same AWS account or not (e.g. different IDs means
we have a cross-account setup).

In practice this means that CP would work as before in `alpha` (as long
as `AWS_COMPUTE_ACCOUNT_ID` and `AWS_COMPUTE_ACCOUNT_ID` as set correctly
to the same account ID where everything sits).

**IMPORTANT**: Removed now-ambiguous `AWS_ACCOUNT_ID` and added
`AWS_COMPUTE_ACCOUNT_ID` and `AWS_COMPUTE_ACCOUNT_ID` configuration.
We need to distinguish between the AWS account where the data/IAM roles
are and the AWS account k8s nodes (tools and apps) are.

**AWS delegate access example**: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_policy-examples.html#example-delegate-xaccount-S3

**Part of ticket**: https://trello.com/c/1DrJ8TN1

### DIFF
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -23,11 +23,32 @@ def s3_arn(resource):
     return arn("s3", resource)
 
 
-def iam_arn(resource, account=settings.AWS_ACCOUNT_ID):
+def iam_arn(resource, account=settings.AWS_DATA_ACCOUNT_ID):
     return arn("iam", resource, account=account)
 
 
-BASE_ROLE_POLICY = {
+def iam_assume_role_principal():
+    '''
+    returns the princical required to assume the IAM role
+
+    - ARN of the assuming role when both roles are in same account
+    - AWS account ID where the assuming IAM role is if in a different account
+
+    See AWS example: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_policy-examples.html#example-delegate-xaccount-S3
+    '''
+
+    cross_account = settings.AWS_COMPUTE_ACCOUNT_ID != settings.AWS_DATA_ACCOUNT_ID
+
+    if cross_account:
+        return settings.AWS_COMPUTE_ACCOUNT_ID
+
+    return iam_arn(
+        f"role/{settings.K8S_WORKER_ROLE_NAME}",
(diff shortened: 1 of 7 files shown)


controlpanel/api/aws.py
```diff
@@ -23,11 +23,22 @@ def s3_arn(resource):
...
+
+    # Principal must be source AWS account ID (not full ARN) in cross-account setup
+    if cross_account:
+        return settings.AWS_COMPUTE_ACCOUNT_ID
+
+    # Principal is full ARN of assuming IAM role in single-account setup
+    return iam_arn(f"role/{settings.K8S_WORKER_ROLE_NAME}")
+
+
+BASE_ASSUME_ROLE_POLICY = {
```
xoen: FYI: I've renamed this slightly to be more explicit about the fact this is not a "normal" IAM policy but instead is the "Assume Role Policy" (AKA "trust relationship" policy)

xoen:
@djsd123 I would have figured it out eventually but not as quickly without @sh31 ❤️ 🌟 

Don't imagine this is something specific of the library (boto).

Apparently when the assuming role in another account the `Principal` needs to be `{"AWS": "$ACCOUNT_ID"}` rather than `{"AWS": "$ROLE_ARN"}` 🤷‍♂ 

Probably (as pointed out by Serj) because in this case you're trusting the whole account to assume that role rather then a single specific IAM role in an account which you don't control (so to remove coupling)

xoen:
FYI: Merged [config](https://github.com/ministryofjustice/analytics-platform-config/pull/180) and [helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/404) changes (after rebasing/fixing conflicts/adjusting helm chart version/etc) so merging this. It shouldn't make any difference in old environments anyway.

PR #773, opened 2019-11-08
Add Google Analytics

calumabarnett:
Adds Google Analytics tracking. This is managed from the analytical_platform@digital.justice.gov.uk account.

https://trello.com/c/OrhTlqQE


controlpanel/frontend/jinja2/base.html
```diff
@@ -151,7 +151,16 @@
 
     <script src="{{ static('app.js') }}"></script>
     <script>window.moj.init();</script>
-
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-151666116-2"></script>
```
andyhd: You should remove the Google Analytics ID from here (and below) and use a template variable, taking the value from the Django settings module instead, eg:
```
<script async src="https://.../js?id={{ google_analytics_id }}"></script>
```
and in `controlpanel/frontend/jinja.py` add an item to the global template environment, eg:
```python
env.globals.update(
    {
        ...
        "google_analytics_id": settings.GOOGLE_ANALYTICS_ID,
    }
)
```
and in `controlpanel/settings/common.py`:
```python
...
GOOGLE_ANALYTICS_ID = "UA-..."
```
Not sure how sensitive the ID is, you might want to get a new one now that this has been published to Github.

calumabarnett: I don't think the ID is sensitive, so I don't think we need to get a new one. We can add filters in our Google Analytics account to ensure it only tracks traffic on our domains, in case someone finds it and tries to misuse it. Even if it is set as an environment variable, would it not be exposed in the source code anyway?

I've created different tracking IDs for alpha, prod and dev (old and new together). Not sure if I've implemented this in the right way though.

controlpanel/settings/common.py
```diff
@@ -555,3 +555,11 @@
...
 AIRFLOW_FERNET_KEY = os.environ.get('AIRFLOW_FERNET_KEY')
 
+
+# -- Google Analytics
+if ENV == 'alpha':
+    GOOGLE_ANALYTICS_ID = os.environ.get('GOOGLE_ANALYTICS_ID', 'UA-151666116-2')
+elif ENV == 'prod':
+    GOOGLE_ANALYTICS_ID = os.environ.get('GOOGLE_ANALYTICS_ID', 'UA-151666116-3')
+else:
+    GOOGLE_ANALYTICS_ID = os.environ.get('GOOGLE_ANALYTICS_ID', 'UA-151666116-4')
```
andyhd: I'm assuming these IDs are not sensitive - but generally the correct place to put things like API keys is in the https://github.com/ministryofjustice/analytics-platform-config encrypted repo, and then inject those values into the deployed environment in https://github.com/ministryofjustice/analytics-platform-helm-charts

andyhd: I think this will work for different envs.

PR #774, opened 2019-11-11
Enable bucket admins to revoke user access

calumabarnett:
The permission `api.delete_items3bucket` does not exist, so bucket admins are not able to revoke user access. This is not intended behaviour.

This PR enables bucket admins to revoke other bucket users' access.

See https://trello.com/c/dvRaJHt7.

### DIFF
--- a/controlpanel/frontend/jinja2/datasource-access-update.html
+++ b/controlpanel/frontend/jinja2/datasource-access-update.html
@@ -22,7 +22,7 @@
   </form>
 </section>
 
-{% if request.user.has_perm('api.destroy_items3bucket', items3bucket) %}
+{% if request.user.has_perm('api.destroy_users3bucket', users3bucket) %}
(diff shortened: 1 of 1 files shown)


controlpanel/frontend/jinja2/datasource-access-update.html
```diff
@@ -22,7 +22,7 @@ <h1 class="govuk-heading-xl">{{ page_title }}</h1>
   </form>
 </section>
 
-{% if request.user.has_perm('api.destroy_items3bucket', items3bucket) %}
+{% if request.user.has_perm('api.destroy_users3bucket', users3bucket) %}
```
andyhd: The context variable `items3bucket` is set in the view (in `controlpanel/frontend/views/datasource.py`), so this needs to remain `items3bucket`. The permission name change should work, though.

PR #776, opened 2019-11-13
Avoid infinite loop waiting for Tool deployment

andyhd:
Break out of the wait loop if the deployment status is anything other than
TOOL_DEPLOYING - this may result in some deployments being reported as failed
incorrectly, but should avoid the consumer getting stuck

### DIFF
--- a/controlpanel/frontend/consumers.py
+++ b/controlpanel/frontend/consumers.py
@@ -200,10 +200,10 @@ def start_background_task(task, message):
     )
 
 def wait_for_deployment(deployment, id_token):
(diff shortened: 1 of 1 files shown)


controlpanel/frontend/consumers.py
```diff
@@ -200,10 +200,10 @@ def start_background_task(task, message):
     )
 
 def wait_for_deployment(deployment, id_token):
-    while True:
```
xoen: RIP `while True` ⚰️ 👍

PR #777, opened 2019-11-18
Use MOJ Design System (remove GOV.UK logo)

andyhd:
* Also remove HMCTS Design System
* Change logout view to accept `GET` request and change disguised form button to link in header
* Fix references to deprecated GOV.UK Frontend colours


controlpanel/frontend/static/components/navbar/macro.html
```diff
@@ -1,19 +1,37 @@
...
-         {%- for key, value in item.attributes %} {{key}}="{{value}}"
-         {%- endfor %}>{{ item.html | safe if item.html else item.text }}</a>
-    </li>
-    {% endif %}
-  {% endfor %}
-  </ul>
-</nav>
+{% macro mojPrimaryNavigation(params) %}
+<div class="moj-primary-navigation {{- ' ' + params.containerClasses if params.containerClasses else ''}}">
+
```
s-block: Does this need so many blank lines?

controlpanel/frontend/jinja2/base.html
```diff
@@ -36,31 +35,48 @@
...
+        'label': 'Account navigation',
+        'items': [
+          {
+            'href': url('index'),
+            'text': request.user.name | default(request.user.email),
+            'active': True
+          },
+          {
+            'html': '<a href="' ~ url('oidc_logout') ~ '">Sign out</a>'
+,
```
s-block: Does this need to be here?

controlpanel/frontend/static/components/navbar/macro.html
andyhd: This is copied from [MOJ Frontend](https://github.com/ministryofjustice/moj-frontend/blob/master/src/moj/components/primary-navigation/template.njk) with minor changes to make it work with Jinja. Probably the fewer differences, the better.

controlpanel/frontend/jinja2/base.html
andyhd: Oops, typo

controlpanel/frontend/static/components/navbar/macro.html
s-block: :+1:

PR #779, opened 2019-11-21
Disable app logs temporarily

andyhd:


r4vi:
will this just hide the html but still make the call to kibana?

andyhd:
no, it comments out the call to kibana

PR #784, opened 2019-11-26
Only superusers may list all users

andyhd:


r4vi:
LGTM - not that user's being able to list everyone is a massive problem

PR #789, opened 2019-12-06
Fix bugs with path-specific bucket access

### LINKED ISSUE #785: Path-specific access is not removed from IAM policy
## Steps to reproduce
1. A user is granted access to `/folder_1/*` within a bucket
2. The user has their access to `/folder_1/*` removed but they are granted access to `/folder_2/*`

The user will still have access to `/folder_1/*`, that is, user IAM policies are not updated when path-specific access is modified.

### LINKED ISSUE #786: Path-specific access is not applied correctly
When a user is given path-specific access to a folder, say `/folder_1/*` as per the user guidance, their IAM policy will provide access to `arn:aws:s3:::bucket/folder_1/*/*`.

The user should have access to `arn:aws:s3:::bucket/folder_1/*`.

### LINKED ISSUE #787: Users cannot access objects in the S3 console when given path-specific access
When a user is given path-specific access they are unable to list files in the S3 console. This is because the `ListBucket` operation applies to a path resource not the bucket itself.

The `ListBucket` operation should apply to the whole bucket not just the specific path.

calumabarnett:
This PR:
- fixes #785: Path-specific access is not removed from IAM policy
- fixes #786: Path-specific access is not applied correctly
- fixes #787: Users cannot access objects in the S3 console when given path-specific access
- updates existing tests
- adds new tests

Linked to https://github.com/moj-analytical-services/user-guidance/pull/11.

### DIFF
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -291,10 +291,7 @@ class S3AccessPolicy:
                     resources.remove(resource)
 
     def grant_object_access(self, arn, access_level):
-        other_level = 'readwrite' if access_level == 'readonly' else 'readonly'
-
         self.add_resource(f'{arn}/*', access_level)
-        self.remove_resource(arn, other_level)
 
     def grant_list_access(self, arn):
         self.add_resource(arn, 'list')
@@ -357,8 +354,8 @@ def grant_bucket_access(role_name, bucket_arn, access_level, path_arns=[]):
     policy.put()
 
 
-def revoke_bucket_access(role_name, bucket_arn=None, path_arns=[]):
-    if not bucket_arn and not path_arns:
+def revoke_bucket_access(role_name, bucket_arn=None):
+    if not bucket_arn:
         log.warning(f'Asked to revoke {role_name} role access to nothing')
         return
 
@@ -426,8 +423,8 @@ def grant_group_bucket_access(group_policy_arn, bucket_arn, access_level, path_a
     policy.put()
 
 
-def revoke_group_bucket_access(group_policy_arn, bucket_arn=None, path_arns=[]):
-    if not bucket_arn and not path_arns:
+def revoke_group_bucket_access(group_policy_arn, bucket_arn=None):
+    if not bucket_arn:
         log.warning(f'Asked to revoke {group_policy_arn} group access to nothing')
         return
 
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -290,17 +290,19 @@ class S3AccessPolicy:
                 if resource.startswith(arn):
                     resources.remove(resource)
 
-    def grant_access(self, bucket_arn, access_level):
+    def grant_object_access(self, arn, access_level):
         other_level = 'readwrite' if access_level == 'readonly' else 'readonly'
 
-        self.add_resource(f'{bucket_arn}/*', access_level)
-        self.remove_resource(f'{bucket_arn}/*', other_level)
-        self.add_resource(bucket_arn, 'list')
+        self.add_resource(f'{arn}/*', access_level)
+        self.remove_resource(arn, other_level)
 
-    def revoke_access(self, bucket_arn):
-        self.remove_resource(f'{bucket_arn}/*', 'readonly')
-        self.remove_resource(f'{bucket_arn}/*', 'readwrite')
-        self.remove_resource(bucket_arn, 'list')
+    def grant_list_access(self, arn):
+        self.add_resource(arn, 'list')
+
+    def revoke_access(self, arn):
+        self.remove_resource(arn, 'readonly')
+        self.remove_resource(arn, 'readwrite')
+        self.remove_resource(arn, 'list')
 
     def put(self, policy_document=None):
         if policy_document is None:
@@ -349,23 +351,20 @@ def grant_bucket_access(role_name, bucket_arn, access_level, path_arns=[]):
     role = boto3.resource('iam').Role(role_name)
     policy = S3AccessPolicy(role.Policy('s3-access'))
     policy.revoke_access(bucket_arn)
+    policy.grant_list_access(bucket_arn)
     for arn in path_arns:
-        policy.grant_access(arn, access_level)
+        policy.grant_object_access(arn, access_level)
     policy.put()
 
 
 def revoke_bucket_access(role_name, bucket_arn=None, path_arns=[]):
-    if not path_arns:
-        if bucket_arn:
-            path_arns = [bucket_arn]
-        else:
-            log.warning(f'Asked to revoke {role_name} role access to nothing')
-            return
+    if not bucket_arn and not path_arns:
+        log.warning(f'Asked to revoke {role_name} role access to nothing')
+        return
 
     role = boto3.resource('iam').Role(role_name)
     policy = S3AccessPolicy(role.Policy('s3-access'))
-    for arn in path_arns:
-        policy.revoke_access(arn)
+    policy.revoke_access(bucket_arn)
     policy.put()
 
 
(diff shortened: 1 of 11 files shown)


controlpanel/api/aws.py
```diff
@@ -349,23 +351,20 @@ def grant_bucket_access(role_name, bucket_arn, access_level, path_arns=[]):
...
     policy = S3AccessPolicy(role.Policy('s3-access'))
     policy.revoke_access(bucket_arn)
+    policy.grant_list_access(bucket_arn)
     for arn in path_arns:
-        policy.grant_access(arn, access_level)
+        policy.grant_object_access(arn, access_level)
     policy.put()
 
 
 def revoke_bucket_access(role_name, bucket_arn=None, path_arns=[]):
```
xoen: I'm a bit confused by this method, was this causing problems? is the `path_arns` still relevant/used? a part from that in the `if` statement it doesn't seem like it's used.

Do the client code still pass this `path_arns` in and expect something to happen?

controlpanel/api/aws.py
```diff
@@ -290,17 +290,19 @@ def remove_resource(self, arn, sid):
...
 
-    def grant_access(self, bucket_arn, access_level):
+    def grant_object_access(self, arn, access_level):
         other_level = 'readwrite' if access_level == 'readonly' else 'readonly'
 
-        self.add_resource(f'{bucket_arn}/*', access_level)
-        self.remove_resource(f'{bucket_arn}/*', other_level)
-        self.add_resource(bucket_arn, 'list')
+        self.add_resource(f'{arn}/*', access_level)
+        self.remove_resource(arn, other_level)
```
xoen: Is it necessary to call `self.remove_resource` here? (genuine question) I see that the client code calls `policy.revoke_access(bucket_arn)` already before looping through the `path_arns` 🤷‍♂

tests/api/test_aws.py
```diff
@@ -293,7 +293,7 @@ def get_statements_by_sid(policy_document):
 @pytest.mark.parametrize(
     'resources',
     [
-        ([],),
+        ([]),
```
xoen: any reason to remove the comma? Isn't python gonna get confused and not understand this is a tuple?

tests/api/test_aws.py
```diff
@@ -303,39 +303,109 @@ def get_statements_by_sid(policy_document):
 )
 def test_grant_bucket_access(iam, users, resources):
     bucket_arn = 'arn:aws:s3:::test-bucket'
-    path_arns = [f'{bucket_arn}{resource}' for resource in resources]
+    path_arns_list = []
+    path_arns_object = []
+    if resources:
```
xoen: I'm not quite following this test but is the `if` necessary? You're using list comprehension so no need to initialise these two variables with an empty list, e.g. 

have you tried this without the `if resources:`?

```
path_arns_list = [f'{bucket_arn}{resource}' for resource in resources]
path_arns_object = [f'{bucket_arn}{resource}/*' for resource in resources]
```

tests/api/test_aws.py
```diff
@@ -303,39 +303,109 @@ def get_statements_by_sid(policy_document):
...
     aws.create_user_role(user)
 
-    aws.grant_bucket_access(user.iam_role_name, bucket_arn, 'readonly', path_arns)
+    aws.grant_bucket_access(user.iam_role_name, bucket_arn, 'readonly', path_arns_list)
 
     policy = iam.RolePolicy(user.iam_role_name, 's3-access')
     statements = get_statements_by_sid(policy.policy_document)
 
-    if path_arns:
+    if path_arns_object:
```
xoen: I'm not so sure about this test, a lot of `if` which mud the water.

I'm not sure what's the best way of writing this test but maybe you can add another parameter with the expected list of resources?

tests/api/test_aws.py
```diff
@@ -303,39 +303,109 @@ def get_statements_by_sid(policy_document):
...
-        ([],),
+        (['/foo/bar', '/foo/baz'])
+    ],
+    ids=[
+        'paths'
+    ]
+)
+def test_revoke_bucket_path_access(iam, users, resources):
+    bucket_arn = 'arn:aws:s3:::test-bucket'
+    path_arns = [f'{bucket_arn}{resource}' for resource in resources]
```
xoen: I suppose it doesn't matter but you may want to add a `/` between the `bucket_arn` and the `resource` in case this test fail as the output may be misleading.

tests/api/test_aws.py
```diff
@@ -345,7 +415,9 @@ def test_grant_bucket_access(iam, users, resources):
 )
 def test_revoke_bucket_access(iam, users, resources):
     bucket_arn = 'arn:aws:s3:::test-bucket'
-    path_arns = [f'{bucket_arn}{resource}' for resource in resources]
+    path_arns = []
+    if resources:
+        path_arns = [f'{bucket_arn}{resource}' for resource in resources]
```
xoen: again, probably it can be left as before without causing a problem as when `resources = []` `path_arns` should just be `[]`, don't think you need the `if`.

controlpanel/api/aws.py
calumabarnett: I was just thinking about this. I think it's superfluous for the reason you mentioned, so it can be removed.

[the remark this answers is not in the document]
calumabarnett: I agree using `self.remove_resource(f'{arn}/', other_level)` would fix the issue you described. Users cannot simultaneously have both read-only and read-write access to different paths under the current permissions model, so I'm not sure this is a problem.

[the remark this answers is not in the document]
calumabarnett: All permissions are revoked every time access is updated and only the specified paths (changed or not) will be added. I think the problem was that not all permissions were being revoked properly.

controlpanel/api/aws.py
calumabarnett: Yes, when bucket access was revoked, access to individual paths was still retained. `path_arns` is no longer used with these changes so it can be removed.

tests/api/test_aws.py
calumabarnett: No, in fact it's probably bad practice to. In this case, I don't think it matters, but I'll revert the change.

tests/api/test_aws.py
calumabarnett: I don't think this is necessary so I'll remove the `if`.

tests/api/test_aws.py
calumabarnett: I should apply this in the following line too.

tests/api/test_aws.py
calumabarnett: 👍

tests/api/test_aws.py
calumabarnett: Would we not always need an `if` statement to distinguish between the case when paths are specified and the case when they are not? Or would it be preferable to use an `if` statement to define a list of `expected_resources` so the assertion itself does not need to go within an `if` statement?

tests/api/test_aws.py
calumabarnett: Maybe I'm not understanding, but would that not cause the test to fail incorrectly with the specified resources?

tests/api/test_aws.py
xoen: FYI: I think without the comma Python *will* get confused and interpret that as a list rather than a tuple, e.g.:

```python
>>> ([],).__class__
<class 'tuple'> # correct

>>> ([]).__class__
<class 'list'> # probably not you want
```

tests/api/test_aws.py
xoen: You're probably right, TBH I can't remember much since that comment but by quick look at it it seems like the slash are already above - if I'll remember what I meant I'll let you know.

controlpanel/api/aws.py
```diff
@@ -357,8 +354,8 @@ def grant_bucket_access(role_name, bucket_arn, access_level, path_arns=[]):
     policy.put()
 
 
-def revoke_bucket_access(role_name, bucket_arn=None, path_arns=[]):
-    if not bucket_arn and not path_arns:
+def revoke_bucket_access(role_name, bucket_arn=None):
```
xoen: I didn't test it or checked properly but from quick look of it you may be able to make the `bucket_arn` argument mandatory without breaking compatibility with the client code.

That's simpler and less sneaky and it would allow you to also get rid of the `if not bucket_arn` and the warning for something that shouldn't really happen unless there is a bug (from what I can understand at least)

controlpanel/api/aws.py
```diff
@@ -426,8 +423,8 @@ def grant_group_bucket_access(group_policy_arn, bucket_arn, access_level, path_a
     policy.put()
 
 
-def revoke_group_bucket_access(group_policy_arn, bucket_arn=None, path_arns=[]):
-    if not bucket_arn and not path_arns:
+def revoke_group_bucket_access(group_policy_arn, bucket_arn=None):
```
xoen: Again, I didn't dive deep into this but from quick look it may be possible to make `bucket_arn` mandatory and remove the `if not bucket_arn` and associated warning.

Is it even possible to have a `RoleGroup` instance without an ARN? 😕

davidread:
@xoen thanks for reading through the code in detail. You or I should test this on dev - I suppose insufficient testing caused this problem in the first place. @calumabarnett have you deployed it there or perhaps one of us should help you with this?

calumabarnett:
@davidread I've not deployed it yet – can I do that without merging the PR?

davidread:
Good point. I see from the readme that you do indeed merge it to deploy to dev. Then one of us test. No doubt you'd be ok to do a release of controlpanel, but let us know if you need any help

controlpanel/api/aws.py
calumabarnett: I think this probably should be the case, though I'm not certain enough so I think it'd be safest to leave it as is. Perhaps, the `bucket_arn` could not exist if a bucket has been deleted in S3 but not in the control panel (?)

xoen: To answer your question the ARN should be generated using the AWS account ID and the bucket name so even when not present it shouldn't be blanck.

But of course it may not be passed to the function somehow - we can always change this later probably good call to be on the safe side 👍

PR #792, opened 2019-12-17
Fix delete app/user when IAM role doesn't exist

xoen:
When an app is deleted its IAM role is deleted as well.
Somehow sometimes this IAM role is not there, this shouldn't prevent
the deletion of the app anyway.

We were ignoring the `role.meta.client.exceptions.NoSuchEntityException`
but for some reason this is not what's actually raised by boto3 when
the IAM role is not there.

> NoSuchEntityException: An error occurred (NoSuchEntity) when calling
> the GetRolePolicy operation: The role with name alpha_app_XXX
> cannot be found.

The state of boto3 exceptions is appalling, see:
- https://github.com/boto/boto3/issues/1262
- https://stackoverflow.com/questions/46174385/properly-catch-boto3-errors

The most reliable way of catching these exception seems to be to catch
`botocore.exceptions.ClientError` (a low level boto3 exception) and check
its `e.response["Error"]["Code"]` is equal to the exception name you're
trying to handle. This is not pretty but at least seem to work.

Sentry exception: https://sentry.service.dsd.io/mojds/control-panel-api/issues/40251/
Ticket: https://trello.com/c/LSOKnEDZ

### DIFF
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -3,6 +3,7 @@ import json
 import logging
 
 import boto3
+import botocore
 from django.conf import settings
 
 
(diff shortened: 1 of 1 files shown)


xoen:
Thanks ☺️

PR #793, opened 2019-12-19
Fixed revoke access when IAM role doesn't exist

xoen:
When trying to revoke access to a bucket the code tries to load the
corresponsing IAM role's `s3-access` inline policy but if the IAM role
doesn't exist this would of course fail.

It's not clear what's causing this but this shouldn't prevent the operaion
to succeed as if there is no IAM role effectively the provided IAM role
(which doesn't exist) already doesn't have access to the bucket and hence
there is nothing to revoke.

Also, as when deleting a user/app role all the associated `AppS3Bucket`
and `UserS3Bucket` records are deleted this exception mean you can't
delete a user/app when it has access to something but no IAM role for
whatever reason.

Ticket: https://trello.com/c/vwgmnhBX

### DIFF
--- a/tests/api/test_aws.py
+++ b/tests/api/test_aws.py
@@ -359,6 +359,18 @@ def test_revoke_bucket_access(iam, users, resources):
     assert 'list' not in statements
 
 
+def test_revoke_bucket_access_when_no_role(iam):
+    role_name = "test_role_non_existent"
+    bucket_arn = "arn:aws:s3:::test-bucket"
+
+    # be sure role doesn't exist before calling revoke_bucket_access()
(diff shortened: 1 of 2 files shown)


tests/api/test_aws.py
```diff
@@ -3,6 +3,7 @@
 from unittest.mock import MagicMock, patch
 
 import boto3
+import botocore
```
calumabarnett: Do you need to import `botocore` here?

xoen: Good catch, I was using `ClientError` or whatever is called from `botocore` in this test file but then I changed it. Thanks, I'll remove 👍

PR #795, opened 2020-01-10
Updated dependencies to get new ca-certificates

xoen:
This will be required soon as Amazon RDS CA certificates will
expire on on March 5, 2020 (see link below) - AWS published new CA certificates on 19 Sep 2019 - I imagine they should be included in alpine's `ca-certificates=20191127-r0` (Nov 2019)

I've unpinned the version of `ca-certificates`, we should really be
interested in the latest we can get ideally (latest version is `20191127-r0`) - although as I've learned
this is linked to the alpine tag so it's not guaranteed we'll always
get the latest version of `ca-certificates` just by re-building the
docker image.

To satisfy the dependencies I had to:
- update the alpine base image tag to `3.11.2` (was `3.10`)
- update Python to `3.8.1-r0` (from `3.7.5-r1`)
- unpin some of the dependencies, to be fair we shouldn't really care
  about specific versions of `build-base`, `libffi-dev`, `libressl-dev`,
  `libstdc++` anyway

See: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html

Part of ticket: https://trello.com/c/SZCNhWm4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.10 AS base
+FROM alpine:3.11.2 AS base
 
 LABEL maintainer="andy.driver@digital.justice.gov.uk"
 
@@ -17,12 +17,12 @@ WORKDIR /home/controlpanel
 
 # install build dependencies
 RUN apk add --no-cache \
-        build-base=0.5-r1 \
-        ca-certificates=20190108-r0 \
-        libffi-dev=3.2.1-r6 \
-        python3-dev=3.7.5-r1 \
-        libressl-dev=2.7.5-r0 \
-        libstdc++=8.3.0-r0 \
+        build-base \
+        ca-certificates \
+        libffi-dev \
+        python3-dev=3.8.1-r0 \
+        libressl-dev \
+        libstdc++ \
         postgresql-dev \
         postgresql-client
 


xoen:
@sh31 / @calumabarnett I think this should avoid problems the newly published RDS CA certificates. I was tempted to update the DB in `dev` first and "break" it to then see if this solves the problem - AWS suggest the other way around: update application and then update DB as that would cause no downtime (which is of course what we want in `alpha`).

I'm tempted to go for the latter: merge this to `dev` and see if it breaks (I don't think it should but you never know with dependencies 🤷‍♂) and then update DB (and again it shouldn't break).

Serj, thoughts?

xoen:
To be honest I like the idea of "breaking" `dev` a bit more as we don't care much about downtime in `dev` and at least we can validate that these changes fix any possible RDS CA certificates problems.

xoen:
FYI: I've laid down a plan of attack in Trello ticket list: https://trello.com/c/SZCNhWm4

xoen:
FYI: We've updated the CP `dev` DB and things seem to still be working so we may not need this.

xoen:
It wasn't necessary. Closing this.

PR #797, opened 2020-03-04
When an S3 bucket is deleted, it's tagged for archival

xoen:
#### Before
When a bucket was deleted from the Control Panel only the record in the DB and IAM permissions to access it, was deleted. The S3 bucket and its data were left alone in AWS.

#### Now
Now when a bucket is deleted from the Control Panel is tagged as `to-archive=true` and also removed from the CP database.

The bucket's data will be archived/moved (by the [`buckets-archiver`](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/418) job) to a separate bucket and once the data is moved, the S3 bucket will be deleted.

### Why
Leaving "deleted" data behind forever is not great from a data retention perspective and also poses potential security problem if the same bucket is re-created in Control Panel by a separate user.

### How
Few things to note regarding the code changes:
- (in the `api.aws` module) we used to tag buckets on creation with the `buckettype=datawarehouse` tag (if they were datawarehouse buckets)
- this logic was moved into an helper function to be reused to tag on archival as well
- `api.aws.tag_bucket()` is part of the "public" interface of `api.aws` and it's higher level and receving the name of the bucket to tag and the tags
- `api.aws._tag_bucket()` is "private" and lower level, it receives a boto resource and performs the heavy lifting of translating the dictionary into the weird list of `{"Key": k, "Value": v}` format used by the AWS/boto Tagging API
- I've improved the existing tagging logic to merge new tags with existing ones to avoid losing old tags when tagging again. This makes `tag_bucket()` more predictable, it also means potentially we could write a migration to tag existing buckets with the standard buckets without losing the `buckettype` tag.
- the other noteworthy thing to notice is that the `controlpanel.api.models.s3bucket.S3Bucket#delete()` method is overridden to mark the bucket for archival on delete. This method is atomic as we want to rollback the DB transaction if for whatever reason the AWS request to tag the bucket failed - e.g. we don't want the bucket to be deleted from the DB but its data still hanging around because the bucket was not tagged as expected.
- existing tagging logic on creation wasn't tested, I've added a test for it

#### On the `cluster`/`models`/`aws` modules separation
The code affected by this PR is structured into 3 main modules with distinct responsibilities:
- `models` is responsible for persistence of data into the DB
- `cluster` is a more higher level module with the logic to propagate the CP changes into external systems (e.g. AWS/IAM/Kubernetes) - for example, when a bucket is created in the DB it also needs to be created in the AWS account
- `aws` is where the lower level logic to interact with AWS is, for example functions to create IAM roles, S3 buckets, tag buckets, etc...

### AWS permissions
[CP already has permissions to tag buckets](https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/control_panel_api/role.tf#L43) so this shouldn't require any additional permission.

**UPDATE**: ⚠️ I was testing this in `dev` and I actually need to add the `GetBucketTagging` permission as we were only able to update tags but we now read existing tags (error I'm getting: `An error occurred (AccessDenied) when calling the GetBucketTagging operation: Access Denied`)

### Related PRs/repositories
- `buckets-archiver`: https://github.com/ministryofjustice/analytics-platform-buckets-archiver
- `buckets-archiver` helm chart: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/418
- `buckets-archiver` helm chart config: https://github.com/ministryofjustice/analytics-platform-config/pull/190
- `buckets-archiver` S3 bucket/permissions: https://github.com/ministryofjustice/analytics-platform-ops/pull/311

### Ticket

Part of ticket: https://trello.com/c/6VdNPbzz

### DIFF
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -207,10 +207,7 @@ def create_bucket(bucket_name, is_data_warehouse=False):
             },
         )
         if is_data_warehouse:
-            bucket.Tagging().put(Tagging={'TagSet': [{
-                'Key': 'buckettype',
-                'Value': 'datawarehouse',
-            }]})
+            _tag_bucket(bucket, {"buckettype": "datawarehouse"})
 
     except bucket.meta.client.exceptions.BucketAlreadyOwnedByYou:
         log.warning(f'Skipping creating Bucket {bucket_name}: Already exists')
@@ -238,6 +235,11 @@ def create_bucket(bucket_name, is_data_warehouse=False):
         },
     )
     return bucket
+
+
+def _tag_bucket(boto_bucket, tags):
+    tag_set = [{"Key": k, "Value": v} for k, v in tags.items()]
+    boto_bucket.Tagging().put(Tagging={"TagSet": tag_set})
 
 
 class S3AccessPolicy:
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -237,6 +237,13 @@ def create_bucket(bucket_name, is_data_warehouse=False):
     return bucket
 
 
+def tag_bucket(bucket_name, tags):
+    """Add the given `tags` to the S3 bucket called `bucket_name`"""
+
+    bucket = boto3.resource("s3").Bucket(bucket_name)
+    _tag_bucket(bucket, tags)
+
+
 def _tag_bucket(boto_bucket, tags):
     """
     Tags the bucket with the given tags
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -238,8 +238,48 @@ def create_bucket(bucket_name, is_data_warehouse=False):
 
 
 def _tag_bucket(boto_bucket, tags):
-    tag_set = [{"Key": k, "Value": v} for k, v in tags.items()]
-    boto_bucket.Tagging().put(Tagging={"TagSet": tag_set})
+    """
+    Tags the bucket with the given tags
+
+    - `boto_bucket` is boto resource
+    - `tags` is a dictionary
+
+    NOTE: The tags provided are merged with existing tags (if any)
+
+    example:
+
+    existing tags : {"buckettype": "datawarehouse"}
+    new tags: {"to-archive": "true"}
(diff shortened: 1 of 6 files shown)


xoen:
FYI: Don't know if any of you had the chance to have a look. I was testing in `dev` and it turns out I need to add the `GetBucketTagging` permission to the CP role as we now read the tags before updating them. 

I'll do it as part of the related PR: https://github.com/ministryofjustice/analytics-platform-ops/pull/311

xoen:
FYI: I've updated the CP IAM role: https://github.com/ministryofjustice/analytics-platform-ops/pull/311/commits/21d6338c1c19f6b6c72a351a052c2fef81c77ec7

xoen:
FYI: Tested in `dev` (after applying [CP role tweak](https://github.com/ministryofjustice/analytics-platform-ops/commit/21d6338c1c19f6b6c72a351a052c2fef81c77ec7)) and it seems to have worked fine:
- bucket get tagged with `to-archive=true`
- then buckets-archiver moved the data and delete the bucket:

```
$ k logs buckets-archiver-1583339700-9kx94

[Wed Mar  4 16:35:05 UTC 2020]  About to archive 'dev-aldo-test-archiver-20200224' ...

[Wed Mar  4 16:35:08 UTC 2020]  Archiving 'dev-aldo-test-archiver-20200224' S3 bucket...
[Wed Mar  4 16:35:08 UTC 2020]  Moving data from 'dev-aldo-test-archiver-20200224' to 'dev-archived-buckets-data'...
move: s3://dev-aldo-test-archiver-20200224/test.txt to s3://dev-archived-buckets-data/dev-aldo-test-archiver-20200224/test.txt
[Wed Mar  4 16:35:09 UTC 2020]  Deleting 'dev-aldo-test-archiver-20200224' S3 bucket...
remove_bucket: dev-aldo-test-archiver-20200224
[Wed Mar  4 16:35:10 UTC 2020]  'dev-aldo-test-archiver-20200224' S3 bucket archived.
```

xoen:
Thanks for the review @dhirajnarwanimoj 👌

PR #799, opened 2020-04-02
On app delete, uninstall helm release and delete Auth0 resources

xoen:
### What

Currently, when an app is deleted, in addition to deleting its record in the DB we also delete its IAM role (which means any running pod for the app will no longer have any permissions).

In addition to this we now also delete its Auth0 Authorization group/roles/permissions (see below) and also the helm release (see below)

### Why
As the number of applications grows there may be applications that are no more needed and these can be deleted. Free up these related resources not only reduces clutter but also helps with the load on the k8s cluster. 

### Auth0 Authorization
We're using Auth0's Authorization extension to keep track of which customers can access an app (our users' users).
Each application has its own Auth0 group (roughly named after the repository name), each group has a role (`app-viewer`) and each role has a permission (`view:app`). 

This groups/roles/permissions relationships are very flexible and potentially a group may have more than one role associated with it - and similarly a role may have multiple permissions (the usual RBAC stuff)

However in our use-case each app has only a single group/role/permission so it's safe to delete these roles/permissions are they're not shared among several apps' groups.

These Auth0 resources are now deleted.

One important thing I've noticed while developing/testing this is that delete a role or permission would fail with a `400 BAD REQUEST` when there are groups/roles associated with this role/permission (unfortunately this was not documented which made building this more "fun").
This is the reason why deletion starts with the group, continues with the roles and finally deletes the permissions (basically try to delete the role before the group or try to delete the permission before the role would fail).

FYI: There is some documentation on this extension's API here: https://auth0.com/docs/api/authorization-extension

### helm release
When a release for an app is created Concourse will deploy an app by releasing the [`webapp`](https://github.com/ministryofjustice/analytics-platform-helm-charts/tree/master/charts/webapp) helm chart. 
As part of this PR the helm release is now uninstalled (which means the app will not be 
available anymore)

### Ticket
https://trello.com/c/nAIlHeAO

### DIFF
--- a/controlpanel/api/auth0.py
+++ b/controlpanel/api/auth0.py
@@ -127,12 +127,44 @@ class AuthorizationAPI(APIClient):
 
         return groups[0]
 
+    def _delete_group_roles(self, group_id):
+        """
+        Deletes all the roles attached to the group
+
+        See: https://auth0.com/docs/api/authorization-extension#delete-group-roles
+        """
+
+        if role_ids:
+            self.request(
+                "DELETE",
+                f"groups/{group_id}/roles",
+                json=role_ids,
+            )
+
     def delete_group(self, group_name):
-        """Deletes a group from the authorization API"""
+        """
+        Deletes a group from the authorization API
+
+        It also deletes all the roles associated with the
+        group.
+
+        NOTE: Roles in the Auth0 Authorization API are
+        meant to be flexible and they could potentially
+        be attached to different groups.
+        However, in our use-case, each app has 1 group
+        and each group has 1 role (`app-viewer`) so it's
+        safe to delete the (only) associated role.
+
+        See: https://auth0.com/docs/api/authorization-extension#delete-group
+        """
 
         group = self.get_group(group_name)
         if group:
-            self.request("DELETE", f'groups/{group["_id"]}')
+            group_id = group["_id"]
+            role_ids = group.get("roles", [])
+
+            self._delete_group_roles(group_id, role_ids)
+            self.request("DELETE", f"groups/{group_id}")
 
     def get_group_members(self, group_name):
         group = self.get_group(group_name)
--- a/controlpanel/api/auth0.py
+++ b/controlpanel/api/auth0.py
@@ -133,20 +133,6 @@ class AuthorizationAPI(APIClient):
             return group["_id"]
 
 
-    def _delete_group_roles(self, group_id, role_ids):
-        """
-        Deletes all the roles attached to the group
-
-        See: https://auth0.com/docs/api/authorization-extension#delete-group-roles
-        """
-
-        if role_ids:
-            self.request(
-                "DELETE",
-                f"groups/{group_id}/roles",
-                json=role_ids,
-            )
-
     def delete_group(self, group_name):
         """
         Deletes a group from the authorization API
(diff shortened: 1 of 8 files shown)


tests/api/permissions/test_app_permissions.py
```diff
@@ -96,5 +97,7 @@ def test_authenticated_user_has_basic_perms(client, users):
 def test_permission(client, app, users, view, user, expected_status):
     u = users[user]
     client.force_login(u)
-    response = view(client, app)
-    assert response.status_code == expected_status
+
+    with patch("controlpanel.api.views.models.App.delete") as _:
```
ntoll, line 101: I believe you don't need the `as _` here, you can just leave it off. Some linters / code quality checkers might complain about an unused object (while others might understand the `_` convention).

tests/api/views/test_app.py
```diff
@@ -87,10 +88,17 @@ def test_detail(client, app):
     assert set(userapp['user']) == expected_fields
 
 
-def test_delete(client, app, aws):
+@pytest.yield_fixture
+def authz():
+    with patch("controlpanel.api.cluster.auth0.AuthorizationAPI") as authz:
+        yield authz.return_value
```
ntoll, line 94: I think you could also do: `yield authz()` to have the same effect. The `return_value` is simply what you get when you call the object.

controlpanel/api/auth0.py
```diff
@@ -127,15 +127,63 @@ def get_group(self, group_name):
 
         return groups[0]
 
-    def get_group_members(self, group_name):
+    def get_group_id(self, group_name):
         group = self.get_group(group_name)
         if group:
+            return group["_id"]
```
ntoll, line 133: If we're programming defensively here (I don't know the code, but if the `"_id"` key is missing [is that likely?] you'll throw a `KeyError` here) you could rewrite the line as: `return group.get("_id")` which will return `None` if the key doesn't exist. Since you're checking the return value of this method when you call it in the code below, a `None` will allow the app to continue without doing anything meaningful without causing a 500-ish error.

tests/api/views/test_app.py
xoen: Cheers, I'll give it a go and see if it works 👍

tests/api/permissions/test_app_permissions.py
xoen: Cool, thanks for the tip, I'll give it a try and see if it works 👍

davidread: TIL!

controlpanel/api/auth0.py
xoen: Yeah in this case the `_id` key would always be there hence it's fine to have it blow up with `KeyError` when it's not. The Auth0 API and "contract" would be broken if this is missing so I'd be happy to raise an exception.

I've done what you've suggested [in another part](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/799/files#diff-4abd63b561b5260e074659dad215fa1cR166) where their API is behaving in an inconsistent way:

> role.get("permissions", [])

For some obscure reasons they decided to omit the `permissions` key when a Role didn't have any associated permission and include it when it did. In that case I definitely wanted to be more defensive and default to an empty list when that key was missing as semantically means "empty list of associated permissions" and we don't want to raise an exception in that case.

Sorry for the long comment, I hope it helps clarify ☺️

xoen:
FYI: 👆 rebased off `master`.

xoen:
Cool, thanks again for the review(s), tests are almost done and I expect them to pass, then as long as someone "formally" approves this I'm happy to merge/progress this ☺️ .

xoen:
FYI: Rebased `master` as it's slowly moving ahead...

... @davidread it would be good if you could review this 🙏

xoen:
FYI: I've merged this PR as it was hanging there for a long time.

PR #802, opened 2020-04-07
Disable app logs viewing

xoen:
### Why
View an app details was timing out because of the big amount of
logs produced.

### What
This is a temporary/short-term workaround which disable the printing
of these ElasticSearch logs.

Instead the following message is shown to try avoid confusing the user:

> Logs viewing currently disabled in Control Panel. See logs by following link below.

### Ticket
https://trello.com/c/cNoHc6pf


xoen:
FYI: @Mohammad-Tari / @ntoll we've put this workaround in place yesterday afternoon for that specific app, it's supposed to be temporary in nature but it's worth for you two to be aware of it.

PR #803, opened 2020-04-08
Merging this PR will update documentation for creating a local dev environment.

ntoll:
## What

This PR contains updates to the docs for setting up and running a local dev environment (so folks can work and test code while working remotely). 

## How to review

1. Just follow the steps in the `doc/running.md` file.
2. It should just work :tm: - you'll be able to start the Django app, log in and use various features.
3. ???
4. Profit!

### DIFF
--- a/doc/environment.md
+++ b/doc/environment.md
@@ -38,6 +38,7 @@
 | `SAML_PROVIDER` | the name of the SAML provider within AWS, which Auth0 integrates with, e.g. `dev-auth0`. This is referenced in user policies to allow them to log in via SAML federation. | |
 | `SECRET_KEY` | Secret key used to encrypt cookies, etc | |
 | `SENTRY_DSN` | Sentry credentials | |
+| `SLACK_API_TOKEN` | Slack token ([more information](https://slack.dev/python-slackclient/auth.html) | Mandatory, but doesn't need to be valid unless you're working with Slack |
 | `TOOLS_DOMAIN` | Domain where tools are deployed | |
 | `*_AUTH_CLIENT_DOMAIN` | OIDC domain for tool instances | [`OIDC_DOMAIN`]() |
 | `*_AUTH_CLIENT_ID` | OIDC client ID for tool instances | |
--- a/doc/running.md
+++ b/doc/running.md
@@ -10,6 +10,7 @@ Install python dependencies with the following command:
 python3 -m venv venv
 source venv/bin/activate
 pip3 install -r requirements.txt
+pip3 install -r requirements.dev.txt
 ```
 
 
--- a/doc/running.md
+++ b/doc/running.md
@@ -15,9 +15,28 @@ pip3 install -r requirements.txt
 
 ## Kubernetes setup
 
-You need a kubeconfig file (`~/.kube/config`) with the credentials for your kubernetes cluster.
+To run a Kubernetes cluster locally, the most convenient solution is probably
+`minikube`. Follow the
+[installation instructions for your OS](https://kubernetes.io/docs/tasks/tools/install-minikube/)
+and check you have something working:
 
-Once you've got that, this should work:
+```
+$ minikube status
+m01
+host: Running
+kubelet: Running
+apiserver: Running
+kubeconfig: Configured
+```
+
+You'll need a kubeconfig file (`~/.kube/config`) with the credentials for your
+kubernetes cluster. Starting `minikube` (e.g.
+`minikube start --driver=virtualbox`) should create these credentials for you
+for your local machine.
+
+You can check things are working with the following command, which will return
+information about the k8s master and KubeDNS:
+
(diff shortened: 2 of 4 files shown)


doc/environment.md
```diff
@@ -38,6 +38,7 @@
 | `SAML_PROVIDER` | the name of the SAML provider within AWS, which Auth0 integrates with, e.g. `dev-auth0`. This is referenced in user policies to allow them to log in via SAML federation. | |
 | `SECRET_KEY` | Secret key used to encrypt cookies, etc | |
 | `SENTRY_DSN` | Sentry credentials | |
+| `SLACK_API_TOKEN` | Slack token ([more information](https://slack.dev/python-slackclient/auth.html) | Mandatory, but doesn't need to be valid unless you're working with Slack |
```
xoen, line 41: 👍

doc/running.md
```diff
@@ -62,7 +122,9 @@ python3 manage.py migrate
 ```sh
 python3 manage.py createsuperuser
 ```
-NB `Username` needs to be your GitHub username
+
+Your `Username` needs to be your GitHub username.
+Your `Auth0 id` needs to be ??? (not working for me yet).
```
xoen, line 127: If you log into Auth0, click on the "Users & Roles" link, then find your own user you'll on the right of your avatar the `user_id:` field, the value will start with `github|`, that's your Auth0 ID and it will have to include the `github|` prefix.

ntoll: I tried exactly that (guessing), but it didn't work. :-( It's probably me, I'll try again.

doc/running.md
```diff
@@ -34,24 +64,55 @@ helm repo update
 
 ## <a name="env"></a>Environment variables
 
-At a mininum, you need to set the following environment variable:
+The simplest solution is to ask for a copy of a working `.env` file from one of
+the other developers with the envars set for development.
```
davidread, line 68: `envars` is a new one on me!

```suggestion
the other developers with the env vars set for development.
```
is maybe a more recognizable abbreviation? ;-)

doc/running.md
```diff
@@ -10,22 +10,52 @@ Install python dependencies with the following command:
...
 pip3 install -r requirements.txt
+pip3 install -r requirements.dev.txt
 ```
 
 
 ## Kubernetes setup
 
-You need a kubeconfig file (`~/.kube/config`) with the credentials for your kubernetes cluster.
+To run a Kubernetes cluster locally, the most convenient solution is probably
+`minikube`. Follow the
```
davidread, line 20: 'Docker for Mac' is probably more convenient for Mac users, so perhaps mention that as another possible option to explore.

doc/running.md
```diff
@@ -34,24 +64,55 @@ helm repo update
 
 ## <a name="env"></a>Environment variables
 
-At a mininum, you need to set the following environment variable:
+The simplest solution is to ask for a copy of a working `.env` file from one of
```
davidread, line 67: Emailing is not a great system. It's a bit of a social barrier and means you can't do this under time pressure, if the other dev is on holiday, or out of hours. If we could put it in a repo it would be better. It would allow us to track changes over time. Run tests with it. I can see there are no personal credentials in there, which is good. I can see there are a few system secrets and lots of implementation details that we might not want to share. How about the .env is checked into the [config repo](https://github.com/ministryofjustice/analytics-platform-config/) and encrypted with git-crypt like the other things?

davidread: The DevOps Handbook puts it better:

> we want developers to run production-like environment on their own workstations, created on demand and self-serviced. By doing this, developers can run and test their code in production-like environments as part of their daily work, providing early and constant feedback on the quality their work. ... Instead of merely documenting the specifications of the production environment in a document or on a wiki page, we crate a common build mechanism that creates all of our environments, such as for development, test, and production. By doing this, anyone can get production-like environments in minutes, without opening up a ticket, let alone having to wait weeks. ... All our requirements are embedded, not in documents or as knowledge in someone's head, but codified in our automated environment build process.

So I think that sharing .env files is not good practice. And if that is the best method we have right now, then let's mark it as such in the readme, and encourage people to improve on it in future, and suggest improvements such as:

a) a 'dev' mode that lets you run, but doesn't require extra setup
b) instructions of how to get the values from AWS, Auth0 etc
c) automated setup (hard)

ntoll: See #806 for a place to discuss requirements for this further.

ntoll:
@Mohammad-Tari is checking this PR as I type. I reckon it's ready to merge.

doc/running.md
davidread: I had the API call to hand that gets a github user ID, so thought I'd share: https://api.github.com/users/davidread So the corresponding Auth0 id is `github|307612`. Whether that's what this field needs, is another matter though.

PR #804, opened 2020-04-08
Merging this package with update python packages to latest versions as of 2020/04

ntoll:
## What

New `requirements.txt` and `requirements.dev.txt` so the latest version of Django works.

Why? Because these modules needed updating in order to set up a new local dev environment.

While this works on my **local** machine -- I've been able to start the app and log in, I've not been able to test any of the functionality (yet).

## How to review

1. Create a new virtualenv.
2. `pip install -r requirements.txt` then `pip install -r requirements.dev.txt`
3. Run the app and give it a spin.


davidread:
@ntoll I feel antsy seeing PRs hanging around more than a couple of days unmerged - it's the old 'lean' principle of aiming for small batch sizes. Also small PRs are easier to review and lower risk. So I'm keen to 'bank' the docs changes, if we can? And if there are more coming, they can be a separate PRs.

With this in mind, how about a separate PR for the dep upgrades? These seem a rather separate concern, which need testing and deployment, so fall naturally into a separate PR.

davidread:
Oh, I see there is a separate PR for the docs - https://github.com/ministryofjustice/analytics-platform-control-panel/pull/803 and just one of those commits is in this PR too. Makes more sense now!

ntoll:
Yeah... sorry, I started with the docs branch, then switched to this one to separate out the package updates so we have a clean separation of concerns, but obviously messed this up a little.

xoen:
FYI: At least one of the tests seem to fail. Something to do with the `kubernetes` package:

```python
AttributeError: module 'kubernetes.client' has no attribute 'apis'
```

I wanna poke around locally, maybe we need to avoid bumping kubernetes-related packages?

requirements.dev.txt
```diff
@@ -1,6 +1,6 @@
-django-debug-toolbar==1.11
+django-debug-toolbar==2.2
 django-debug-toolbar-requests==1.0.5
-django-elasticsearch-debug-toolbar==1.2.0
-pylint==2.3.1
-pylint-django==2.0.11
+django-elasticsearch-debug-toolbar==2.0.0
+pylint==2.4.4
+pylint-django==3.9.0
```
xoen, line 5: For some reason this version of `pylint-django` doesn't seem to exist 🤷‍♂️:

https://pypi.org/project/pylint-django/#history

Latest version seems to be `2.0.15` from the April 14, 2020.

requirements.txt
```diff
@@ -1,34 +1,34 @@
...
-gunicorn==19.9.0
+django-health-check==3.12.1
+django-redis==4.11.0
+djangorestframework==3.11.0
+djproxy==2.3.5
+elasticsearch-dsl==7.1.0
+gunicorn==20.0.4
 Jinja2>=2.10.1,<3.0
-kubernetes==10.0.1
+kubernetes==11.0.0
```
xoen, line 19: Looks like we should stick with `kubernetes==10.x`  because of a breaking change in their API.

I poke this to try and use `kubernetes==10.1.0` which should be compatible (at least by the look of the semantic version) but it seems like upgrading this requires an older version of `PyYAML` which is puzzling, see below:

```python
kubernetes 10.1.0 has requirement pyyaml~=3.12, but you'll have pyyaml 5.1.2 which is incompatible.
```

I say it's puzzling because `kubernetes==10.0.1` was happy with `PyYAML==5.1.2` but `kubernetes==10.1.0` seems to want an older version, weird 🤔 (`pyyaml~=3.12`)

xoen: Seem to have fixed the failing test ☺️

requirements.dev.txt
ntoll: OK... we should use 2.0.15 and assume `pylint-django` have retracted/messed up a version.

requirements.txt
ntoll: You :star:

requirements.dev.txt
xoen: yeah that was weird, how did we ended up with a version that doesn't even exist? 🤔 (unless they unpublish it or something) - oh well, I guess it doesn't matter.

requirements.txt
davidread: It sounds like the breaking change is a simple rename:

> kubernetes.client.apis package is renamed to kubernetes.client.api

From: https://github.com/kubernetes-client/python/blob/master/CHANGELOG.md#v1100a1

If that's the case, maybe we can simply change the name and use the latest version of `kubernetes` library?

ntoll: I'm :+1: on that suggested change from @davidread. Getting this ticket to a merge-able state is one of today's low-hanging-fruit tasks for me. :kiwi_fruit: :banana: :strawberry:

ntoll:
@xoen @davidread I've updated the `kubernetes` module to the latest version. I've also made the necessary changes to our code so the API renaming is accounted for. AFAICT, this needs checking on the `dev` environment (which I cannot do at this moment in time).

Aldo, as for the error you were reporting,

```
unauthorized_client: Callback URL mismatch. https://cpanel-master.services.dev.mojanalytics.xyz/oidc/callback/ is not in the list of allowed callback URLs
```

You mention the callback path has changed and it now has `/oidc/callback`  instead of old/working `/callback` - so it seems like one of the packages upgrade changed this somehow. I've taken a look into our `urls.py` (the ultimate source of truth for such things) and I see this:

```
path("oidc/", include("mozilla_django_oidc.urls")),
```

Which is how the `mozilla_django_oidc` app is "namespaced" to a path. Looking in the `urls.py` of `mozilla_django_oidc` I can see this:

```
url(r'^callback/$', OIDCCallbackClass.as_view(), name='oidc_authentication_callback'),
```

Which tells me that OIDC callback path is done via a reverse look-up of the `name` field, which Django's `urls` handling will always (and always should have) resolved to `oidc/callback`.

Basically, the behaviour and `urls.py` configuration I describe above is exactly what I'd expect from a Django app written in the last 5 years or so. I'm not sure what may have changed as a result.

xoen:
@ntoll you're right, it doesn't seem the app behaviour changed, still not 100% sure why I was getting the error in `dev` but it seems to be the case of old CP going to the callback URL via HTTP rather than HTTPs which was not in the list of "Allowed Callback URLs".

It's working in `dev` now, I'll also add the HTTPs callback URL for `/oidc` in the `alpha` application.

xoen:
Cool, _seems_ to run fine locally with the newer `kubernetes` version from what I can tell.

Happy to merge this, as said during our "how to deploy" chat once this is merged into `master` it gets deployed to `dev` (https://cpanel-master.services.dev.mojanalytics.xyz/tools/) for potential/advisable further manual testing. 

And once happy create a release in GH (following semantic versioning of course) will release to `alpha` (http://cpanel-master.services.alpha.mojanalytics.xyz/)

ntoll:
:shipit:

ntoll:
@xoen are you gonna merge and test this on `dev`..? Anything I can do to help..?

davidread:
Ah, our policy is for the PR author to do the merge, so go for it!

ntoll:
:+1:

xoen:
FYI: I was **quickly** checking this in `dev`, logged in, clicked around a bit, etc...and it seems to be working fine, or at the very least I didn't spot any obvious problems

So we can deploy this to `alpha` on Monday.

Great work, thanks again Nicholas ❤️

PR #805, opened 2020-04-14
Merging this PR will allow app admins to grant and remove app admin permission to other users.

ntoll:
## What

Describe what you have changed and *why*.

This is an initial fix for [this Trello ticket](https://trello.com/c/jxUgm3IP/570-app-admins-cant-grant-admin-permissions-to-other-users). Please see my questions on that ticket for more context, pasted below for convenience:

> Do we really want App admins to have this power? Imagine the following situation:
>
>  *  Alice is an app admin of project X.
>  *  Bob wants to become an app admin of project Y.
>  *  Alice, who has nothing to do with project Y, gives Bob the permission he needs because they're buddies. Bob is elevated to the "app admin" set of users.
>  *  Bob, for accidental or other reasons, gives Eve admin privileges to project X.
>  *  Alice wonders who on earth Eve is and what she's doing with her app.
>
> How can Alice stop Bob from giving accidental privileges to Eve..? By having the superuser group
> sitting above the level of app admins, with only super users able to elevate regular users to the 
> role of app admins, you avoid the (perhaps inevitable) result of everyone just elevating everyone 
> else to the role of app admin because it's just more convenient to work with those sets of privileges.
>
> Does this make sense? Also, I'm painfully aware that I'm probably missing something here. :-)

I've ensured that app-admins can also undo their action, by giving them the additional permission to remove app admin from a user. Please shout if I've assume the wrong thing here.

Finally, if the object passed into the  predicate function is None then the user was assumed to be a member of the `app_admin` group for any app (the default was to return `True`). While I realise the `obj` probably won't be `None`, this felt like the wrong default behaviour. We're essentially saying, if we don't have a reference to the app (as an `App`, `AppS3Bucket` etc) then we just assume the user **IS** an app admin. Why do we do this? My initial reaction is that it's a bit "generous". Surely, if we can't determine the user is associated with anything, then we should return `False`, as I've changed things in this PR.

Feels like I'm poking at something for which I should know more context given the above behaviour.

## How to review

I'm unsure, since I don't have a working dev environment. I'll update once this is fixed. However, I can't find any unit tests in our code which reference this sort of behaviour or the named permissions I've updated in the PR. Some guidance would be welcome. #newstarter :slightly_smiling_face:

### DIFF
--- a/controlpanel/api/rules.py
+++ b/controlpanel/api/rules.py
@@ -29,6 +29,9 @@ def is_app_admin(user, obj):
     if obj is None:
         return True
 
+    if is_superuser(user):
+        return True
+
     if isinstance(obj, App):
         return user in obj.admins
 
@@ -49,8 +52,8 @@ add_perm('api.update_app', is_authenticated & is_superuser)
 add_perm('api.destroy_app', is_authenticated & is_superuser)
 add_perm('api.add_app_customer', is_authenticated & is_app_admin)
 add_perm('api.remove_app_customer', is_authenticated & is_app_admin)
-add_perm('api.add_app_admin', is_authenticated & is_superuser)
-add_perm('api.revoke_app_admin', is_authenticated & is_superuser)
+add_perm('api.add_app_admin', is_authenticated & is_app_admin)
+add_perm('api.revoke_app_admin', is_authenticated & is_app_admin)
 add_perm('api.add_app_bucket', is_authenticated & is_superuser)
 add_perm('api.remove_app_bucket', is_authenticated & is_superuser)
 add_perm('api.view_app_logs', is_authenticated & is_app_admin)
--- a/controlpanel/frontend/static/components/app-logs/macro.html
+++ b/controlpanel/frontend/static/components/app-logs/macro.html
@@ -1,8 +1,11 @@
 {% macro app_logs(app, kibana_base_url) %}
   <div class="app-logs">
-    <pre>{% for entry in app.get_logs() -%}
+    <pre>{# for entry in app.get_logs() #}
+         {% for entry in [] -%}
 <span class="timestamp">{{ entry.timestamp.strftime('%Y-%m-%d %H:%M:%S') }}:</span> <span class="message">{{ entry.message }}</span>
-{% endfor %}</pre>
+{% endfor %}
+Logs viewing currently disabled in Control Panel. See logs by following link below.
+</pre>
   </div>
   <p class="govuk-body">
     <a href="{{ kibana_link(app, kibana_base_url) }}">View full logs</a>
--- a/controlpanel/frontend/static/components/app-logs/macro.html
+++ b/controlpanel/frontend/static/components/app-logs/macro.html
@@ -1,11 +1,8 @@
 {% macro app_logs(app, kibana_base_url) %}
   <div class="app-logs">
-    <pre>{# for entry in app.get_logs() #}
-         {% for entry in [] -%}
+    <pre>{% for entry in app.get_logs() -%}
 <span class="timestamp">{{ entry.timestamp.strftime('%Y-%m-%d %H:%M:%S') }}:</span> <span class="message">{{ entry.message }}</span>
-{% endfor %}
-Logs viewing currently disabled in Control Panel. See logs by following link below.
-</pre>
+{% endfor %}</pre>
   </div>
   <p class="govuk-body">
     <a href="{{ kibana_link(app, kibana_base_url) }}">View full logs</a>
--- a/tests/frontend/views/test_app.py
+++ b/tests/frontend/views/test_app.py
@@ -166,11 +166,11 @@ def connect_bucket(client, app, _, s3buckets, *args):
         (delete, 'normal_user', status.HTTP_403_FORBIDDEN),
 
         (add_admin, 'superuser', status.HTTP_302_FOUND),
-        (add_admin, 'app_admin', status.HTTP_403_FORBIDDEN),
+        (add_admin, 'app_admin', status.HTTP_302_FOUND),
         (add_admin, 'normal_user', status.HTTP_403_FORBIDDEN),
 
         (revoke_admin, 'superuser', status.HTTP_302_FOUND),
-        (revoke_admin, 'app_admin', status.HTTP_403_FORBIDDEN),
+        (revoke_admin, 'app_admin', status.HTTP_302_FOUND),
         (revoke_admin, 'normal_user', status.HTTP_403_FORBIDDEN),
 
         (add_customers, 'superuser', status.HTTP_302_FOUND),


xoen:
@ntoll Will have a look, but by reading at your description/questions:

> Do we really want App admins to have this power?

Probably a question for our PM (@samtazzyman ) but my "feeling" is that if someone is an admin for a specific app we trust that person to be responsible for it.
E.g. if they wants to grant access to this app/its data to someone else it shouldn't be a problem (although they should be of course careful)
But of course maybe my assumption is wrong and we want to keep a tighter control over access to apps. Of course if that's the case it means we'll still rely on superusers to do things on behalf of app admins (current situation AFAICT).

> While I realise the obj probably won't be None, this felt like the wrong default behaviour. We're essentially saying, if we don't have a reference to the app (as an App, AppS3Bucket etc) then we just assume the user IS an app admin. Why do we do this? My initial reaction is that it's a bit "generous". Surely, if we can't determine the user is associated with anything, then we should return False, as I've changed things in this PR.

Not sure neither. But (without having look at the code or diff) I tend to agree with you we should generally err on the side of caution when it comes to permissions. E.g. deny if unsure (as you said you changed 👌 )

> I can't find any unit tests in our code which reference this sort of behaviour or the named permissions I've updated in the PR.

I think some tests failed so I guess you got your answer ☺️ 


PS: Great that you poke around permissions 🏆

ntoll:
Aha... thanks @xoen for the heads up. @samtazzyman any thoughts on the questions ^^^..? Thanks in advance. :+1:

controlpanel/api/rules.py
```diff
@@ -27,6 +27,9 @@ def is_app_admin(user, obj):
     :param obj App|AppS3Bucket|UserApp: The object to check
     """
     if obj is None:
+        return False  # TODO: Check this change is appropriate.
```
xoen, line 30: I think this makes sense as it's more conservative and assumes someone is not an app admin when the "app" is `None` 👍 

I wonder if this condition could be `if not obj:`? `obj is None` seems very specific. 
As you said it's a bit of an edge case but you never know with these things. For "sure" returning `False` is a better default.

samtazzyman: Thanks for this, you raise some good points about Alice, Bob, and Eve. I think that as long as we have a record of who added whom and when (which we presumably do), I am content for admins to be able to create other admins. This is because:
 - I think the risk of Eve getting added accidentally is probably not that large
 - As it stands there is a burden of admin on superusers for exactly this reason
 - It's particularly a pain when someone has left the organisation without handing over admin duties (because they couldn't do so)
 - The superusers are further from projects and so less able to judge the suitability of Eve as an app admin, and more likely to just grant her the role in exchange for not seeming to be a blocker or jobsworth

controlpanel/api/rules.py
```diff
@@ -49,8 +52,8 @@ def is_app_admin(user, obj):
 add_perm('api.destroy_app', is_authenticated & is_superuser)
 add_perm('api.add_app_customer', is_authenticated & is_app_admin)
 add_perm('api.remove_app_customer', is_authenticated & is_app_admin)
-add_perm('api.add_app_admin', is_authenticated & is_superuser)
-add_perm('api.revoke_app_admin', is_authenticated & is_superuser)
+add_perm('api.add_app_admin', is_authenticated & is_app_admin)
```
xoen, line 55: This makes sense to me.

I'm trying to wrap my head around whether we could end up in the situation where a user Alice could be admin of the App "Foo" but not have access (whether is admin or not) for the corresponding app bucket "Foo" and whether that would be a problem or not. Maybe not. 

Maybe the granting admin will also have to grant permission to the app bucket as well if the wants to give access to the data. It may not be a bad thing as it would be more _explicit_ and potentially in the future we may have users who can admin an app (e.g. deploy/restart it or something) but not have access to the data. I don't know, I'm thinking "loudly" here.

I think the permission to grant access to the data is called `api.grant_s3bucket_access` and bucket admins can already grant permissions to other users.

xoen: (sorry, stream of consciousness comment 🤣 )

ntoll: https://www.osnews.com/story/19266/wtfsm/ :+1: :wink:

controlpanel/api/rules.py
ntoll: @samtazzyman ack on the points you make. For me it was more a case of everyone just elevating everyone else to the role of app admin because it's just more convenient to work with those sets of privileges (in which case, why have app admins). However, from you say, I guess this probably won't happen given the current practices in the org.

samtazzyman:
Thanks for this, you raise some good points about Alice, Bob, and Eve. I think that as long as we have a record of who added whom and when (which we presumably do), I am content for admins to be able to create other admins. This is because:

I think the risk of Eve getting added accidentally is probably not that large
As it stands there is a burden of admin on superusers for exactly this reason
It's particularly a pain when someone has left the organisation without handing over admin duties (because they couldn't do so)
The superusers are further from projects and so less able to judge the suitability of Eve as an app admin, and more likely to just grant her the role in exchange for not seeming to be a blocker or jobsworth

controlpanel/api/rules.py
samtazzyman: I think it's sensible to have separate administrative cover for app and bucket - there could conceivably be situations where someone needs to be in charge of the one but not the other.

controlpanel/api/rules.py
xoen: @ntoll bear in mind something that you probably already know from previous comment, someone is an app admin for a **specific** app.

Being an "app admin" is more of a state (as in there is a `UserApp` record with `admin=True`) rather than a "role" as in the traditional sense as in "someone is an app admin and therefore can admin any app". 

Hence make someone an app admin out of laziness wouldn't have an app-specific blast radius.

controlpanel/api/rules.py
xoen: @samtazzyman yeah that was also the conclusion I reached by talking with myself 🤣

ntoll: OK... so AFAICT, given the `api.grant_s3bucket_access` permission and the change in this PR, we're covered in terms of the situation as described by @samtazzyman..? (I'm painfully aware that I don't have the org context, depth of knowledge and this is my first PR so I don't want to make assumptions). :+1: Thanks!

xoen: @ntoll yeah bucket admins can already `api.grant_s3bucket_access` and you tweaked the permission to `api.add_app_admin` so I think your original changes should be enough once tests are also fixed/adjusted ☺️

ntoll: > once tests are also fixed/adjusted

Easier said than done..! :wink:

xoen: Sorry 😁!! The medium doesn't help, didn't mean to be pushy or put pressure on you of course ☺️

ntoll:
So I've fixed the tests.

I reverted my change to handling a `None` value passed in as `obj` into the predicate. It turns out this does have a side effect that we want to keep. I'm not sure why it's written this way (it feels counter-intuitive to me), but it was causing tests to fail. I also rebased on the branch which updates the package versions (I was getting the old kubernetes error because the code was using the old version of the API).

@xoen ready for further review.

xoen:
let me have another look, could you rebase off `master` please 🙏 as it makes reviewing this much easier ☺️  (without the commits from other branches which already merged into master)

ntoll:
@aldo rebased

davidread:
I quick bit of commit squashing, to get a neater commit history, would be helpful

controlpanel/api/rules.py
```diff
@@ -29,6 +29,18 @@ def is_app_admin(user, obj):
...
+        return True
+
+    if is_superuser(user):
+        return True
+
+    if is_superuser(user):
+        return True
+
+    if is_superuser(user):
+        return True
```
xoen, line 42: This same `if` statement is repeated for some reason, maybe an editor glitch? 🤷‍♂️

controlpanel/api/rules.py
```diff
@@ -29,6 +29,18 @@ def is_app_admin(user, obj):
     if obj is None:
         return True
 
+    if is_superuser(user):
+        return True
```
xoen, line 33: I see why this was added.

It makes me wonder if this is the right thing to do semantically, e.g. I may be a super user but not an admin of a certain app. 

A part from the semantic of it, maybe there are not side effects of returning true for super admins here.

The annoying thing of not adding this here is that the predicated which was changed from

```python
add_perm('api.add_app_admin', is_authenticated & is_superuser)
```

to 

```python
add_perm('api.add_app_admin', is_authenticated & is_app_admin)
```

would have to have an OR to maintain compatibility for superusers if you know what I mean.

This is probably the simplest solution but I thought it was worth highlight this.

controlpanel/api/rules.py
```diff
@@ -49,8 +61,8 @@ def is_app_admin(user, obj):
 add_perm('api.destroy_app', is_authenticated & is_superuser)
 add_perm('api.add_app_customer', is_authenticated & is_app_admin)
 add_perm('api.remove_app_customer', is_authenticated & is_app_admin)
-add_perm('api.add_app_admin', is_authenticated & is_superuser)
-add_perm('api.revoke_app_admin', is_authenticated & is_superuser)
+add_perm('api.add_app_admin', is_authenticated & is_app_admin)
+add_perm('api.revoke_app_admin', is_authenticated & is_app_admin)
```
xoen, line 65: Random thought (not related to your changes): These predicates seems to repeat `is_authenticated ` a lot, maybe it's to be safe in case the other predicated forget to check the user is authenticated before carrying one with other checks?

From an high level perspective of course is a user is a `superuser` for example, I'd imply that it's also authenticated but logic and the code may disagree 😬

controlpanel/api/rules.py
ntoll: Huh.... that's weird. Yeah, it's probably vim playing up. ;-) I'll fix an push.

controlpanel/api/rules.py
ntoll: IIRC `is_authenticated` is built into the module we're using and we're creating new predicates by explicitly combining two existing ones (one built into the module, the other defined by us).

controlpanel/api/rules.py
ntoll: OK... fixed. I also rebased with master and force pushed.

ntoll:
@davidread squashed. Will merge when CI turns green.

PR #808, opened 2020-04-27
Dockerfile: unpin `ca-certificates` to avoid problems

xoen:
### What
Use latest stable version of alpine's `ca-certificates` package in `Dockerfile`.

### Why
Alpine released a new version of this package which caused a
docker build problem because of the fact this package was pinned in
our `Dockerfile`:

```
  ca-certificates-20191127-r0:
    breaks: world[ca-certificates=20190108-r0]
```

There shouldn't be any specific reason to not use the latest version
of `ca-certificates` anyway.

### How to review

Check that tests and quay.io builds are fine.

### Ticket
https://trello.com/c/M1mhemDt

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ WORKDIR /home/controlpanel
 # install build dependencies
 RUN apk add --no-cache \
         build-base=0.5-r1 \
-        ca-certificates=20190108-r0 \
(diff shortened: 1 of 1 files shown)


ntoll:
LGTM :+1: :shipit:

PR #810, opened 2020-05-13
Uninstall old RStudio (tools) release before helm upgrade

xoen:
### What
Fix problem where some users with an old RStudio instance fail to upgrade.

Uninstall old, "badly"-named, release of a tool before installing/upgrading the new one.

Tools releases are now `TOOL-USERNAME` while old RStudio instances were installed on `USERNAME-rstudio`, causing problem below.

### Why
Due to the change in the release naming scheme of RStudio, old users
can't upgrade their RStudio from the Control Panel because of an
underlying helm error.

[This can be manually resolved](https://github.com/ministryofjustice/analytics-platform/wiki/Deploying-R-Studio#2-release-problem) by running `helm delete --purge` on
old release but it's annoying from a support and UX standpoint.

After this change the CP will check if there is a release with
the old naming scheme (e.g. `USERNAME-rstudio` instead of the correct
`rstudio-USERNAME` name) and delete it before installing the new release.

Ticket: https://trello.com/c/llxQG72F

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -232,7 +232,32 @@ class ToolDeployment():
     def release_name(self):
         return f"{self.chart_name}-{self.user.slug}"
 
-    def install(self, **kwargs):
+    def _delete_legacy_release(self):
+        """
+        At some point the naming scheme for RStudio
+        changed. This cause upgrade problems when
+        an old release with the old release name is
+        present.
+
+        We're going to temporarily check/uninstall
+        releases with the old name before installing
+        the new release with the correct name.
+
+        We can remove this once every user is on new naming
+        scheme for RStudio.
+        """
+
+        old_release_name = f"{self.user.slug}-{self.chart_name}"
+        if old_release_name in helm.list_releases(old_release_name):
+            helm.delete(True, old_release_name)
+
+
+    def _set_values(self, **kwargs):
+        """
+        Return the list of `--set KEY=VALUE` helm upgrade arguments
+
+        Extracted from `install()` method for clarity.
+        """
         values = {
             "username": self.user.username.lower(),
             "Username": self.user.username.lower(),  # XXX backwards compatibility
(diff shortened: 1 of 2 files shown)


controlpanel/api/cluster.py
```diff
@@ -252,7 +274,24 @@ def install(self, **kwargs):
...
+        found = helm.list_releases(self.old_release_name)
+        for release in found:
+            # helm list if very fuzzy and could
+            # return other releases containing the release
+            # name we're looking for.
+            # To avoid accidentally uninstall a different
+            # release we check equality here.
+            if release == self.old_release_name:
+                helm.delete(True, self.old_release_name)
+                break
```
xoen, line 290: I may actually move this little bit of logic into a `_delete_legacy_release()` helper function.
This way I can also get rid of that `old_release_name` property only used by this and use a simple variable.

xoen: For the record, this should have been a "private" property called `_old_release_name` anyway instead to signal the fact it is not to be used outside this module. (but it doesn't matter as I'm going to get rid of it anyway)

controlpanel/api/cluster.py
```diff
@@ -232,7 +232,39 @@ def k8s_namespace(self):
...
+        releases with the old name before installing
+        the new release with the correct name.
+
+        We can remove this once every user is on new naming
+        scheme for RStudio.
+        """
+
+        old_release_name = f"{self.user.slug}-{self.chart_name}"
+        found = helm.list_releases(old_release_name)
+        for release in found:
```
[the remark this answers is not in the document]
xoen: Yes! Thanks @ntoll - yeah much better and I'm surprised I didn't do like that to begin with 😬

xoen:
FYI: Also just rebased `master`, as there were some changes there.

xoen:
Thanks for the review again, I'm merging this to do a bit more manual testing in `dev` before I roll this out.

PR #811, opened 2020-05-15
Avoid a 500 error if the app admin user has no Auth0 id.

ntoll:
## What

This afternoon we had reports that some users were experiencing 500 errors from the application. A look at the logs turned up this error:

```
cpanel-5db49bc644-qwdjr backend   File "/home/controlpanel/controlpanel/frontend/jinja2/webapp-detail.html", line 160, in block "content"
cpanel-5db49bc644-qwdjr backend     href="{{ url('manage-user', kwargs={ "pk": user.auth0_id }) }}">
cpanel-5db49bc644-qwdjr backend   File "/usr/lib/python3.7/site-packages/django/urls/base.py", line 87, in reverse
cpanel-5db49bc644-qwdjr backend     return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))
cpanel-5db49bc644-qwdjr backend   File "/usr/lib/python3.7/site-packages/django/urls/resolvers.py", line 677, in _reverse_with_prefix
cpanel-5db49bc644-qwdjr backend     raise NoReverseMatch(msg)
cpanel-5db49bc644-qwdjr backend django.urls.exceptions.NoReverseMatch: Reverse for 'manage-user' with keyword arguments '{'pk': ''}' not found. 1 pattern(s) tried: ['users/(?P<pk>[^/]+)/$']
```

It turns out that some users added as app admins have not, perhaps, already logged into Auth0, so their Auth0 id is `None`. Hence the URL created in the template resolves to `users/` rather than `users/1234`. This is what leads to the `NoReverseMatch` exception.

This PR simply adds a guard around the block of could that could produce such a URL and ensures the username is displayed, but since there's not (yet) an Auth0 id associated with the user, you can't click on their name to edit their details.

## How to review

This is a minor change of an added conditional. We'd need to try to re-create a problem user in the `dev` environment before pushing to `alpha`. This can be done as part of the review process.

cc/@andylightfoot

### DIFF
--- a/controlpanel/frontend/views/app.py
+++ b/controlpanel/frontend/views/app.py
@@ -1,4 +1,5 @@
 import re
+import logging
 
 from django.conf import settings
 from django.contrib.auth.mixins import LoginRequiredMixin
@@ -36,6 +37,9 @@ from controlpanel.frontend.forms import (
 )
 
 
+log = logging.getLogger(__name__)
+
+
 class AppList(LoginRequiredMixin, PermissionRequiredMixin, ListView):
     context_object_name = 'apps'
     model = App
@@ -68,6 +72,11 @@ class AppDetail(LoginRequiredMixin, PermissionRequiredMixin, DetailView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         app = self.get_object()
+
+        # Log data consistency warnings. (Missing Auth0 ID).
+        for user in app.admins:
+            if not user.auth0_id:
+                log.warning("User without Auth0 ID, {}, is admin for app: {}.".format(user, app))
 
         context["app_url"] = cluster.App(app).url
         context["admin_options"] = User.objects.filter(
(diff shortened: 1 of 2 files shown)


xoen:
@ntoll it makes me wonder if this is related to the Auth0 user for the Concourse non-interactive user which has a blank Auth0 ID has it doesn't have one. 

Still strange as in theory this "user" shouldn't be the admin of any app. Maybe something else sometimes goes wrong and a blank ID for the user is posted/persisted in the DB for some apps?
I don't even know if that's the case or how that could happen.

Theories aside, guard agains this should prevent the user getting an exception when accessing the webapp page.

Don't know if you found anything else by inspecting the data in the DB.

ntoll:
OK... I'll try to add some logging in the view.

ntoll:
@davidread added a guard and logging.

ntoll:
I've added another guard, so only users with an `auth0_id` can be added as app admins - as a result, we shouldn't have corrupt data in the DB.

ntoll:
@xoen thanks for the approval. It's late in the afternoon and deploying at the end of the day makes me nervous. I'll merge in the morning. :+1:

controlpanel/frontend/views/app.py
```diff
@@ -291,13 +300,16 @@ class AddAdmin(UpdateApp):
...
+        if user.auth0_id:
+            userapp = UserApp.objects.create(
+                app=app,
+                user=user,
+                is_admin=True,
+            )
+            userapp.save()
+            messages.success(self.request, f"Granted admin access to {user.name}")
+        else:
+            messages.error(self.request, "Failed to grant admin access to {user.name} because they lack an Auth0 ID.")
```
davidread, line 312:
```suggestion
            messages.error(self.request, f"Failed to grant admin access to {user.name} because they lack an Auth0 ID.")
```

ntoll: @davidread aha... I wondered about `f` strings and opted to go for the conservative option. Will revise.

ntoll: OK fixed, squashed and pushed.

PR #812, opened 2020-05-20
Show deployed tools versions

xoen:
### What
Show users the current version of the tools they're using.

This is **not** the version of the helm chart but it's meant to show the version of the packaged tool.

This is part of the ongoing improvements to the Tools Catalogue.

Screen Shot 2020-05-20 at 10 10 48

### How
We use the helm repository index as source of true when it comes to version of tools installed.
Helm charts have some metadata and we've been using the `appVersion` field to record the version of the underlying tool (e.g. RStudio 1.2, R 3.4, etc...).

Unfortunately this information is partial and we rely on the helm repository metadata to be accurate.

Furthermore the `appVersion` field was added only "recently" and for very old charts this is not available. When the version can't be determined we show "Unknown" as version.

Ticket: https://trello.com/c/74QeufKf

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -20,6 +20,7 @@ TOOL_DEPLOY_FAILED = 'Failed'
 TOOL_IDLED = 'Idled'
 TOOL_NOT_DEPLOYED = 'Not deployed'
 TOOL_READY = 'Ready'
+TOOL_RESTARTING = 'Restarting'
 TOOL_UPGRADED = 'Upgraded'
 TOOL_STATUS_UNKNOWN = 'Unknown'
 
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -345,6 +345,23 @@ class ToolDeployment():
 
         return deployments[0]
 
+
+    def get_installed_chart_version(self, id_token):
+        """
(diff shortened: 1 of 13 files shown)


xoen:
FYI: Rebased `master`.

xoen:
@ntoll / @Mohammad-Tari I've rebased and also made some changes so that CP dynamically updates the new version of a tool on the "Tools" page.

As I said, testing was light on these parts so I'll probably spend some time to add some tests but in the meantime it would be good if you start to have a look and see if this makes sense, etc...(bear in mind I'm not that familiar with this part of the CP myself so I tried to keep changes to the minimum without moving more than I thought was required).

xoen:
@Mohammad-Tari / @ntoll I've pushed some tests for the `frontend.consumers` module (this wasn't tested at all as far as I can tell)

Have a look at this PR when you get to it, we could deploy next week if it makes sense to you and I can continue to work on the tools catalogue epic from here.

Thanks in advance ☺️

xoen:
FYI: Rebased `master` after a couple of PRs were merged, no conflicts and all tests passed locally which is good 👌

xoen:
OK, tested this in `dev` and it seems to work fine (as it was locally) so I'm going to merge this and possibly create a release soon.

PR #816, opened 2020-05-26
Add versioning and simple lifecycle to S3 buckets.

ntoll:
## What

This PR adds versioning and a simple life-cycle to S3 buckets configured by the AP control panel.

The related ticket in Trello is here: https://trello.com/c/g5NFH4Mi/588-s3-bucket-backups-versioning

What's the use case?

* Small accidental deletes - mojap-land and other data engineering buckets are versioned, so data coming in that way has a back up, and all warehouse and app data based off that should be reproducible, so not too concerned about accidental deletion of a small number of files.
* Large accidental deletes - potentially we could have a rogue script that does mass deletion of buckets and their data in it. That would have serious business consequences if lots of apps and models are suddenly not available for the while it would days or weeks to reproduce everything. The versioned copies would also be vulnerable - we'd probably want backups held elsewhere, like we do for logs.
* Non-reproducible data - there is data that is uploaded by users to their buckets. It makes sense to back up this.
* We're not too worried about S3 buckets losing data themselves - their durability is 11 nines or something.

This PR adds these features:

* a script to change existing buckets.
* a change in the control panel so that new buckets have this config too.

The new configuration is, for each bucket:

* Versioning is switched *ON*.
* A new life cycle definition is created so that non-current versions of files are set to glacier storage after 30 days.

I've also updated the unit tests to confirm the *code* works.

## How to review

**DO NOT TEST AGAINST ALPHA.**

1. Please check the changes relating both the versioning and life cycle correctly reflect the description above.
2. Try out the script on some TEST infrastructure to ensure it works correctly.
3. Confirm that the `dev` version of this branch creates new S3 buckets with the expected versioning and life cycle correctly configured.


controlpanel/api/aws.py
```diff
@@ -199,13 +199,39 @@ def delete_role(name):
 
 def create_bucket(bucket_name, is_data_warehouse=False):
     try:
-        bucket = boto3.resource('s3').create_bucket(
+        bucket = boto3.resource("s3").create_bucket(
```
davidread, line 202: :trollface: ;-)

ntoll: I was using emacs at the time... ;-)

ntoll:
Ack. I'll split this PR so code goes to the right repos. :+1: on re-using Python code.

ntoll:
Related PR for the single shot script: https://github.com/ministryofjustice/analytics-platform-ops/pull/320

ntoll:
@xoen hmm... this worked in `alpha` when I tested against a "test" bucket.

xoen:
> @xoen hmm... this worked in alpha when I tested against a "test" bucket.

How bizarre!

I've checked the Control Panel's role in AWS and the only `s3:*` permissions I see are these:

```JSON
                "s3:CreateBucket",
                "s3:PutBucketLogging",
                "s3:PutBucketPublicAccessBlock",
                "s3:PutEncryptionConfiguration"
```

Hang on, how did you test it? I guess it worked because you assumed restricted admin role which has more permissions? I'm not sure but it shouldn't in theory work unless the CP role has the right permissions (which I see you're in the progress of [adding here](https://github.com/ministryofjustice/analytics-platform-ops/pull/320/commits/0f78f46f5489f9cac3bce5860eb032ae8b27ef36) anyway ⭐ )

(I suppose that was one of the arguments for having a dev role with the CP IAM policy attached instead as this way the local test user would have the same permissions of CP, but it's a bit more fiddly than generating temporary creds to assume the restricted admin role.)

davidread:
As discussed, this code is reviewed and can be merged but we should note & discuss here the deploy plan. (Which includes deploying the IAM role from https://github.com/ministryofjustice/analytics-platform-ops/pull/320 )

ntoll:
Deploy Plan.

TL;DR - we "dry run" with the `dev` environment and then deploy to `alpha`.

Steps:

1. Confirmation that the IAM role user by the app has the correct/necessary permissions to update S3 buckets for versioning and life cycle.
2. Deploy this change and immediately check newly created buckets have the expected versioning flag and life cycle.
3. Monitor the situation.

I'd like to understand more about step 1 - especially how it relates to our use of Terraform (https://github.com/ministryofjustice/analytics-platform-ops#terraform). I'd like to be certain the steps needed to fulfil step 1 are understood and documented somewhere.

davidread:
Deploy plan looks good. Do grab someone to go through the terraform steps together.

ntoll:
Will do.

ntoll:
Terraform steps were carried out yesterday. Merging and cutting a release.

PR #817, opened 2020-06-01
Ensure admin tabs are highlighted in red.

ntoll:
## What

See this Trello ticket: https://trello.com/c/rz1BbD1f/562-investigate-labeling-the-superuser-tab-and-other-superuser-things-in-control-panel-so-that-when-doing-support-we-know-what-regul

This change ensures admin only tabs are highlighted in red, as per the screenie "mock-up" shown below:

admins

## How to review

Start the dev app as an admin. Ensure only the admin tabs are highlighted in red. Requires an Eyeball Mk.1 to accomplish. :eyes:

### DIFF
--- a/controlpanel/frontend/jinja2/base.html
+++ b/controlpanel/frontend/jinja2/base.html
@@ -82,7 +82,8 @@
         "hide": not request.user.is_superuser,
         "text": "Home",
         "href": home_url,
-        "active": page_name == "home"
+        "active": page_name == "home",
+        "admin": request.user.is_superuser
       },
       {
         "text": "Analytical tools",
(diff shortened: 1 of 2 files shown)


davidread:
I'm no designer, but I'm not a fan of the red. I'm not sure administration = "danger".

Maybe consider having a badge that says 'admin', with a hover text that explains that this is only visible to administrative roles, or something? Or an icon?

xoen:
@ntoll strangely enough the docker build fails with:

```
 
  python3-dev-3.7.7-r0:
 
    breaks: world[python3-dev=3.7.5-r1]
```

despite no `Dockerfile` change in the diff? Weird, I see this has changed in the [related PR](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/818/files) tho 🤷‍♂️

ntoll:
OK... red is obviously not the colour of the season. ;-) How about something like a spanner icon in the admin tabs..? :wrench:

ntoll:
Here's a not-red spanner thanks to HTML entity `&#128295;`

admin_spanner

PR #818, opened 2020-06-01
Show parameters tab.

ntoll:
## What

**THIS PR IS STACKED UPON #817**

This PR ensures the parameters tab in the main menu is displayed for all (non-admin) users.

Fixes this Trello ticket: https://trello.com/c/Hc7F4OX0/602-add-parameters-tab-for-users-who-arent-superusers

The end result looks like this:

parameters

## How to review

1. Start the dev server locally.
2. Log in as a regular user.
3. Can you see the parameters tab (and when clicking on it, does it take you to the expected location)..?

Requires an Eyeball Mk.1 to test. :eyes:

### DIFF
--- a/controlpanel/frontend/jinja2/base.html
+++ b/controlpanel/frontend/jinja2/base.html
@@ -83,29 +83,29 @@
         "text": "Home",
         "href": home_url,
         "active": page_name == "home",
-        "admin": request.user.is_superuser
+        "admin": request.user.is_superuser,
       },
       {
         "text": "Analytical tools",
         "href": url("list-tools"),
-        "active": page_name == "tools"
+        "active": page_name == "tools",
       },
       {
         "text": "Warehouse data",
         "href": url("list-warehouse-datasources"),
-        "active": page_name == "warehouse-datasource-list"
+        "active": page_name == "warehouse-datasource-list",
       },
       {
         "hide": not request.user.users3buckets.filter(s3bucket__is_data_warehouse=False).exists(),
         "text": "Webapp data",
         "href": url("list-webapp-datasources"),
-        "active": page_name == "webapp-datasource-list"
+        "active": page_name == "webapp-datasource-list",
       },
       {
         "hide": not request.user.userapps.filter(is_admin=True).exists(),
         "text": "Webapps",
         "href": url("list-apps"),
-        "active": page_name == "webapps"
+        "active": page_name == "webapps",
       },
       {
         "text": "Parameters",
@@ -117,9 +117,9 @@
         "text": "Groups",
         "href": url("list-policies"),
         "active": page_name == "groups",
-        "admin": request.user.is_superuser
-      }
-    ]
+        "admin": request.user.is_superuser,
+      },
+    ],
   }) }}
   {% endif %}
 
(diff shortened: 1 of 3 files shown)


controlpanel/frontend/jinja2/base.html
```diff
@@ -107,16 +108,16 @@
         "active": page_name == "webapps"
       },
       {
-        "hide": not request.user.is_superuser,
```
xoen, line 110: FYI: I'll focus my review on this line as the rest seem to be carried from a different/related branch.

I'm keen on having the permissions' source of true in few key places in the codebase, one of them currently being [the `api/rules.py` file](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/controlpanel/api/rules.py#L196-L200).

This has some benefits, including predictability (assuming someone knows permissions are defined there) and also the fact that reduces duplication and if for whatever reason we make the current logic to show the parameters more complicated we only need to change it there.

The `api.list_parameter` is already defined so that would mean use it in this template:

```jinja
"hide": not request.user.has_perm("api.list_parameters"),
```

(👆 or something like that, not tested)

Mind you, the net result is the same and you may argue it's annoying to have to jump in the `api/rules.py` file to check who can list parameters but it reduces the chance of future bugs or different part of the CP showing different things in the event of changing this simple logic to list parameters.

My 2 cents anyway ☺️

ntoll: Aha... I'll revise. I completely agree WRT single source of truth for permissions and I'll re-use the thing you pointed out. :+1:

controlpanel/frontend/jinja2/base.html
```diff
@@ -107,16 +108,16 @@
...
         "text": "Parameters",
         "href": url("list-parameters"),
-        "active": page_name == "parameters"
+        "active": page_name == "parameters",
```
davidread, line 113: maybe add the trailing comma to L120 instead?

controlpanel/frontend/jinja2/base.html
ntoll: OK... it was interesting to learn about the rules and how they're applied in templates. But, AFAICT, we want *all* users to be able to set parameters. As a result, we don't need any `hide` flag on the tab.

controlpanel/frontend/jinja2/base.html
ntoll: :slightly_smiling_face: so this isn't consistent with the rest of the dicts defined in the file. I personally prefer to add commas, but whoever did this work was obviously of the "other school". ;-)

xoen: @ntoll or they may have just forgot 😬 - in the future feel free to add trailing commas in part of the code you're touching - unless of course is syntactically wrong (I'm looking at you JSON! 😬)

I don't think anyone would tell you off for being inconsistent for adding a trailing comma 😉

ntoll: OK... with the latest commit I've taken the executive decision to add training commas to all the appropriate things in the `base.html` file.

For discussion and perhaps a further PR, what do folks think about using [Black](https://black.readthedocs.io/en/stable/) to automagically format our code for us so we don't need any arbiter of "code style"? As the Black docs say, "you can have any colour so long as..." FWIW after an initial "wat?" reaction I've had fantastic experience of using Black and would recommend it without hesitation.

PR #821, opened 2020-06-24
Turn plain <br> that's escaped by template renderer to a full stop.

ntoll:
## What

See ticket: https://trello.com/c/ThgiskvR/641-stray-br-showing-on-control-panel-app-registration-form

It was possible for a stray "escaped" fragment of HTML `<br>` to appear in certain forms in the control panel. Django/Jinja2 are doing the correct behaviour and we shouldn't put plain HTML into random strings expecting them to be rendered properly. I've replaced `<br>` with `. `  to delineate between error messages.

## How to review

1. Try to create an app in the erroneous way as described in the Trello ticket.
2. Observe the magnificent absence of `<br>`.

### DIFF
--- a/controlpanel/frontend/jinja2/datasource-create.html
+++ b/controlpanel/frontend/jinja2/datasource-create.html
@@ -39,7 +39,7 @@
         "maxlength": "60",
       },
       "value": form.name.value(),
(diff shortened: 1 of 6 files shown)


ntoll:
Since this is a trivial cosmetic change it can wait for the next release (rather than cutting one for this PR).

PR #822, opened 2020-06-26
Show tool version on tool Deploy/Upgrade

xoen:
### What

When a user click on the Deploy or Upgrade button the Control Panel
now shows them the version of the tool they're about to install or
upgrade to.

This version is actually the `appVersion` information coming from
the helm chart, this should contain useful information, for example
for RStudio would be something like

> RStudio: 1.2.1335+conda, R: 3.5.1, Python: 3.7.1, patch: 10

This will give the user enough details to take an informed decision
on whether to install or more importantly upgrade.

Example of prompt:

Screen Shot 2020-06-26 at 12 18 51

### Some technical notes
As part of this change I've moved the logic to get a tool/chart
`appVersion` into the `HelmRepository` class as it's now used in more
than one place and it makes sense for this to be in the class used to
query the helm repository.

### Ticket
https://trello.com/c/jHI3iCHq

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -289,7 +289,7 @@ class ToolDeployment():
             return helm.upgrade_release(
                 self.release_name,
                 f'{settings.HELM_REPO}/{self.chart_name}',  # XXX assumes repo name
-                # f'--version', tool.version,
+                f'--version', self.tool.version,
                 f'--namespace', self.k8s_namespace,
                 *set_values,
             )
@@ -310,7 +310,6 @@ class ToolDeployment():
             self.k8s_namespace,
             label_selector=(
                 f"app={self.chart_name}"
-                # f'-{tool_deployment.tool.version}'
             ),
         )
 
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -15,14 +15,14 @@ from controlpanel.utils import github_repository_name
 log = logging.getLogger(__name__)
 
 
-TOOL_DEPLOYING = 'Deploying'
-TOOL_DEPLOY_FAILED = 'Failed'
-TOOL_IDLED = 'Idled'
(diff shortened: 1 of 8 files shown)


xoen:
hi @ntoll I've tested this locally and it seems to work well, I hope to do some little more manual testing in `dev` on Monday before releasing it (unless of course there are surprises), it would be great if you could manage to review this before Monday ☺️ (if possible of course, otherwise it'll have to wait)

BTW: As part of this I've ran the `black` code formatter on the Python files touched by this, I've kept this in separate commits so it may help you review this by seeing the two commits with the "real" changes individually (without the formatting noise, my 2cents of course ☺️ ):
- https://github.com/ministryofjustice/analytics-platform-control-panel/pull/822/commits/a3d5f08dfba93298fdbef017b566587bb562dc7e
- https://github.com/ministryofjustice/analytics-platform-control-panel/pull/822/commits/5327586bb9f3630853d214a6f32c5ea6f901f5ef

ntoll:
@xoen hahahaha... my heart dropped when I looked at the diff and realised you'd run `black` to reformat the code. :laughing: Thanks for pointing out the "real" changes. :+1:

ntoll:
OK... found it. It appears to be an existing bug/naive practice (as per the comment in the function): https://github.com/ministryofjustice/analytics-platform-control-panel/blob/master/controlpanel/api/models/tool.py#L124

Since this is an *existing* "bug" (the version of the chart is checked, not the version of the actual tool) and not anything to do with this feature, I'll mark as approved rather than (as originally specified) requesting changes. I'll add a ticket to our triage column in Trello.

ntoll:
Here's the Trello card: https://trello.com/c/AHKgn0b5/656-correctly-ascertain-app-version-in-control-panel

ntoll:
Argh... sorry, clicked wrong button when submitting the "Trello" comment. Re-opening this PR. :facepalm:

PR #823, opened 2020-07-06
Allow users to self-reset their home directory.

ntoll:
## What

Sometimes a user's home directory will get itself into a bad state. This is often due to manual configuration relating to conda and r-studio. These changes automate the commands [found here](https://github.com/ministryofjustice/analytics-platform/wiki/Reset-your-R-Studio-environment) in the control panel app. They relate to [this issue in Trello](https://trello.com/c/CxKqlUpN/654-ripley-button-resets-a-users-conda-environment-etc).

This PR introduces the control panel changes needed to update the UI so this functionality is available to end users, and also adds a way to kick off a background job to run a helm chart to do the actual changes in the user's home directory.

## How to review

1. Currently a draft PR. Will fill these out once finished.

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -77,10 +77,9 @@ class User:
         """
         Reset the user's home directory.
         """
-        helm.reset_home(
+        helm.upgrade_release(
             f"reset-user-home-{self.user.slug}",
             f"{settings.HELM_REPO}/reset-user-home",
-            "--version=0.4.1",
             f"--namespace=user-{self.user.slug}",
             f"--set=Username={self.user.slug}",
         )
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -78,9 +78,10 @@ class User:
         Reset the user's home directory.
         """
         helm.reset_home(
-            f"0.1.0",
+            f"reset-user-home-{self.user.slug}",
             f"{settings.HELM_REPO}/reset-user-home",
-            f"--namespace={self.k8s_namespace}",
+            "--version=0.4.1",
+            f"--namespace=user-{self.user.slug}",
             f"--set=Username={self.user.slug}",
         )
 
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -73,11 +73,16 @@ class User:
             f"--set=Username={self.user.slug}",
         )
 
-    def reset(self):
+    def reset_home(self):
         """
         Reset the user's home directory.
         """
-        helm.reset_home(f"init-user-{self.user.slug}")
+        helm.reset_home(
+            f"reset-user-{self.user.slug}",
+            f"{settings.HELM_REPO}/reset-user",
+            f"--namespace={self.k8s_namespace}",
+            f"--set=Username={self.user.slug}",
+        )
 
     def delete(self):
         aws.delete_role(self.user.iam_role_name)
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -23,6 +23,16 @@ TOOL_READY = "Ready"
 TOOL_RESTARTING = "Restarting"
 TOOL_UPGRADED = "Upgraded"
 TOOL_STATUS_UNKNOWN = "Unknown"
+HOME_RESETTING = "Resetting"
+HOME_RESET_FAILED = "Failed"
+HOME_RESET = "Reset"
+
+
+class HomeDirectoryResetError(Exception):
+    """
+    Raised if a home directory cannot be reset.
+    """
+    pass
 
 
 class User:
@@ -62,6 +72,12 @@ class User:
             f"--namespace={self.k8s_namespace}",
             f"--set=Username={self.user.slug}",
         )
+
+    def reset(self):
+        """
+        Reset the user's home directory.
+        """
+        helm.reset_home(f"init-user-{self.user.slug}")
 
     def delete(self):
         aws.delete_role(self.user.iam_role_name)
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -78,8 +78,8 @@ class User:
         Reset the user's home directory.
         """
         helm.reset_home(
-            f"reset-user-{self.user.slug}",
-            f"{settings.HELM_REPO}/reset-user",
+            f"0.1.0",
+            f"{settings.HELM_REPO}/reset-user-home",
             f"--namespace={self.k8s_namespace}",
             f"--set=Username={self.user.slug}",
         )
--- a/controlpanel/api/helm.py
+++ b/controlpanel/api/helm.py
@@ -78,7 +78,13 @@ class Helm(object):
     def reset_home(self, release, chart, *args):
         HelmRepository.update()
         return self.__class__.execute(
-            "upgrade", "--install", release, chart, *args,
+            "upgrade",
+            "prometheus-resources",
+            chart,
+            "--install",
+            "--version",
+            release,
+            *args,
         )
 
     def delete(self, purge=True, *args):
--- a/controlpanel/api/helm.py
+++ b/controlpanel/api/helm.py
@@ -79,11 +79,8 @@ class Helm(object):
         HelmRepository.update()
         return self.__class__.execute(
             "upgrade",
-            "prometheus-resources",
-            chart,
-            "--install",
-            "--version",
             release,
+            chart,
             *args,
         )
 
--- a/controlpanel/api/helm.py
+++ b/controlpanel/api/helm.py
@@ -72,13 +72,10 @@ class Helm(object):
         HelmRepository.update()
 
         return self.__class__.execute(
-            "upgrade", "--install", "--wait", "--force", release, chart, *args,
-        )
-
-    def reset_home(self, release, chart, *args):
-        HelmRepository.update()
-        return self.__class__.execute(
             "upgrade",
+            "--install",
+            "--wait",
+            "--force",
             release,
             chart,
             *args,
--- a/controlpanel/api/helm.py
+++ b/controlpanel/api/helm.py
@@ -75,6 +75,13 @@ class Helm(object):
             "upgrade", "--install", "--wait", "--force", release, chart, *args,
         )
 
+    def reset_home(self, user):
+        HelmRepository.update()
+
+        return self.__class__.execute(
+            "reset", "--install", "--wait", "--force", user
+        )
+
     def delete(self, purge=True, *args):
         default_args = []
         if purge:
--- a/controlpanel/api/helm.py
+++ b/controlpanel/api/helm.py
@@ -75,11 +75,10 @@ class Helm(object):
             "upgrade", "--install", "--wait", "--force", release, chart, *args,
         )
 
-    def reset_home(self, user):
+    def reset_home(self, release, chart, *args):
         HelmRepository.update()
-
         return self.__class__.execute(
-            "reset", "--install", "--wait", "--force", user
+            "upgrade", "--install", "--wait", "--force", release, chart, *args,
         )
 
     def delete(self, purge=True, *args):
--- a/controlpanel/api/helm.py
+++ b/controlpanel/api/helm.py
@@ -78,7 +78,7 @@ class Helm(object):
     def reset_home(self, release, chart, *args):
         HelmRepository.update()
         return self.__class__.execute(
-            "upgrade", "--install", "--wait", "--force", release, chart, *args,
+            "upgrade", "--install", release, chart, *args,
         )
 
     def delete(self, purge=True, *args):
--- a/controlpanel/api/models/tool.py
+++ b/controlpanel/api/models/tool.py
@@ -206,30 +206,10 @@ class ToolDeployment:
         cluster.ToolDeployment(self.user, self.tool).restart(id_token)
 
 
-class HomeDirectoryManager:
-    """
-    Emulates a Django model manager.
-    """
-
-    def create(self, *args, **kwargs):
-        home_directory = HomeDirectory(*args, **kwargs)
-        home_directory.save()
-        return home_directory 
-
-    def filter(self, **kwargs):
-        return []
-
-
 class HomeDirectory:
     """
     Represents a user's home directory in the cluster
     """
-
-    DoesNotExist = django.core.exceptions.ObjectDoesNotExist
-    Error = cluster.HomeDirectoryResetError
-    MultipleObjectsReturned = django.core.exceptions.MultipleObjectsReturned
-
-    objects = HomeDirectoryManager()
 
     def __init__(self, user):
         self._subprocess = None
@@ -242,7 +222,7 @@ class HomeDirectory:
         """
         Update the user's home directory (asynchronous).
         """
-        self._subprocess = cluster.User(self.user).reset()
+        self._subprocess = cluster.User(self.user).reset_home()
 
     def get_status(self):
         """
--- a/controlpanel/api/models/tool.py
+++ b/controlpanel/api/models/tool.py
@@ -204,3 +204,68 @@ class ToolDeployment:
         """
 
         cluster.ToolDeployment(self.user, self.tool).restart(id_token)
+
+
+class HomeDirectoryManager:
+    """
+    Emulates a Django model manager.
+    """
+
+    def create(self, *args, **kwargs):
+        home_directory = HomeDirectory(*args, **kwargs)
+        home_directory.save()
+        return home_directory 
+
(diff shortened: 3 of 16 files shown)


controlpanel/api/helm.py
```diff
@@ -75,6 +75,13 @@ def upgrade_release(self, release, chart, *args):
             "upgrade", "--install", "--wait", "--force", release, chart, *args,
         )
 
+    def reset_home(self, user):
+        HelmRepository.update()
+
+        return self.__class__.execute(
+            "reset", "--install", "--wait", "--force", user
```
xoen, line 82: I'm not sure I follow this.

AFAICT helm doesn't have a `reset` command, does it? (Checking...) https://helm.sh/docs/helm/

Maybe you meant install an helm chart called `reset` into the user home?

But this seem to try and run `helm reset --install --wait --force USER` which I'm pretty sure would raise an exception as helm doesn't recognise it.

It may be worth trying it.

May I suggest you re-use the `Helm.upgrade_release()` function instead? I just started to have a look at this so I may be wrong but AFAICT you're trying to install a helm release into the user home which contains a `Job` of some sort.

Also, it may be worth specifying the user namespace as well so that this releases would be inside the user namespace with the other user-specific things.

controlpanel/api/cluster.py
```diff
@@ -63,6 +73,12 @@ def create(self):
             f"--set=Username={self.user.slug}",
         )
 
+    def reset(self):
```
xoen, line 76: I'll probably leave this comment multiple times but bear with me ☺️  (I'll try not tho 😬 )

Call these this operation just "reset" is a bit ambiguous, I would be more explicit and call it `reset_home`/`ResetHome`/`Reset home` everywhere to avoid any possible confusion.

(For example "old" desktop computers used to have a "reset" button which was restarting the machine so it may lead to someone thinking this has something to do with restarting something)

I know the context in this method is a "User" but I would err on the side of consistency and being explicit everywhere ☺️

controlpanel/api/helm.py
ntoll: This is a placeholder for whatever the real command for helm will be. ;-) (I should have commented this was the case, but it is, after all, a work in progress).

controlpanel/api/cluster.py
```diff
@@ -63,6 +73,12 @@ def create(self):
...
+        """
+        Reset the user's home directory.
+        """
+        helm.reset_home(f"init-user-{self.user.slug}")
```
xoen, line 80: Why this is passing the `init-user-USER` helm release name?

We're not planning to re-use [the `init-user` helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/tree/master/charts/init-user) right?

ntoll: PLaceholder (again)

controlpanel/api/cluster.py
```diff
@@ -23,6 +23,16 @@
 TOOL_RESTARTING = "Restarting"
 TOOL_UPGRADED = "Upgraded"
 TOOL_STATUS_UNKNOWN = "Unknown"
+HOME_RESETTING = "Resetting"
+HOME_RESET_FAILED = "Failed"
+HOME_RESET = "Reset"
```
xoen, line 28: As I said in other comment I'd be explicit and call this "Reset Home" etc everywhere to avoid any possible ambiguity with other related or not operations.

controlpanel/api/cluster.py
ntoll: Aldo, it literally says, "Resets the user's home directory" in the comment directly below this line. :rofl: I'm trying to work out how more explicit I can be. For the record, I really dislike `def long_java_like_methods_that_really_should_be_comments(self):`. I guess this is a question of taste. :wink:

controlpanel/api/models/tool.py
```diff
@@ -204,3 +204,68 @@ def restart(self, id_token):
...
+    Emulates a Django model manager.
+    """
+
+    def create(self, *args, **kwargs):
+        home_directory = HomeDirectory(*args, **kwargs)
+        home_directory.save()
+        return home_directory 
+
+    def filter(self, **kwargs):
+        return []
```
xoen, line 220: I got rid of the `ToolDeploymentManager` as it was more confusing than anything.

I get that it's an abstraction but I found that make it look like it's a persisted model was more confusing than anything.

In this case is less confusing as we don't have any related models but I would still avoid this abstraction.

ntoll:
Thanks for the feedback @xoen - I can see `reset_home` might be a useful change to make the code more readable. :boom:  :house:

controlpanel/api/models/tool.py
```diff
@@ -204,3 +204,68 @@ def restart(self, id_token):
...
+
+
+class HomeDirectory:
```
xoen, line 223: I don't think this should be among the models as it's more of nice way to interact with the cluster (and possibly not in the `api.models.tool` but this is a WIP PR so I know any of this is subject to change ☺️ )

I'd move this class in the `cluster` module where the other classes we use to interact with 3rd party systems are.

The views can still use this, don't think we need to have the model when in this case there is nothing really persisted in the DB (as the state is in k8s cluster).

controlpanel/api/models/tool.py
ntoll: Yeah... I did wonder about this. Just trying to "follow precedence". The manager object on a model instance is kinda expected in Django's ORM which is why I thought we had the `ToolDeployManager` (I couldn't see any other reason for having it). If there isn't I'll remove the equivalent from this PR. :+1:

Thanks for clearing that up - I did wonder and it was on my list of questions to ask when I submitted this out of draft for proper review.

controlpanel/api/models/tool.py
ntoll: Again, just following the coding conventions found in the code base. Again, I did wonder, and agree cluster is probably a better place to put it.

controlpanel/api/models/tool.py
```diff
@@ -204,3 +204,68 @@ def restart(self, id_token):
...
+        """
+        Poll the deployment subprocess for status
+        """
+
+        if self._subprocess.poll() is None:
+            return cluster.HOME_RESETTING
+
+        if self._subprocess.returncode:
+            log.error(self._subprocess.stderr.read().strip())
+            return cluster.HOME_RESET_FAILED
```
xoen, line 269: these `cluster.STATUS` references may be another clue that this class may belong in the `cluster` module ☺️

ntoll: Again, just following the conventions found in the tool deploy related classes in the `tool` module. It does seem a bit odd that any of this stuff should be in this part of the code base. A sign that we should refactor perhaps..? :+1:

controlpanel/frontend/urls.py
```diff
@@ -53,5 +53,6 @@
     path("webapps/<int:pk>/datasource-access/", views.GrantAppAccess.as_view(), name="grant-app-access"),
     path("webapp-datasource-access/<int:pk>/", views.UpdateAppAccess.as_view(), name="update-app-access"),
     path("webapp-datasource-access/<int:pk>/delete/", views.RevokeAppAccess.as_view(), name="revoke-app-access"),
+    path("reset/", views.Reset.as_view(), name="home-reset"),
```
xoen, line 56: again (sorry!) I'd be more explicit in the URL, something like `/user-home-reset/`.

controlpanel/api/cluster.py
xoen: Sure if you read the documentation but the thing is in the client code you'll see `ntoll.reset()` but what does that mean? To reset your MFA? To reset your user and its resources to the very initial state where you didn't even have a namespace?

From the Zen of Python:
> Explicit is better than implicit.

So I'd go with `reset_home()` 🙏 . It's not quite `long_java_like_methods_that_really_should_be_comments` ;)

controlpanel/api/models/tool.py
xoen: No problem, I don't think it wasn't working or anything but from experience with the other `ToolDeploymentManager` class it was more confusing than anything (to me at least) as it was adding an additional layer of abstraction and tricked me to think it was something to do with a database record.

Have a play and see what feels simpler ☺️

controlpanel/api/models/tool.py
xoen: I'm not sure, the whole checking for the helm release process to finish is a bit low-level. I don't have an answer to that but I haven't touched it as I'm moving few other things around and I tend to avoid changing many things at the same time.

But it was something I thought we could have a look after this Tool Catalogue bit as there may be some little bugs in the tool deployment status logic etc...

ntoll: OK... I've left as is (see incoming commit).

controlpanel/api/cluster.py
```diff
@@ -63,6 +73,18 @@ def create(self):
...
         )
 
+    def reset_home(self):
+        """
+        Reset the user's home directory.
+        """
+        helm.reset_home(
+            f"reset-user-home-{self.user.slug}",
+            f"{settings.HELM_REPO}/reset-user-home",
+            "--version=0.4.1",
```
xoen, line 83: It may be worth adding this in a setting or even getting rid of it and always install the latest version of this `reset-user-home` helm chart. 

The latter is potentially the best option as I imagine we're interested in the latest/greatest way of resetting the user home rather than to pin it to a specific version. Plus this way we don't have to get back and update/re-release the CP if that helm chart gets updated.

IMHO of course.

ntoll: :+1: What's the flag for always using the latest version of the chart..?

controlpanel/api/helm.py
```diff
@@ -75,6 +75,15 @@ def upgrade_release(self, release, chart, *args):
...
         )
 
+    def reset_home(self, release, chart, *args):
+        HelmRepository.update()
+        return self.__class__.execute(
+            "upgrade",
+            release,
+            chart,
+            *args,
+        )
```
xoen, line 85: I'm not sure this method adds much to the existing `Helm.upgrade_release()` method which we use to install/upgrade an helm chart.

Any reason to not use that instead? It's doing the same thing AFAICT no?

Maybe I'm missing something.

controlpanel/api/cluster.py
ntoll: OK... no need to use the `--version` flag - if it's not there, it'll always try the latest version.

ntoll: Good catch..! :+1:

controlpanel/frontend/consumers.py
```diff
@@ -166,6 +168,23 @@ def get_tool_and_user(self, message):
...
+        
+        home_directory.reset()
+
+        status = wait_for_home_reset(home_directory)
+
+        if status == HOME_RESET_FAILED:
+            log.error(f"Failed to reset home directory for user {user}")
+        else:
+            log.debug(f"Reset home directory for user {user}")
+
```
xoen, line 187: Just a question: We send the status update but the UI doesn't use it right? I guess it's still good to send this updates even if the UI at the moment doesn't show any progress or errors (and it may be working and overkill to do something more clever I guess)

controlpanel/api/helm.py
ntoll: Originally they were slightly different. I'm not sure we need the `--wait` or `--force` flags for resetting the home. Putting aside these queries, if they're doing the same thing, I agree we should just re-use existing code.

controlpanel/frontend/jinja2/reset.html
```diff
@@ -0,0 +1,27 @@
...
+{% block content %}
+  <div class="markdown-container" id="reset-container">
+    <h1>Home Directory Reset</h1>
+    <p>Even the best curated home directories can get into a bad
+    state (for example, a broken conda environment or Jupyter and RStudio won't
+    start).</p>
+
+    <p>Reset your home directory to recreate your conda environment and shell
+    settings to a clean state. Technical details of this process can be
+    <a href="">found here</a>.</p>
```
xoen, line 15: should this anchor have an empty href? I've just started to have a look at this PR so apologies if it's set dynamically somehow 🤷‍♂️

controlpanel/frontend/consumers.py
ntoll: Correct. It's there but not used. I was copying the code pattern I found already for background tasks and wanted to make it "complete". ;-)

controlpanel/frontend/jinja2/reset.html
ntoll: Nope... it's not yet written. ;-)

ntoll: Actually, I know where I could point it.

controlpanel/api/cluster.py
xoen: Yep I was about to reply ☺️

controlpanel/api/helm.py
xoen: Yeah I think it's conceptually the same thing, the `--force` seem to help with helm idiosyncrasies and the `--wait` is mainly there because our Tools can take long to deploy but in this case it shouldn't harm. So I'd personally avoid adding another method but of course it's your call.

controlpanel/frontend/views/reset.py
```diff
@@ -0,0 +1,37 @@
...
+from django.views.generic.edit import FormView
+from controlpanel.frontend.forms import ResetHomeDirectoryForm
+from controlpanel.frontend.consumers import start_background_task
+
+
+class ResetHome(LoginRequiredMixin, FormView):
+    """
+    A view to help automate the resetting of the user's home directory. To be
+    used when, for example, their conda or r-studio deployment gets into a bad
+    state. See the related helm chart (TBD) for the details of what actually
```
xoen, line 14: Maybe worth adding a link in this comment to the chart, now that it's in the helm charts repo https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/main/charts/reset-user-home/

I personally find these clues/crumbs useful in the code.

controlpanel/api/cluster.py
ntoll: Resolved in latest commit.

controlpanel/api/helm.py
ntoll: Resolved in latest commit.

controlpanel/frontend/jinja2/reset.html
ntoll: Resolved in latest commit.

controlpanel/frontend/views/reset.py
ntoll: Resolved in latest commit.

controlpanel/frontend/views/reset.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def get_success_url(self):
+        """
+        The form is valid, so kick off the background job to run the helm
+        chart and signal to the user the reset is underway.
+        """
+        start_background_task(
+            "home.reset",
+            {
+                "user_id": self.request.user.id,
```
xoen, line 29: Random info follows:

FYI: In this case the background job doesn't seem to do anything with the k8s API so we technically don't need to pass the ID Token to the worker.

But so that you know, in general if the background consumer needs to interact with the k8s API using the user credentials it would need the ID Token and it would not be able to retrieve it as it's in a separate process from the frontend.

e.g. they can't call something like `self.request.user` as this is in a different context.

Random info but I thought it may be useful for you to know for the future 😬

controlpanel/api/cluster.py
```diff
@@ -63,6 +73,17 @@ def create(self):
...
         )
 
+    def reset_home(self):
+        """
+        Reset the user's home directory.
+        """
+        helm.upgrade_release(
+            f"reset-user-home-{self.user.slug}",
+            f"{settings.HELM_REPO}/reset-user-home",
+            f"--namespace=user-{self.user.slug}",
```
xoen, line 83: Not that it makes any difference as it yields the same result but as you can see from `Helm.upgrade_release()` call above you may want to use the `self.k8s_namespace` which returns the user namespace.

As I said, it's the same string at the end of the day but I thought you may want to know that there is a method/property in the user model for that.

controlpanel/api/models/tool.py
```diff
@@ -204,3 +204,48 @@ def restart(self, id_token):
...
+    """
+    Represents a user's home directory in the cluster
+    """
+
+    def __init__(self, user):
+        self._subprocess = None
+        self.user = user
+
+    def __repr__(self):
+        return f"<HomeDirectoryManager: {self.user!r}>"
```
xoen, line 219: Is the trailing `Manager` intentional? I guess it doesn't make much difference but it may confuse one of us in the future is the class name is slightly off 🤷‍♂️

PR #824, opened 2020-07-10
User can choose version of a tool from a potential list of supported versions

xoen:
### What

We want to potentially support more than one version of a tool to allow users to chose between very few blessed versions with potentially different environment (e.g. RStudio with R v1 vs RStudio with R v2).

### UI Changes

UI had to change quite significantly to allow users to select a version:

Screen Shot 2020-07-10 at 12 21 41

Notes:
- "Upgrade" button is gone, user now select a version and "Deploy" it
- "Deploy" button is disabled if user didn't select any version or selected the version currently deployed
- The version of the helm chart is also shown in square parenthesis to avoid confusion when a user is upgrading to a version of a tool which only changes "internally" (e.g. some of the kubernetes resources changed or maybe the auth-proxy version was bumped but the rest is the same)
- Buttons are no longer hidden/shown but they're instead disabled/enabled

### Other technical implementation details
The current/old version of Control Panel rely on the fact that there is only one `Tool` instance for each of the tools.
So for example you may have `[{RStudio version="v1"}, {JupyterLab version="v1"}]` but **not** `[{RStudio version="v1"}, {RStudio version="v2"}, {JupyterLab version="v1"}]` because this would confuse everything from queries expecting to find only one record with a given `chart_name` to the frontend template which looped through the tools and it would show duplicates etc...

This mean that I had to move quite few things behind the scenes to allow more than a `Tool` record for the same chart (but with different versions of course).

The list of versions is still a whitelist coming from the `Tools` table and probably at the beginning we'll not have multiple versions anyway (maintain/support multiple versions of a tool still requires effort).

Other notable changes:
- Code related to the "Upgrade" button was removed, including `Tool.outdated()` method which was used to determine whether to show the button
- Removed apparently unused (apart from the tests) `ToolDeployment` view and related code. This is nice as I was able to remove the `ToolDeploymentManager.filter()` method which never felt right and other boilerplate code.
- Formatted some of the code touched with `black`

### How to review

1. Try it locally
2. ???
3. Profit!

### Ticket
https://trello.com/c/DBJhJKPT


damacus:
This is ace, and opens up a few possibilities for switching notebook versions which I'm likely to use in the next month or so 🥳 

## 🤔 

- What happens to the old version of the tool if to select a new one? (I'm assuming it gets deleted)
- As the version numbers can get really verbose, might it be worth having some sort of pop up with a nicely formatted list of the installed tools? e.g. 

- User clicks on RStudio version selection drop-down (like now)
- Use clicks on "details" or ℹ️
- Popup (or something more friendly) appears with a formatted list of the pre-installed tools:

```text
RStudio 1.2.335
Conda 4.5.11
Conda Python 3.6
Python 3.7.1
```

Being verbose about the installed tool version will help us diagnose any problems with installed tools at a later date

davidread:
This is great!

This might well be enough for an MVP that we can ship now. In general it's useful to deliver the value of work asap. And redesigns are improved if you have some user feedback. However I'll put forward my thoughts to add to the mix in your consideration.

As well as @damacus, I'm keen on putting this selecting action onto a pop-up or separate page. Here's my thinking:

* The primary action on this page is "Open RStudio" and I fear that this is getting more and more lost as we add detailed things like this
* [Radio buttons are preferred over select boxes in GDS design world](https://govuk-elements.herokuapp.com/form-elements/#form-select-boxes). It's an accessibility issue. And radios would also give us more space to nicely format and I think we'll probably want to add a bit of help text.

ntoll:
:wave: A summary of the feedback I gave Aldo via Slack.

* I believe getting some design/content help for the UI is very important.
* The buttons on the RHS are squashed.
* The  drop down is veeeeerrrrryyyyyyy looooonnnnnnnggggg and, in the case of R-Studio simply contains way too much uninteresting information IMHO (too much noise in the signal).
* It's not obvious to a user what exactly they're supposed to do on this page. Some friendly instructions might help.
* It feels odd that we're telling them which version is currently installed within the dropdown for selecting other versions to be installed. We're mixing up functions in a single widget.

Aldo and I will pair on this to see if we can think of ways to neaten things up a bit. I completely agree that getting this in front of users is very important.

ntoll:
Just to be clear, I see this as if we're baking a cake. I'm saying we could tidy up the icing on top. It's all still sweet an lovely.

ntoll:
Having discussed with @xoen, we think this MVP should be merged. It meets the initial criteria of a select box for the version of the app to use.

**However**, we feel this is very much a first step and much more UX and user journey thought needs to be put into the application as a whole. I imagine @davidread will have helpful input in this regard. 

Some observations that fell out of our conversation:

* We like the tool table list - it's nice and simple. The update in this PR makes the tool list a lot more complicated and difficult to understand.
* We think that each tool could get its own configuration page to hide complexity and give us enough screen real-estate for appropriate widget use such as radio buttons and additional configuration that may be needed in the future. (This also means the tool list is just that, a list for tools rather than having a configuration story to it).
* We'd like to use GDS design system and would appreciate help from folks with UX/design expertise in this.

We also believe that the work on the home directory reset and this PR subtly interacts with each other. We need to work out a way of consolidating these changes into a solution that works together.

xoen:
FYI: I've added a Trello ticket with these ideas for future improvements: https://trello.com/c/kKBaBL7f - if you have more ideas feel free to add there.

xoen:
@ntoll FYI: Removed the "None" when a tool wasn't installed: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/824/commits/bb69483742607e41b8cd46cd333cfb5fab14a142

ntoll:
Do you think we could very quickly find some design related input..? Would be great to simplify the UI before revealing it to the users. :wink:

davidread:
It's so great to be able to see this visually, and joined up with the helm stuff underneath. But the more I study it the more I worry about how users will struggle with it. I take the rap for originally suggesting a simple drop-down. But seeing that it is also used to display the current version, and needs a button to change the version, these details add up to a UX that feels jarring to me. We've also got the Ripley button coming onto this page too imminently. I'm feeling that we need a designer to guide this page more confidently than we can.

Of course our team's a bit unique letting engineers release MVPs without design input. In a normal service team a feature would have been designed by a designer using the design system, then would go to the team for a design critique workshop, before then going to the engineer to implement the MVP. I think we've got away without design input too long on this page. So I'd like to help quickly get a designer in.

xoen:
@ntoll was having a chat with @samtazzyman and he seemed very happy with my demo of this etc...I'm keen on putting this in front of the users for their feedback and then iterate rather than keeping this branch on the side forever until everything is perfect (which seem like the opposite of working in an agile way).

xoen:
> It's so great to be able to see this visually, and joined up with the helm stuff underneath. But the more I study it the more I worry about how users will struggle with it. I take the rap for originally suggesting a simple drop-down. But seeing that it is also used to display the current version, and needs a button to change the version, these details add up to a UX that feels jarring to me. 

I showed this to @samtazzyman - which arguably understand our users better than anyone else - and he seemed happy with it and didn't seem to suggest this is a UX disaster and he also understood that we can always iterate/improve (we're not building skyscrapers that stay the same for decades after all)

> We've also got the Ripley button coming onto this page too imminently. I'm feeling that we need a designer to guide this page more confidently than we can.

That's not really in this page a part from a brief mention with a link to that feature page.

> Of course our team's a bit unique letting engineers release MVPs without design input. In a normal service team a feature would have been designed by a designer using the design system, then would go to the team for a design critique workshop, before then going to the engineer to implement the MVP. I think we've got away without design input too long on this page. So I'd like to help quickly get a designer in.

And a frontend to implement the frontend part of features 🙏?

controlpanel/frontend/jinja2/tool-list.html
```diff
@@ -9,78 +9,110 @@ <h1 class="govuk-heading-xl">Your tools</h1>
...
-            {{ deployment.get_installed_app_version(id_token) or "Unknown" }}
-          {% endif %}
-        </small>
-      </td>
-      <td class="govuk-table__cell">
-        <div class="tool-status-label">
-          {{ deployment and deployment.get_status(id_token) | default("") }}
+
+{# Grid Container for the list of tools (1 column, n rows) #}
+<div class="tools-list" style="display: grid; grid-template-columns: 1fr; border-top: 1px solid #b1b4b6; margin-bottom: 30px;">
```
davidread, line 14: I think the inline CSS is something we should avoid on principle, for the sorts of reasons here https://stackoverflow.com/questions/2612483/whats-so-bad-about-in-line-css. I think this is implied by our frontend guidance when it mentions:

> fundamental principles like separation of style and content

https://ministryofjustice.github.io/technical-guidance/principles/frontend-development-principles/

samtazzyman:
I think the functionality in this is great - it's a really important and valuable step towards getting the software catalogue we've spoken about before.

I agree there are potentially snags with the UX, however, as @xoen himself has acknowledged. @davidread do you think it's going to be feasible to get some designer time (no doubt they will also want user researcher time)? If it'll be a long time before that can be set up I'd be keen to release as-is for the moment (with updated user guidance) to avoid it becoming an unused side feature.

I'm happy to talk this through with everyone some time so we can reach a consensus. I think we all agree Aldo's done a great job on the functionality; the question is how do we maximise the value of what he's built by making it easy for the user to interact with it.

ntoll:
:wave:

The UI/UX aspects of both this and the Ripley button PR could be improved immeasurably and made easier for developers with the input of a UX/UI expert. Suggesting such input is in no way meant as any criticism of the work done so far on either of the tickets. Rather, I believe it's good to acknowledge the UI/UX definitely has shortcomings (with no blame attached).

The shortcomings are a natural by-product of stuff, context and capabilities changing. But these shortcomings also make it harder in both the long and short term to reveal all the cool stuff the control panel can do.

We should fix it while we can rather than plough on with building an ever-more-complicated page and potentially creating technical debt in the UI layer..! :art: 

(And we should ask for help too - we have talented UI/UX colleagues who can help us.) :+1:

controlpanel/frontend/jinja2/tool-list.html
davidread: Fixed - that was quick!

ntoll:
This is superseded by #829 (i.e. the changes in here were landed via that other PR). Thanks for the awesome work Aldo..!

PR #832, opened 2020-08-24
Fix accessibility issue (badly formed anchor tag) in navigation panel.

ntoll:
## What

Some badly formed data was being passed into the navigation panel which caused an invisible item to appear in the tab order (discovered during accessibility checking). This PR fixes the problem.

## How to review

1. Visit the website with the fix applied.
2. Tap the TAB key to cycle through the available UI elements and check there's no "invisible" item between your username and the "Sign out" link.
3. That's it!

Related to [this Trello ticket](https://trello.com/c/sl8kBwa6/755-accessibility-checklist-in-control-panel-investigate-and-audit).

### DIFF
--- a/controlpanel/frontend/jinja2/base.html
+++ b/controlpanel/frontend/jinja2/base.html
@@ -7,6 +7,7 @@
 
 {% set service_name = "Analytical Platform Control Panel" %}
 {% set home_url = url("index") %}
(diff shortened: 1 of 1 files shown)


xoen (approved):
LGTM 👍

damacus (approved):

PR #836, opened 2020-09-04
Upgrade the Python dependencies

damacus:
## What

Upgrade the Python dependencies in preparation to moving to Helm 3

## How to review

1 . Run the Docker build (make build)
2. Make sure nothing explodes

Dependencies upgraded using `pur`


ntoll (approved):
LGTM :shipit:

ntoll:
@damacus can we fix the broken build on CircleCI..?

xoen:
I'm not clear how this is related to move to Helm 3 🤷‍♂️ 

I'm seeing quite few big version number changes which in theory may be breaking changes. It may be worth trying this in `dev`. 

Also, [isn't the `requirements.lock` the dependencies files actually used in the `Dockerfile`](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/be949a54086c8f06f09a590ded1888b82773b8f4/Dockerfile#L39-L42)? I think that's why the [Makefile has a target for it](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/be949a54086c8f06f09a590ded1888b82773b8f4/Makefile#L31), but you could also update manually...but as I said, in theory it should be fine but in practice worth checking considering the big version changes.

xoen:
FYI: as discussed with @damacus [I've updated the `requirements.lock` file](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/836/commits/28d3f9484fddcfb879e885003676d31260d970b4). This file is the one used when running the CP locally and [when building the docker image](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/Dockerfile#L41).

Tests pass locally and will very likely pass also in CircleCI (Dan updated this branch in line with `main` which solved the `docker build` error).

It would be good to have also your eyes on the diff @ntoll in case you spot any breaking changes here.

ntoll:
LGTM :+1: :shipit:

xoen:
FYI: Rebased since `main` moved a little since last commit...

xoen:
BTW forgot to say earlier 🤦 , thanks for your review @ntoll ☺️

PR #837, opened 2020-09-08
Add a svelte way to highlight and remove unused user accounts.

ntoll:
## What

This change makes it easy for admin users of the control panel to quickly identify users who have not logged in after approximately three months (90 days). Furthermore, links are provided to GitHub for the specific users' team membership settings for the MoJ organisation so they can be easily removed. This work needs to be done since the number of seats in the organisation is limited and we need a way to quickly curate such membership as admin users.

Setting aside the erroneous last-login time (done for testing purposes), the functionality looks like this:

unused_users

**Remember, the screenie was created with test data.** Obviously, the "unused" flag/link and related warning message won't appear unless the user hasn't logged in for 90 days.

## How to review

1. Make sure the `timedelta` used in the `three_months_ago` function in your local copy of this branch returns something not so far in the past.
2. Log in as an admin user.
3. Confirm you can see behaviour similar to that in the attached screenie.

The related ticket in Trello can be [found here](https://trello.com/c/uH1cuRhJ/65-check-users-are-still-active-and-automatically-suspend-inactive-accounts).

### DIFF
--- a/controlpanel/frontend/jinja2/user-detail.html
+++ b/controlpanel/frontend/jinja2/user-detail.html
@@ -24,8 +24,7 @@
       </h2>
 
       <p>
-        <strong>The user of this account has not logged in for over three
-        months.</strong>
+        <strong>The user of this account has not logged in for over 90 days.</strong>
       </p>
 
       <p>If appropriate, please <a href="https://github.com/orgs/moj-analytical-services/people/{{ user.username }}" target="_blank">visit the GitHub page for their membership of the
--- a/controlpanel/frontend/jinja2/user-detail.html
+++ b/controlpanel/frontend/jinja2/user-detail.html
@@ -13,6 +13,27 @@
 <span class="govuk-caption-xl">User</span>
 <h1 class="govuk-heading-xl">{{ page_title }}</h1>
 
+{% if unused %}
+<div class="govuk-grid-row">
+  <div class="govuk-column-two-thirds">
+
+    <div class="govuk-error-summary" role="alert" aria-labelledby="error-summary-heading-example-1" tabindex="-1">
+
+      <h2 class="govuk-heading-m govuk-error-summary-heading">
+      This Account Could be Unused
+      </h2>
+
+      <p>
+        <strong>The user of this account has not logged in for over three
+        months.</strong>
+      </p>
+
+      <p>If appropriate, please <a href="https://github.com/orgs/moj-analytical-services/people/{{ user.username }}" target="_blank">visit the GitHub page for their membership of the
+      Ministry of Justice organisation</a> to free up space.</p>
+    </div>
+</div>
+{% endif %}
+
 {% if request.user.has_perm('api.add_superuser') %}
 <section class="cpanel-section">
   <form action="{{ url('set-superadmin', kwargs={ "pk": user.auth0_id }) }}" method="post">
--- a/controlpanel/frontend/views/user.py
+++ b/controlpanel/frontend/views/user.py
@@ -13,13 +13,13 @@ from controlpanel.api import auth0
 from controlpanel.api.models import User
 
 
-def three_months_ago():
+def ninety_days_ago():
     """
     Returns a datetime object referencing approximately three months in the
     past, from the current date. The assumption made here is that a month is
     roughly 30 days (so the result is always 90 days in the past).
     """
-    return datetime.now() - timedelta(days=30*3)
+    return datetime.now() - timedelta(days=90)
 
 
 class UserList(LoginRequiredMixin, PermissionRequiredMixin, ListView):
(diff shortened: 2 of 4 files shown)


controlpanel/frontend/views/user.py
```diff
@@ -12,6 +13,15 @@
 from controlpanel.api.models import User
 
 
+def three_months_ago():
```
damacus, line 16:
```suggestion
def ninety_days_ago():
```

Maybe 90 days makes it clearer how long it actually is?

damacus:
Apart from the minor wording from 3 months --> 90 days. This looks ace!

controlpanel/frontend/jinja2/user-detail.html
```diff
@@ -13,6 +13,27 @@
...
+  <div class="govuk-column-two-thirds">
+
+    <div class="govuk-error-summary" role="alert" aria-labelledby="error-summary-heading-example-1" tabindex="-1">
+
+      <h2 class="govuk-heading-m govuk-error-summary-heading">
+      This Account Could be Unused
+      </h2>
+
+      <p>
+        <strong>The user of this account has not logged in for over three
```
xoen, line 27: Don't know whether we should be explicit and say "has not logged in the Control Panel for over three months" or not (to avoid any possible confusion)

I know it's a bit extreme but **potentially** a user could have not logged into the Control Panel for months but still actively working on things on GitHub.

Just a thought.

xoen: Actually I see the code in list view checks for last login in Auth0 and then fall back to `last_login` in CP/DB record.

controlpanel/frontend/views/user.py
```diff
@@ -36,11 +51,20 @@ def get_context_data(self, **kwargs):
                 continue
 
             try:
-                last_login = dateutil.parser.parse(user.get('last_login'))
+                auth0_last_login = user.get("last_login")
```
xoen, line 54: I have to say this code is (not the stuff you added but the whole function) surprisingly confusing.

For example, it wasn't immediately apparent to me that this `user` wasn't the user DB record as in `models.User` but in fact the Auth0 user info. Potential suggestion => rename `users` to `auth0_users` and `user` to `auth0_user`.

The `last_logins` was also a bit surprising because it seemed like a list but then it finally got converted to a dictionary. Potential suggestion => Maybe we should just declare as an empty dictionary `{}` from the beginning and use `last_logins[auth0_id] = {...}`.

Speaking of `last_logins`, it used contain just the last logins but now we add more login-related things to it so potentially we could rename it to `login_info` (or something like that).

ntoll: I agree. The code as it was originally was very confusing to work out just what was happening.

I went for the "make it work" rather than "refactor all the things" way with this PR since I'd be refactoring into a very deep rabbit hole. If you think it useful, I'm happy to simplify and refactor.

controlpanel/frontend/views/user.py
ntoll: :+1:

controlpanel/frontend/views/user.py
```diff
@@ -36,11 +51,20 @@ def get_context_data(self, **kwargs):
...
+                    db_user = auth0_mapper.get(auth0_id, None)
+                    if db_user:
+                        last_login = db_user.last_login
             except (TypeError, ValueError):
                 last_login = None
 
-            last_logins.append((auth0_id, last_login))
+            last_logins.append((auth0_id, {
+                "last_login": last_login,
+                "unused": last_login.utcnow() < three_months_ago(),
```
xoen, line 66: Maybe this is a bit objective but when I saw `unused` in the templates it wasn't immediately clear what this was (when I just started to look at this) so maybe call this variables `not_logged_recently` may help (or some better name than this).

xoen (approved):
Looks good to me 👍 

I've left some comments/ideas but nice first cut.

controlpanel/frontend/views/user.py
xoen: Up to you, I agree with you: I wouldn't go too deep into the refactoring hole but if you feel comfortable and think it's useful even rename few things in this function IMHO could help a lot the future us with readability ☺️

ntoll: :+1:

ntoll:
@damacus @xoen I've just push these changes:

* Three months -> 90 days. Good spot Dan.
* Messy code: I've annotated a bunch of information into the problem methods with TODO suggestions for the next person to work on this.

Once the tests turn green I'll merge and cut a release for deployment.

controlpanel/frontend/views/user.py
ntoll: See new comments in code. ;-)

controlpanel/frontend/views/user.py
ntoll: Done

damacus (approved):

controlpanel/frontend/jinja2/user-detail.html
ntoll: :+1:

controlpanel/frontend/views/user.py
ntoll: I really don't mind what we call it, ;-) but I chose `unused` because it's short and simple and is a good flag name to indicate the account maybe no longer used. The addition of the comments in the code should clarify what it's supposed to represent.

PR #838, opened 2020-09-10
Refactor/simplify last-login handling in admin user management.

ntoll:
## What

This PR refactors the way last-login is handled in the admin pages relating to users (the user list and user detail endpoints).

The original version used to get data from Auth0. However, this was problematic:

* It was complicated.
* It introduced a network call to the Auth0 API.
* It didn't work for all users.

After introducing a way to remove user who had not logged in for over 90 days it became clear this solution was problematic - to the extent that we'd inadvertently introduced a bug by adding to this complicated code.

As a result I've totally re-written the logic and simplified things:

* All logins from Auth0 are actually also stored in our local database as the `User` object's `last_login` field. Thus we don't need to ping Auth0's API.
* The Auth0API only returns information about that last 24 hours anyway - making it useless for our purposes for checking last login times.
* Much of the complexity in the original code was due to dealing with Auth0's responses.
* By using the built-into-Django `last_login` field we remove all Auth0 complexity/dependency.
* Some users in our database don't have a `last_login` value associated with them. This indicates they have *never* logged in (and are thus candidates for removal).

## How to review

1. Describe the steps required to test the changes
2. ???
3. Profit!

Related Trello ticket: https://trello.com/c/mHAoS6S2/784-fix-bug-in-last-login-handling-in-admin-screens

### DIFF
--- a/controlpanel/frontend/jinja2/user-list.html
+++ b/controlpanel/frontend/jinja2/user-list.html
@@ -29,20 +29,17 @@
       </td>
       <td class="govuk-table__cell">{{ user.email }}</td>
       <td class="govuk-table__cell">
-        {%- with login_details=last_login.get(user.auth0_id) -%}
-          {%- if login_details -%}
-          {%- with login = login_details.get("last_login") -%}
-          <span data-last-login="{{login}}"
-                title="{{login.strftime("%Y/%m/%d %H:%M:%S")}}">
-            {{ timesince(login) }} ago
-          </span>
-          {%- if login_details.get("unused") -%}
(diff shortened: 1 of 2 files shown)


damacus (approved):
I can't see anything obviously wrong with this. Nice spot with the local database 🥳

ntoll:
The problem is our data. The bug was introduced because some of our user data isn't "clean" in the way our test data is. I think we won't know it has worked until I've deployed it and tried some of the problem users (which I've already done "manually" via the database shell, so it *should* work).

In any case, this is a relatively low risk since it only affects admin related functionality (i.e. us) and not our users.

ntoll:
This PR contained a bug relating to datetime comparisons. Fixed by #839.

PR #840, opened 2020-09-10
Add "worker_health" command to check worker health

xoen:
This command will be used by Kubernetes to check on the Django Channels
worker using a "command" probe:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command

Why?
====
We noticed that sometimes the workers stop processing background tasks despite their containers are running fine.
This seems to be caused by some sort of connectivity problem related to Redis.

Regardless of the cause, [the `worker` containers of the Control Panel pods](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/main/charts/cpanel/templates/deployment.yaml#L54-L64) don't have a liveliness probe able to determine whether the worker is working (no pun intended) or not. Kubernetes should be able to use this command as a "command" probe to determine whether a worker container is still processing background tasks or not.

How does it work?
=================

The idea behind this is that the command will submit a "worker_health"
background job (asyncronous) and then check the existance/freshness of
the `/tmp/worker_health.txt` file.

As:
1. kubernetes will periodically execute this command and
2. the command will periodically submit this "heathbeat" `worker_health`
   background job

this means that the file should be almost always relatively fresh (
freshness may need tweaking using `--seconds` argument).

There are some subtleties to bear in mind:
1. there will be several instances of the workers (currently 3)
2. a background task submitted on a pod may be executed on a different pod
2a. This means that on average you can expect a given worker to process
    1/n of the tasks (and this needs to be backed into the `--seconds` and
    k8s probe configuration)
3. the worker probe command run on the same container where the worker
   health task will run, hence using a local, temporary file is
   satisfactory as a way for these to "communicate" (and it's very simple
   and understandable)
4. 2a and 3 above means that even if a health task is processed correctly
   every `--seconds` seconds the actual file touched may not be the same
   (again, this is OK but will need to be baked into the k8s probe
   configuration)
5. From local testing the worker health task is processed very quickly
   but because of the asyncronous nature of these background tasks and
   the existance of several workers on independent containers processing
   them it means that probe will need to be tweaked in a certain way
   so that few failures are acceptable

What the above mean is that the kubernetes liveliness probe will need to
be configured so that few failures are acceptable and it would not be
too aggressive into killing a worker container because of false positives.
(this is in general a best practice when it comes to liveliness probes,
configuring them is part science and part art).

More on Kubernetes Lifeliness probe
===================================
From testing of this probe on `dev` it seems like the "worker health"
jobs are processed very regularly.

Something like the following seem to work well:

```
livenessProbe:
   exec:
      command:
      - python3
      - manage.py
      - worker_health
      - "--stale-after-secs=120"
   initialDelaySeconds: 10
   periodSeconds: 30
   failureThreshold: 6
   successThreshold: 1
   timeoutSeconds: 30
```

- this will trigger/check the "health" task every 30s
- from observation of workers' logs in `dev` each worker process these
  health tasks every ~30s
- `--stale-after-secs=120` means that each individual probe will fail when
  the worker hasn't processed the health task in the last 2 minutes
- `failuresThreashold=6` means that individual probes failing will **NOT**
  trigger the restart of the individual worker container BUT 6 probes
  failing in a row will. This means that if a worker hasn't processed
  these tasks in 5 minutes it will be restarted (6x30s = 3m + 2m of the
  first probe failing because nothing run in 2m = 5m)
- `successThreshold=1` means that the moment a health task has been processed
  in the last 2m the probe succeed and the counter restart. E.g. if
  5 probes in a row fail but then the 6th succeeds a failure of the 7th
  probe will not trigger the restart of the container as a new cycle starts

If above is somehow not clean I advice to read more on Kubernetes probes
from their official documetation and possibly from other sources:

- https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command

Part of ticket: https://trello.com/c/eUMRw1gK

Also see PR with helm chart probe configuration: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/468

### DIFF
--- a/controlpanel/frontend/consumers.py
+++ b/controlpanel/frontend/consumers.py
@@ -2,6 +2,7 @@ import asyncio
 from datetime import datetime
 import json
 import logging
+from pathlib import Path
 from time import sleep
 import uuid
 
@@ -22,6 +23,9 @@ from controlpanel.api.cluster import (
 )
 from controlpanel.api.models import Tool, ToolDeployment, User, HomeDirectory
 from controlpanel.utils import PatchedAsyncHttpConsumer, sanitize_dns_label
+
+
+WORKER_HEALTH_FILENAME = "/tmp/worker_health.txt"
 
 
 channel_layer = get_channel_layer()
@@ -179,6 +183,11 @@ class BackgroundTaskConsumer(SyncConsumer):
             log.error(f"Failed to reset home directory for user {user}")
         else:
             log.debug(f"Reset home directory for user {user}")
+
+    def workers_health(self, message):
+        Path(WORKER_HEALTH_FILENAME).touch()
+
+        log.debug(f"Worker health ping task executed")
 
 
 def send_sse(user_id, event):
--- a/controlpanel/frontend/management/commands/worker_health.py
+++ b/controlpanel/frontend/management/commands/worker_health.py
@@ -1,0 +1,58 @@
+from datetime import datetime, timedelta
+from pathlib import Path
+from sys import exit
+
+from asgiref.sync import async_to_sync
+from channels.layers import get_channel_layer
+from django.core.management.base import BaseCommand, CommandError
+
+from controlpanel.frontend.consumers import WORKER_HEALTH_FILENAME
+
+
+class Command(BaseCommand):
+    help = "Checks if this worker is still running tasks"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--stale-after-secs",
+            type=int,
+            default=120,
+            help="For how many seconds the last executed task is considered recent enough for the health check to pass",
+        )
+
+    def handle(self, *args, **options):
+        stale_after_secs = options["stale_after_secs"]
+
+        # 1. Submit "workers health" background task
+        channel_layer = get_channel_layer()
+        async_to_sync(channel_layer.send)(
+            "background_tasks",
+            {"type": "workers.health"},
+        )
+
+        # Attempt to read worker health ping file
+        # NOTE: This may initially fail depending on timing of health task
+        # execution but that's fine as Kubernetes' `failureThreashold`
+        # will be more than 1 anyway
+        try:
+            last_run_at_epoch = Path(WORKER_HEALTH_FILENAME).stat().st_mtime
+            last_run_at = datetime.utcfromtimestamp(last_run_at_epoch)
+        except FileNotFoundError:
+            # Health ping file not found. Health task hasn't run on this worker yet
+            self.stderr.write(self.style.ERROR("Health ping file not found"))
+            exit(-1)
+
+        self.stdout.write(f"Last run on this worker at: {last_run_at}")
+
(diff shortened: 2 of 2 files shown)


ntoll (approved):
Code looks good. :+1: 

Trying to get my head around what exactly it does. I think it does this (just writing it out so I get it clear in my head):

* k8s probe causes the command to be run and checks the time a known `worker_health.txt` file was modified.
* The command puts something on the message queue.
* This message could be processed by any of the N number of pods running the worker process.
* Once the message is processed, then the worker (which could be on any pod) `touch`'s the known file on that specific pod instance to indicate the message was received.
* The k8s probe might need tweaking so that failure happens only after some number of seconds so that the load `worker_health.txt` file can be touched.

Questions (perhaps for webops folks): 

* can the results of the probes for the N pods be aggregated in some way - for instance, so that success is indicated by at least one having a `worker_health.txt` file updated within `n` seconds..?
* alternatively, do the worker pods have access to a shared mounted directory into which the `worker_health.txt` could be written / checked..? (i.e. they're all writing to the same place).

Suggestion: 

* write up how this works in the wiki for our future selves/colleagues. :wink:

xoen:
FYI: I've renamed the command argument to `--stale-after-secs` to make it more explicit/clear. 
I've also added more beef to how the k8s probe will be configured/works in both the commit message and the PR description. 

I've chatted with Andy and AFAICT he found this clear but if note I hope this helps.

I think with this is the case of understanding how probes work in kubernetes which can be a bit confusing at first.

xoen:
FYI: Also have a look at "twin" PR with changes in Helm chart and explanation may help: https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/468

xoen:
@ntoll regarding your comment:

First part basically yeah, you understood it ☺️.

> can the results of the probes for the N pods be aggregated in some way - for instance, so that success is indicated by at least one having a worker_health.txt file updated within n seconds..?

Not sure I understand the question. But each worker container is independent and checked independently. Each container has its own health ping file which is "touched" only when that specific worker has processed a health task.

Above actually works better than it sounds because jobs are picked up/performed by all the workers uniformly and unless a worker is having a wobble they'll process roughly a message every 30s (confirmed by `dev` testing/observation).

> alternatively, do the worker pods have access to a shared mounted directory into which the worker_health.txt could be written / checked..? (i.e. they're all writing to the same place).

No, each pod is independent and they don't share state and they shouldn't. Bear in mind each pod could physically be sitting at hundreds of km apart from each other. E.g. running on different nodes in different Availability Zones (and this in general what happens and it's a good thing for reliability reasons).

So a shared mounted volume wouldn't be good.

Additionally a liveliness probe for a specific `worker` container should ideally succeed/pass independently from other containers for the same `Deployment`.

Home above (plus additional write up on how the `livelinessProbe` is configured and the helm chart PR) helped clarify how this works ☺️

PR #842, opened 2020-09-18
Allow admins to order columns in user list.

ntoll:
## What

In the end, the actual change needed was very small (and it should be easy to add to other pages that need ordering by column).

Clicking on the column title will re-load the page with the table ordered by column selected column. See screenie below for this behaviour demonstrated:

order_columns

## How to review

1. Go visit `dev` as an admin.
2. Click on the user list.
3. Order the columns by clicking on the column title as per the screenie above.

Related Trello ticket: https://trello.com/c/cApsTxp5/781-add-ordering-of-columns-in-admin-lists-on-control-panel

### DIFF
--- a/controlpanel/frontend/views/user.py
+++ b/controlpanel/frontend/views/user.py
@@ -29,14 +29,17 @@ class UserList(LoginRequiredMixin, PermissionRequiredMixin, ListView):
     queryset = User.objects.exclude(auth0_id='')
     template_name = "user-list.html"
 
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
+    def get_ordering(self):
         order_by = self.request.GET.get("o", "username")
         valid_fields = ["username", "email", "last_login", ]
         if order_by not in valid_fields:
             order_by = "username"
+        return order_by
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
         unused_users = []
-        for user in self.queryset.order_by(order_by):
+        for user in self.queryset:
             if user.last_login:
                 if user.last_login.date() < ninety_days_ago():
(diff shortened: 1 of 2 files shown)


controlpanel/frontend/views/user.py
```diff
@@ -29,6 +29,13 @@ class UserList(LoginRequiredMixin, PermissionRequiredMixin, ListView):
     queryset = User.objects.exclude(auth0_id='')
     template_name = "user-list.html"
 
+    def get_ordering(self):
+        order_by = self.request.GET.get("o", "username")
```
xoen, line 33: Is a Django convention to call this ordering query string param just `o`? Or we could change to something more explicit like `order_by` for example? (basically same as you named the variable).

xoen (approved):
Nice one Nick,

(I guess at some later point we may decide to have a separate column for `username` and full name to sort these independently but this is already a great improvement for Cal)

controlpanel/frontend/views/user.py
xoen: (TBH from screenie, it already seems to be sorted alphabetically by `username` so that's good)

ntoll: So this is the golden nugget of information: yes, it's the convention (but finding out the conventions often involves knowing what to look for / ask). When I'd done this before I'd either used a JS library (which I wanted to avoid) or done it with response functions (rather than classes).

Every model can have a default ordering IIRC. For users I believe that's username, otherwise it's PK if not set. Since we inherit from `django.contrib.auth.User` then the default was sort-by-username.

The beauty is we can override this in such a small amount of code. :-)

ntoll:
Thanks @xoen. I'll merge and cut a release on Monday.

PR #843, opened 2020-09-30
Ability to connect securely to Redis with SSL/TLS (REDIS_SCHEME setting)

xoen:
### What
Ability to optionally connect to Redis using SSL/TLS.

Added `REDIS_SCHEME` settings, defaults to `redis` for retro-compatibility but `rediss` can be passed to allow secure connection to Redis cluster.

### Why
This is necessary in order to connect to the Redis cluster on AWS ElastiCache which uses TLS for
encryption in transit as explained here: https://github.com/jazzband/django-redis/issues/353#issuecomment-447799503

[Redis cluster is there on AWC EC](https://github.com/ministryofjustice/analytics-platform-ops/pull/368/) and should be in theory accessible but we see problems that we pinpointed to the use of TLS for encryption in traffic. This is an attempt to tell CP to use TLS when talking with Redis.

### Retro-compatibility
This change should be retro-compatible as we still default to use `redis://` when this setting/environment variable is not set so the end result on existing helm chart/pods should be to connect with the same connection string in theory.

### Ticket
Part of ticket: https://trello.com/c/YG9anK8A

### DIFF
--- a/controlpanel/settings/common.py
+++ b/controlpanel/settings/common.py
@@ -456,16 +456,25 @@ NFS_HOSTNAME = os.environ.get("NFS_HOSTNAME")
 REDIS_HOST = os.environ.get('REDIS_HOST', 'localhost')
 REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD')
 REDIS_PORT = os.environ.get('REDIS_PORT', '6379')
+REDIS_SCHEME = os.environ.get("REDIS_SCHEME", "redis")
+if REDIS_SCHEME not in ["redis", "rediss"]:
+    raise ValueError(f"Invalid value for 'REDIS_SCHEME' environment variable. Must be 'redis' or 'rediss' (to use SSL/TLS). It was '{REDIS_SCHEME}' which is invalid.")
+
+REDIS_URI = f"{REDIS_SCHEME}://{REDIS_HOST}:{REDIS_PORT}/1"
 
 # -- Async
 
 ASGI_APPLICATION = f"{PROJECT_NAME}.routing.application"
 
+# See: https://pypi.org/project/channels-redis/
 CHANNEL_LAYERS = {
     'default': {
         'BACKEND': 'channels_redis.core.RedisChannelLayer',
         'CONFIG': {
-            'hosts': [{'address': (REDIS_HOST, REDIS_PORT)}],
+            'hosts': [{
+                "address": REDIS_URI,
+                "timeout": 30,
+            }],
         },
     },
 }
@@ -477,7 +486,7 @@ if REDIS_HOST and REDIS_PORT and REDIS_PASSWORD:
     CACHES = {
         "default": {
             "BACKEND": "django_redis.cache.RedisCache",
-            "LOCATION": f"redis://{REDIS_HOST}:{REDIS_PORT}/1",
+            "LOCATION": REDIS_URI,
             "OPTIONS": {
                 "CLIENT_CLASS": "django_redis.client.DefaultClient",
                 "PASSWORD": f"{REDIS_PASSWORD}"


xoen:
Regarding this, both `frontend` and `worker` pods now can successfully connect to [Redis cluster with encryption in-transit/TLS in `dev`](https://eu-west-1.console.aws.amazon.com/elasticache/home?region=eu-west-1#redis:) (see [PR with Terraform changes](https://github.com/ministryofjustice/analytics-platform-ops/pull/368/files) which I'm also tidying up)

In theory (but please let me know your thoughts) I've added this setting so that:
1. if environment variable is not set (as it will be the case while helm chart changes follow) it will default to `redis` scheme, e.g. will connect as it does now (using `redis://...`)
2. once we're happy with this, we can merge and create a release for this which will re-release using old helm chart and it should hopefully work as before (connecting to current redis node running on k8s)
3. I'm in the process of updating helm chart (to add this value for `REDIS_SCHEME`, remove Redis dependency etc...)
4. that may be a bit tricky but in theory as long as `alpha` Redis cluster is also there and config is correct re-deploy should work

In regard to this PR, point 1 and 2 should mean that this can be merged/released independently and should be retro-compatible as it should only add the setting to use `rediss://` instead of `redis://`...other points may need some caution but I'll let you know once I finish the polishing of these changes to helm chart/config/terraform.

Does it make sense?

controlpanel/settings/common.py
```diff
@@ -456,16 +456,25 @@
 REDIS_HOST = os.environ.get('REDIS_HOST', 'localhost')
 REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD')
 REDIS_PORT = os.environ.get('REDIS_PORT', '6379')
+REDIS_SCHEME = os.environ.get("REDIS_SCHEME", "redis")
+if REDIS_SCHEME not in ["redis", "rediss"]:
+    raise ValueError(f"Invalid value for 'REDIS_SCHEME' environment variable. Must be 'redis' or 'rediss' (to use SSL/TLS). It was '{REDIS_SCHEME}' which is invalid.")
+
+REDIS_URI = f"{REDIS_SCHEME}://{REDIS_HOST}:{REDIS_PORT}/1"
```
ntoll, line 463: Why the digit `1` at the end of this URI..? (It didn't used to have one... no pun intended).

controlpanel/settings/common.py
```diff
@@ -456,16 +456,25 @@
...
 # -- Async
 
 ASGI_APPLICATION = f"{PROJECT_NAME}.routing.application"
 
+# See: https://pypi.org/project/channels-redis/
 CHANNEL_LAYERS = {
     'default': {
         'BACKEND': 'channels_redis.core.RedisChannelLayer',
         'CONFIG': {
-            'hosts': [{'address': (REDIS_HOST, REDIS_PORT)}],
```
ntoll, line 468: Hmm... I notice this doesn't have a `1`.

controlpanel/settings/common.py
```diff
@@ -456,16 +456,25 @@
...
+            'hosts': [{
+                "address": REDIS_URI,
```
ntoll, line 475: Yet this will..! (Have a `1`)

controlpanel/settings/common.py
```diff
@@ -477,7 +486,7 @@
     CACHES = {
         "default": {
             "BACKEND": "django_redis.cache.RedisCache",
-            "LOCATION": f"redis://{REDIS_HOST}:{REDIS_PORT}/1",
```
ntoll, line 480: Aha.... found the "`1`". Still not sure of its significance or (apparent) difference in use from before (see comments above).

xoen: I'm not massively familiar with Redis neither but that number after the "slash" is the DB it will use: https://stackoverflow.com/a/20770414/455642

It seemed to work fine in `dev`, both login/session and background jobs submission/execution. I suspect this may be more useful when you share a Redis cluster among several applications or something like that.

ntoll: Ahh... yes. I know about REDIS database numbers, but didn't realise this was the reference to it. ;-) :+1:

xoen: Me neither, I only found out while refactoring this.

In theory unless both Django and Django Channels use the same keys it should be a problem for them to use the same DB/keyspace and it doesn't seem to be a problem in `dev`, but do you think we should use two DBs here?

ntoll (approved):

PR #845, opened 2020-10-16
[MVP] Show all datasources to users (including ones they can't acces)

xoen:
### What

**NOTE**: This is an MVP:

> [...] on the Warehouse Data Sources and App Data Sources tabs, below the list of buckets they have access to, let's show a list of all the other buckets too i.e. the ones they DON'T have access to.
> [...]
> this ticket is for an MVP, which is just displaying the name, and letting the user request access in the normal ways.)

Main reason for this is to give visibility of existing data sources to users. They have a rough idea of what's already available.
They can therefore avoid reinventing the wheel if the data is already present.

**NOTE**: This do **not** change [the process for requesting access to datasources](https://user-guidance.services.alpha.mojanalytics.xyz/data/amazon-s3/#request-access-to-a-bucket). Users will still request access to these as before.

### Privacy concerns
We discussed this few times and the team is happy to show the list of datasources to all users. 
We only show the datasources' names and their admins so it shouldn't constitute a problem.

The items in the list don't link to the actual datasources or users so no permission changes were required.

Also note that users will be able to see only the admins list (which was already a "nice to have" on top of the MVP), **not** the full list of users for each datasource and their access level.

### Ticket
https://trello.com/c/AIOEPlHG

### DIFF
--- a/controlpanel/frontend/jinja2/datasource-list.html
+++ b/controlpanel/frontend/jinja2/datasource-list.html
@@ -34,10 +34,10 @@
 {% if other_datasources %}
   <h3 class="govuk-heading-m">Other {{ datasource_type }} data sources</h3>
   <p class="govuk-body">
-    The following is a list of {{ datasource_type }} data sources you currently <span class="govuk-!-font-weight-bold">do not</span> have access to.
+    You currently <span class="govuk-!-font-weight-bold">do not</span> have access to the following {{ datasource_type }} data sources.
   </p>
   <p class="govuk-body">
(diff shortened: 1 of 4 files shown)


ntoll (approved):
Code looks great. Nice and simple.

Also, I didn't know about `prefetch_related`. You learn something new every day. :+1:

:shipit:

xoen:
Yeah I didn't know about `prefetch_related` but hopefully will avoid the classic n+1 classic problem ☺️

PR #846, opened 2020-10-19
Add admin functions for creating, retrieving, updating and deleting releases of tools in the control panel.

ntoll:
## What

This release adds functionality to the admin area of the control panel that allows us to add multiple releases of tools and/or releases that are restricted to only certain users. You can see a GIF of the general workflow below.

release_crud

This release adds a couple of new columns to the `Tool` model and, as a result, a new database migration script that needs to be run as part of the deployment process (the migration script is automatically generated by Django and we don't need to review it). I've also ensured that the permissions required for being able to do any of this activity are only available to super-users.

**NOTE**
Once folks are happy about my database changes (the names and so on) I'll try this out on our `dev` instance. (As far as I can tell, the `manage.py migrate` command is automatically run when a new version of the control panel is deployed anyway.) Why the caution..? Migrating a database is a one-way operation and I don't want to do it until I've had folks agree that my changes look good. I will remove the DRAFT status of this PR once I've seen it working on `dev`. Best to be safe rather than sorry..!

I've tried to piggy-back on Django's built in CRUD-y things (why else should we use a framework if we don't allow it to do the heavy lifting for us?). This involves some new class views. When I've had to add code that's beyond the usual Django way to do things (i.e. when processing user names for the target users list) then I've added a unit test too.

Finally, as suggested by David, the naming shouldn't be specific to "beta" users, so I've gone with the more generic "restricted" release language, along with target users.

## How to review

1. Grab the branch
2. Run `./manage.py migrate` against your local database **THIS CANNOT BE UNDONE**.
2.1 (N.B. I've had to drop database and recreate a blank database a couple of times during the testing of this code - something you shouldn't have to do, but the instructions in the `docs/running.md` page should help with this -- assuming we remember to update the AWS section now that things have changed around there).
3. `make test` - to run the unit tests
4. `./manage.py runserver` and visit the site as an admin user. Play through the CRUD lifecycle of a release and let me know if you spot anything.

This work is related to [this Trello ticket](https://trello.com/c/FIWqwKn1/780-add-means-of-adding-new-tool-releases-to-the-control-panel).


controlpanel/api/rules.py
```diff
@@ -145,6 +145,12 @@ def is_self(user, other):
 add_perm('api.reset_mfa', is_authenticated & is_superuser)
 
 
+add_perm('api.list_release', is_authenticated & is_superuser)
+add_perm('api.create_release', is_authenticated & is_superuser)
+add_perm('api.destroy_release', is_authenticated & is_superuser)
+add_perm('api.update_release', is_authenticated & is_superuser)
```
xoen, line 151: I think there is some ambiguity here which we can remove by adding the word "tool" to the mix.

Users currently don't release webapps via the Control Panel but being more explicit here doesn't harm as it helps avoid any possible confusion.

I see below the form for example is called `ToolReleaseForm` which is an **excellent and very clear** name, so consider using something similar for the permissions name as well.

ntoll: We're configuring *releases* of Tools (i.e. a tool is Jupyter or RStudio). The name of the model is wrong IMHO :-P the `Tool` model should be called something like `ToolRelease`. :+1: But I felt that was a refactor too far. ;-)

ntoll: Having said that.... I'm not really fussed what we call things, but the way folks have talked about this work in the past is in terms of "releases" rather than "tools" (hence my following this convention). Basically, we've just gotta agree names... ;-)

xoen: Agreed on agreeing names. 

These are permissions to list tool releases, create a tool release, etc...hence my suggestion. 
Within our AP domain there are also webapp releases. Users currently don't manage these in the Control Panel, but I still think it's less confusing to call the permissions above something like `api.list_tool_release`, `api.create_tool_release`, etc...as sidestep the potential ambiguity and issue.

ntoll: Ahh... I see what you mean now. :+1: on that then. ;-)

xoen: Sorry must have explained myself really badly in the first comment 🤦

controlpanel/frontend/jinja2/home.html
```diff
@@ -13,6 +13,7 @@ <h2 class="govuk-heading-m">Superuser functions</h2>
     <li><a href="{{ url('list-users') }}">List all users</a></li>
     <li><a href="{{ url('list-all-apps') }}">List all apps</a></li>
     <li><a href="{{ url('list-all-datasources') }}">List all data sources</a></li>
+    <li><a href="{{ url('list-releases') }}">List all releases</a></li>
```
xoen, line 16: Similar comment to one above about making this more explicit:

`url('list-tools-releases')` and `List all tools releases` (or similar) avoid potential confusion for "non-initiated" people which may confuse this with webapp releases somehow.

controlpanel/api/models/tool.py
```diff
@@ -27,6 +27,10 @@ class Tool(TimeStampedModel):
     name = models.CharField(max_length=100, blank=False)
     values = JSONField(default=dict)
     version = models.CharField(max_length=100, blank=False)
+    # The release is restricted to only certain users.
+    is_restricted = models.BooleanField(default=False)
```
xoen, line 31: I haven't got to the meat of the PR yet (the logic to show/hide tools versions to the users) so maybe this field is here to simplify that logic/query but isn't a release restricted when there are 1+ target_users associated with it?

What I'm wondering is if this field is redundant (just writing it down, maybe it's a silly observation will clear itself once I get to the logic using it)

controlpanel/frontend/jinja2/release-create.html
```diff
@@ -0,0 +1,109 @@
...
+      "errorMessage": { "html": form.is_restricted.errors|join(". ") } if form.is_restricted.errors else {}
+    }) }}
+    {{ govukInput({
+      "label": {
+        "text": "Target Users",
+        "classes": "govuk-label--m",
+      },
+      "classes": "govuk-!-width-two-thirds",
+      "hint": {
+        "text": 'A comma separated list of usernames of usernames of users who have access to this release'
```
xoen, line 100: duplicated "of usernames of usernames", glad I'm not the only one glitching and writing things like this 😝

controlpanel/api/models/tool.py
ntoll: Good question and I thought long and hard about it. In the end I went with:

* This is explicit.
* This is unambiguous.
* The name of the flag says what it means (thus helping future self/maintainers).

I also think it makes the logic simpler to understand when used elsewhere - you're checking a boolean flag rather than comparing values.

In any case, this column is a negligible addition but has the benefits described above. Hope this makes sense..! :+1:

controlpanel/frontend/jinja2/release-create.html
ntoll: What do you what do you mean mean about glitching out out glitching..? :stuck_out_tongue_closed_eyes: Nothing wrong wrong with that.

(Seriously, good catch, I'll fix.)

controlpanel/frontend/views/release.py
```diff
@@ -0,0 +1,90 @@
...
+
+class ReleaseList(LoginRequiredMixin, PermissionRequiredMixin, ListView):
+    """
+    Used to display a list of releases of the tools available in the control
+    panel application.
+    """
+    context_object_name = 'releases'
+    model = Tool 
+    permission_required = 'api.list_release'
+    queryset = Tool.objects.all()
```
xoen, line 22: Are `queryset` and `ordering` required when `model` is set? 
I suppose it would use ordering from `Tool` model if not right? (less is more but this is not an issue with the code, more of a question)

ntoll: The model is used for introspection so Django gets to know about the fields, their descriptions and constraints when doing things like, say, a `ModelForm` (based on such metadata).

The default Django ordering would apply, which is PK IIRC. It's easy to change. I suspect alphabetical order on the `name` field and then `release` in `Tools` would be a sensible default..? Happy to make this happen.

xoen (approved):
Nice work Nicholas ☺️ 

It all looks sensible, I added some comments but nothing wrong with this. 
I suppose my main (and possibly annoying!) point would be on the URLs/permissions/templates etc names missing the word "tool". 

At the moment we don't manage webapps releases via the Control Panel but they're still a thing and **potential** source of ambiguity for new devs/users if we call "tools releases" just "releases".

Having said that, there are a lot of occurrences so I know it may be a bit of a chore to rename those. Not a huge deal anyway, just mentioning it.

controlpanel/frontend/views/release.py
xoen: Yeah I like the ordering you chose here, it makes a lot of sense so if you decide to move it to the model I'll be more than happy with that ☺️

controlpanel/api/models/tool.py
xoen: Yeah I saw the logic and I think the logic with `Q` and this flag is probably easier than having to select records with no associated stuff.

Cool, happy with this field 👍

controlpanel/api/rules.py
ntoll: Fixed

controlpanel/frontend/views/release.py
ntoll: The ordering is already in place - see line 24. ;-)

controlpanel/frontend/jinja2/release-create.html
ntoll: Fixed fixed

controlpanel/frontend/jinja2/home.html
ntoll: Fixed.

ntoll:
Thanks for all the input. I'll merge and cut a release now.

PR #847, opened 2020-10-22
CI: Replacing CircleCI with GitHub Actions

xoen:
### Why?

We're aiming to simplify our current CI/CD and reduce the number of
technologies involved.
Ideally we'd like to use GitHub Actions for all things CI/CD.
This is a step in that direction and remove CircleCI out of the equation.

### What

The workflow is very very similar to old CircleCI workflow with
few tweaks:
- Docker caching is done using the `satackey/action-docker-layer-caching`
  action which is doing the same docker save/load under the hood
- `CIRCLE_SHA1`/`CIRCLE_BRANCH` replaced with more long winded but
  equivalent way of getting the same information in GH Actions
- Quay.io username/password stored in GH Repository secrets instead of
  being stored in CircleCI
- Speaking of secrets, the credentials/robot for this GH Actions are new
  and different from old CircleCI one to avoid any possible confusion

NOTE: `npm install` in `testjs` job is cached but the
benefits are very small as non-cached `npm-instal` was
taking 16-20s anyway.

### Getting branch name and commit hash

Unfortunately GH Actions doesn't seem as polished as CircleCI
when it comes to getting basic info like branch name and commit hash.

There is no easy way (that I found) to get these info by using a simple
environment variable and even the available variables like `GITHUB_REF`,
`GITHUB_HEAD_REF` and `GITHUB_SHA` have different values depending on
what triggered the build - how confusing!!

This is the reason why the `Get branch name (when xxx)` and
`Get commit SHA1 (when xxx)` are not as simple as I'd like.

These steps use may run on not depending on what triggered the build,
e.g. a Pull Request or else (hence the `if: github.event_name ...`)

To make things worse sometimes to get the correct info you have to get
it from the event payload like in the case of the commit hash when it's
a PR (`github.event.pull_request.head.sha`) or otherwise you'll get an
odd commit hash instead.

See resources I've found to help me get these info for more details:
- Get branch name: https://stackoverflow.com/a/64210623/455642
- Get correct commit hash: https://github.community/t/github-sha-isnt-the-value-expected/17903

### Further splitting in smaller jobs

I tried to split the current final job into smaller, independent, jobs which
were:
- building the docker image
- running the tests with docker-compose
- push to quay (once tests passed)

Bear in mind different jobs don't share state.

I tried to use GH Actions artifacts to store the docker image but this
was slow because of size of docker image and upload/download required
to transfer this between jobs.
In the end the pragmatic thing to do was to do all these things in the
final job together (which was also the approach taken in the old
CircleCI workflow).

### Ticket

Ticket: https://trello.com/c/jyRlofox


xoen:
FYI: Stating the obvious here but the CircleCI is "failing" because I've removed the `.circleci/` directory in this PR.

.github/workflows/test-and-push-docker-image.yaml
```diff
@@ -0,0 +1,122 @@
...
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
```
damacus, line 35: Should we cache using the matrix version too? Given the packages are likely to be different between Node versions.

Or should we just pick one version of Node

.github/workflows/test-and-push-docker-image.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Install JavaScript dependencies
+        run: |
+          npm install
```
damacus, line 39:
```suggestion
        run: npm install
```

This can just be one line to tidy things up slightly. Maybe?

damacus (commented):
Is it worth reducing this down to one Node version, considering we're only using one version in Prod?

.github/workflows/test-and-push-docker-image.yaml
xoen: > Should we cache using the matrix version too? Given the packages are likely to be different between Node versions.

I thought about that, and my conclusion was that it wouldn't matter and `npm install` would handle that anyway. So I kept it simple. TBH, this caching doesn't make much of a difference in our case as run the `npm install` took between 16-20 seconds before the caching 😬 

What do you reckon?

> Or should we just pick one version of Node
We weren't doing that in the old CircleCI workflow but I thought it was a good idea as it will give us visibility over JS stuff breaking in future node versions before we decide to upgrade.

xoen: From https://github.com/actions/cache/blob/main/examples.md#node---npm : 

> Note: It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci

It doesn't seem to break in our case but I'll tell you what, since the benefit is marginal (maybe 10s in the best case) I'll just keep it simple and remove the caching of this altogether unless you have any other insight here.

.github/workflows/test-and-push-docker-image.yaml
xoen: 🤔 😬  I kinda liked the consistency of these but I'll move to same line when there is only 1 command if it pleases you ☺️

xoen:
@damacus FYI: I've removed the Node.js caching to keep things simple: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/847/commits/2c59a203f220e7ddd1202afbb5b5964fbfe7759d

damacus (approved):
source

xoen:
Thanks @damacus ⭐ ☺️

PR #848, opened 2020-10-22
Add accessibility statement. First draft.

ntoll:
## What

This PR adds several things:

* An "Accessibility Statement" link into our footer.
* A view for displaying the a11y statement.
* A template containing the content, based upon [this sample](https://www.gov.uk/government/publications/sample-accessibility-statement/sample-accessibility-statement-for-a-fictional-public-sector-website).

I've added the suggested site-specific comments to the appropriate sections of the template:

* we have been through an accessibility checklist for the Control Panel and believe it is compliant
* we understand there is reasonable accessibility compliance of R Studio and Jupyter, and will be updating to the latest versions
* we are in the process of booking a full accessibility audit of the Analaytical Platform and will promptly update this statement

**PLEASE CHECK THE CONTENT** - feedback and suggested changes most welcome as part of the review process for this PR.

## How to review

1. Start the app locally then log in.
2. Find the a11y statement link in the footer.
3. Click it and wonder at the content therein.

Related Trello ticket: https://trello.com/c/Fx2OdCRq/791-write-an-accessibility-statement

### DIFF
--- a/controlpanel/frontend/jinja2/a11y.html
+++ b/controlpanel/frontend/jinja2/a11y.html
@@ -12,9 +12,9 @@
     <p>This accessibility statement applies to the Ministry of Justice's
     Analytical Platform.</p>
 
-    <p>This service is run by Analytical Platform Team of the Ministry of
+    <p>This service is run by Analytical Platform team of the Ministry of
     Justice. We want as many people as possible to be able to use this service.
-    For example, on the Control Panel website (this website) you
+    For example, on the Control Panel application (this website) you
     should be able to:</p>
 
     <ul class="govuk-list govuk-list--bullet">
@@ -27,7 +27,7 @@
 
     <p>We've also made our textual content as simple as possible to understand.</p>
 
-    <p><a rel="external" href="https://mcmw.abilitynet.org.uk/" class="govuk-link">AbilityNet</a>
+    <p><a rel="external" target="_blank" href="https://mcmw.abilitynet.org.uk/" class="govuk-link">AbilityNet</a>
     has advice on making your device easier to use if you have a disability.</p>
 
     <h3 class="govuk-heading-m" id="how-accessible-this-service-is">How accessible this service is</h3>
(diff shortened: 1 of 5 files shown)


controlpanel/frontend/jinja2/base.html
```diff
@@ -154,6 +154,10 @@
             'href': "https://asdslack.slack.com/messages/C4PF7QAJZ#",
             'text': "Analytical platform slack channel",
           },
+          {
+            'href': url('accessibility'),
+            'text': "Accessibility Statement",
```
davidread, line 159:
```suggestion
            'text': "Accessibility statement",
```

davidread (commented):
Review of the content only. Please could someone check it displays/works ok?

Content is really good. Just a couple of niggles.

* It feels slightly confusing that the title says it's for the whole AP but the first sentence says it's for CP. Maybe the bit that says "For example," say "For example, on the Control Panel:".

* I'm not keen on calling AP or CP a "website". How about we simply call it a "service"?

controlpanel/frontend/jinja2/base.html
ntoll: Done

ntoll:
@davidread thanks for the feedback. I've just pushed some updates to address your points:

* Clarified the copy at the start of the statement.
* Fixed the footer capitalization (or not, as the case may be)
* Clarified service/website. The template was for a *website*, and this term is used all over the place in the template in a way that makes sense for the term "website" (if you see what I mean) rather than "platform". However, I've made it clear in the opening paragraph that this a11y statement is for the *platform* but use the example of the control panel website when describing all the aspects of our a11y work. Does this make sense..?

In any case, as always, happy to change, refine etc... this stuff takes thought and is never right on first draft or via a handy template.

controlpanel/frontend/jinja2/a11y.html
```diff
@@ -0,0 +1,103 @@
...
+      <li>listen to most of the website using a screen reader (including the
+      most recent versions of JAWS, NVDA and VoiceOver)</li>
+    </ul>
+
+    <p>We've also made the website text as simple as possible to understand.</p>
+
+    <p><a rel="external" href="https://mcmw.abilitynet.org.uk/" class="govuk-link">AbilityNet</a>
+    has advice on making your device easier to use if you have a disability.</p>
+
+    <h3 class="govuk-heading-m" id="how-accessible-this-website-is">How accessible this website is</h3>
```
davidread, line 33: This section is about the wider platform, not "this website"

controlpanel/frontend/jinja2/a11y.html
```diff
@@ -0,0 +1,103 @@
...
+
+    <p>We know some parts of this website are not fully accessible:</p>
```
davidread, line 35: ditto

davidread (commented):
I still contend that the only point you should be talking about the website is in this one sentence: "For example, on the Control Panel website (this website) you should be able to:"

ntoll:
@davidread ok.... I'll `s/website/service/g` all the things (and check it still reads OK). ;-)

ntoll:
@davidread I've retained the "website" mention in the very first section as per our discussion of it being an example of how we address a11y. All other mentions of "website" are now "service". :+1:

davidread:
Text all looks great now 👍

davidread (approved):
I've tested in a browser now. Some minor comments for you to review, but I'll approve in advance.

* Control Panel should probably be capitalized
* "control panel application" seems inconsistent with "Control Panel website" - either's fine, but let's be consistent
* "Analytical Platform team" or "Analytical Platform Team" - again, consistency
* External links should open a new browser tab (e.g. slack, user-guidance, eass, 18f.gov)

ntoll:
@davidread thanks for the feedback. I've fixed the copy as per your suggestions and will merge once CI turns green. :+1:

PR #849, opened 2020-10-30
Add help tab for user support content/videos

ntoll:
## What

This is a work in progress. Discussion of content/copy should happen in this PR ticket and I'll refine the code as we find a consensus for what this should look like.

## How to review

1. Start the app locally.
2. Log in.
2. Click the "Help" tab.

Related Trello tickets:

* https://trello.com/c/NQTZLrqf/825-add-a-help-tab-to-the-control-panel-for-easy-to-access-user-guidance-and-videos
* https://trello.com/c/ghEd5K3C/810-create-a-reset-user-home-2-min-video-training-resource-to-explain-the-process-and-reasons-for-using-it
* https://trello.com/c/LiXzZWYq/807-create-a-how-to-start-using-ap-2-minute-video-training-resource-giving-step-by-step-instructions-for-signing-up-and-logging-in

### DIFF
--- a/controlpanel/frontend/jinja2/help.html
+++ b/controlpanel/frontend/jinja2/help.html
@@ -1,0 +1,25 @@
+{% extends "base.html" %}
+
+{% set page_title = "Help" %}
+
+{% block content %}
+  <div class="govuk-grid-row" id="reset-container">
+      <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Help</h1>
+
+    <p class="govuk-body">Common queries, support tasks and problems are
+    explained in the following videos.</p>
+
+    <p class="govuk-body">Should you still require help, please visit our
+    <a href="https://asdslack.slack.com/messages/C4PF7QAJZ#">support channel on Slack</a>, ask your question whilst remembering to tell us your
+    username and the name of any affected S3 buckets or applications on the
+    analytical platform that are part of the context of your request.</p>
+
+    <h2 class="govuk-heading-m">How do I..?</h2>
+
+    <p>Home Reset:</p>
+
+    <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/ziu8dWRRcVM" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+      </div>
+    </div>
+{% endblock %}
--- a/controlpanel/frontend/jinja2/base.html
+++ b/controlpanel/frontend/jinja2/base.html
@@ -121,6 +121,11 @@
         "active": page_name == "groups",
         "admin": request.user.is_superuser,
       },
+      {
+        "text": "Help",
+        "href": url("help"),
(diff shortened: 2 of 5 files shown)


controlpanel/frontend/jinja2/help.html
```diff
@@ -0,0 +1,25 @@
+{% extends "base.html" %}
+
+{% set page_title = "Help" %}
+
+{% block content %}
+  <div class="govuk-grid-row" id="reset-container">
```
xoen, line 6: I'm assuming this `id="reset-container"` is just leftover from some other view ☺️ (I don't think it should matter anyway unless we have any JS that uses these IDs, right?)

ntoll: Doh. Correct. Will fix. ;-)

controlpanel/frontend/jinja2/help.html
```diff
@@ -0,0 +1,25 @@
...
+      <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Help</h1>
+
+    <p class="govuk-body">Common queries, support tasks and problems are
+    explained in the following videos.</p>
+
+    <p class="govuk-body">Should you still require help, please visit our
```
xoen, line 13: Question: Should we add a paragraph or mention the User Guidance here as well so that we somehow nudge users to look at the documentation before raising a support request? Thoughts?

ntoll: Yes we should. Feedback and suggestions like this are exactly what I'm looking for. While the code is in place, the copy for what should be on the page is not... so feel free to tell me what you think should be there. :+1:

controlpanel/frontend/jinja2/help.html
```diff
@@ -0,0 +1,25 @@
...
+    <a href="https://asdslack.slack.com/messages/C4PF7QAJZ#">support channel on Slack</a>, ask your question whilst remembering to tell us your
+    username and the name of any affected S3 buckets or applications on the
```
xoen, line 15: I would probably be pedantic here and say "GitHub username" just to be sure users don't get confused with who knows what username 😬  (e.g. some users log into O365 using their email address as username IIRC so being more explicit may avoid confusion)

controlpanel/frontend/jinja2/help.html
```diff
@@ -0,0 +1,25 @@
...
+    analytical platform that are part of the context of your request.</p>
```
xoen, line 16: Can't comment on the wording here but I have two observations:
1. do you think also mentioning something about "include any error you get" would be too much?
2. I notice we have a long-winded version of this nice summary on how to get support in the [User Guidance](https://user-guidance.services.alpha.mojanalytics.xyz/support.html#content), we may also decide to link to that.

xoen (approved):
Hey @ntoll this is great ✨ !

I know this is in "Draft" but I had a look and left some comments anyway (just some potential ideas more than anything, feel free to either incorporate or ignore)

ntoll:
@xoen :+1:

PR #850, opened 2020-11-02
Add timeout to log search.

ntoll:
## What

This change ensures that if Elastic is down/off then empty logs are returned rather than a 500 error.

## How to review

1. Will deploy to `dev` (where Elastic is currently switched off).
2. Take a look at an app and note there isn't a 500 error.

Related Trello ticket here: https://trello.com/c/IO05G008/818-control-panel-shouldnt-500-if-logging-for-webapp-is-off-broken

### DIFF
--- a/controlpanel/api/elasticsearch.py
+++ b/controlpanel/api/elasticsearch.py
@@ -1,8 +1,12 @@
+import logging
 from dateutil.parser import parse
 from django.conf import settings
 from elasticsearch import Elasticsearch
 from elasticsearch_dsl import Q, Search
 from elasticsearch_dsl.query import Range
+
+
+log = logging.getLogger(__name__)
 
 
 def bucket_hits_aggregation(bucket_name, num_days=None):
@@ -29,6 +33,7 @@ def bucket_hits_aggregation(bucket_name, num_days=None):
 def app_logs(app, num_hours=None):
     # Disable logs for noisy app ("covid19-early-release")
     # to prevent its App details page to timeout
+    log.info("Get app logs for: " + str(app.id))
     if app.id == 171:
         return []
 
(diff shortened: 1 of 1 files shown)


controlpanel/api/elasticsearch.py
```diff
@@ -59,7 +68,13 @@ def app_logs(app, num_hours=None):
...
-    for entry in s.scan():
-        entry['timestamp'] = parse(entry['@timestamp'])
-        yield entry
 
+    try:
+        logs = list(s.scan())
+        log.info("Got {} log entries.".format(len(logs)))
+        for entry in logs:
+            entry['timestamp'] = parse(entry['@timestamp'])
+    except Exception as ex:
```
xoen, line 77: I'm assuming there isn't a more specific exception for when the request timeout 👌 .

controlpanel/api/elasticsearch.py
```diff
@@ -59,7 +68,13 @@ def app_logs(app, num_hours=None):
...
+        log.error(ex)
+        logs = []
```
xoen, line 79: I wonder if we could add a message in this case, something like "Failed to get logs for this webapp" or if it's a bad idea?

Probably the wrong place add this logic, alternatively we let the exception bubble up here and handle in the view but I don't know if that makes things more complicated.

xoen (approved):
LGTM 👍

controlpanel/api/elasticsearch.py
ntoll: I tried the expected `ConnectionError` but that wasn't caught. :-/

xoen: Yeah, I didn't find anything more specific neither among the exceptions on GH so probably this is the best we can do. 

I suppose it's the kind of additional information on the page that if can't be retrieved [because of an exception] it's not the end of the world and as I user I'd rather have the page mostly work, and I can always refresh to retry.

ntoll: I was going to say I could make the template output a helpful message if no logs are available. But the `app-log` macro calls directly to the API and I'm trying to work out why this is a good idea (i.e. I don't want to change something because I don't know the wider context).

controlpanel/api/elasticsearch.py
ntoll: The log message will include all the information to help us identify that this was to do with getting logs from Elastic (the module generating the exception is called something like `elastic_dsl` thus making it obvious what's going on).

ntoll:
OK... going to merge as-is and we can refine in subsequent tickets.

PR #851, opened 2020-11-09
Add a help page for home directory reset

ntoll:
## What

This branch builds on the work for the help tab. Rather than using video content, this PR includes a help page for a specific topic.

## How to review

1. Pull
2. Run
3. Click the "Help" tab and then the first link therein.
4. Take a look
5. Compare/contrast with video based tutorial.


xoen:
The downside of this approach is that adding more help topics will require adding a Jinja template, adding a new URL and re-deploying the CP. I'm not saying this is super-hard but it adds friction.

I'm wondering if an hybrid approach where we have a link from the Control Panel to either the User Guidance (preferred as users already should look for help there) or the Wiki would be preferable. The idea being that this will be in theory easier to expand with new content, etc...Plus non-devs could do it, potentially even some of our users themselves.

Maybe the Control Panel has a shortcut to where the documentation actually lives for convenience for the user and also so that they don't miss the fact this is available.

It's just an observation, I don't mean to criticise this. The code change itself look good, it should work and I'm sure it looks very pretty and pleasant for the users.

davidread:
Yeah I agree, writing help text in this place is a pain to sustain. I think keep all this sort of material in the User Guidance.

I suggest we have a Help page that simply has:

* Platform user guidance
* Platform support
* R Slack channel
* Python slack channel
* Analytical platform slack channel 

moved there from the footer.

User Guidance has [a section about the Reset Page](https://user-guidance.services.alpha.mojanalytics.xyz/tools.html#rstudio-memory-issues) but could usefully have material added e.g.:
```
    <p class="govuk-body">You need to reset your home directory if any of the
    following applies:</p>

    <ul class="govuk-list govuk-list--bullet">
        <li>Your RStudio won't start.</li>
        <li>Your Conda environment is a mess.</li>
        <li>You need to start RStudio development from a clean state.</li>
```

PR #852, opened 2020-11-09
Add 'list bucket version' permission on our S3 buckets.

ntoll:
## What

Due to an update in AWS, and because all our S3 buckets are versioned, all users need to have the `s3:ListBucketVersions` permission in order to be able to see what's in their buckets. See image reported by user in our support channel:

image

## How to review

1. ask @damacus


damacus (approved):

damacus:
As discussed, this won't actually fix the existing users as we don't manage the the roles.

PR #854, opened 2020-11-11
Revise AWS S3 actions

ntoll:
## What

This change ensures the actions for S3 buckets are of the correct *revised* types given recent S3 changes.

## How to review

1. Ask Dan to explain why the actions are revised to the new settings.

Related to https://github.com/ministryofjustice/analytics-platform/issues/121.

Trello reference: https://trello.com/c/HqDQMr6s/843-fix-s3-user-permissions

### DIFF
--- a/controlpanel/api/aws.py
+++ b/controlpanel/api/aws.py
@@ -96,14 +96,14 @@ BASE_S3_ACCESS_POLICY = {
     "Version": "2012-10-17",
     "Statement": [
         {
-            "Sid": "console",
+            "Sid": "ListUserBuckets",
             "Action": [
-                "s3:GetBucketLocation",
                 "s3:ListAllMyBuckets",
-                "s3:ListBucketVersions",
+                "s3:ListAccessPoints",
+                "s3:GetAccountPublicAccessBlock"
             ],
             "Effect": "Allow",
-            "Resource": ["arn:aws:s3:::*"],
+            "Resource": "*",
         },
     ],
 }
@@ -112,6 +112,8 @@ READ_ACTIONS = [
     's3:GetObject',
     's3:GetObjectAcl',
     's3:GetObjectVersion',
+    "s3:GetObjectVersionAcl",
+    "s3:GetObjectVersionTagging"
 ]
 
 WRITE_ACTIONS = [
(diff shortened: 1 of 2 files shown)


controlpanel/api/aws.py
```diff
@@ -96,14 +96,14 @@ def iam_assume_role_principal():
     "Version": "2012-10-17",
     "Statement": [
         {
-            "Sid": "console",
```
xoen, line 99: I'd be wary of changing the Sid of the statements. I can't remember if this specific one is used programmatically but at least some of them are (not just in CP but potentially from other tools, e.g. I think `s3tools` was checking the IAM policy to determine permissions at some point?)

Basically I'd leave this to `console` unless there are very good reason not to.

ntoll: :+1:

controlpanel/api/aws.py
```diff
@@ -96,14 +96,14 @@ def iam_assume_role_principal():
...
+            "Sid": "ListUserBuckets",
             "Action": [
-                "s3:GetBucketLocation",
                 "s3:ListAllMyBuckets",
-                "s3:ListBucketVersions",
```
xoen, line 103: Are we sure `s3:GetBucketLocation` and `s3:ListBucketVersions` are not used/required for certain operations? If they're not used than good to go 👍

xoen (approved):
The code itself looks fine to me 👍  but I've asked some questions regarding the IAM permissions, probably for @damacus .

Also, as said, probably best to keep the `Sid` as it was to be on the safe side unless you really have good reasons to change

ntoll:
Cheers @xoen - @damacus and I will revise and check. This is a bit of a fiddly fix to get right, and, like you, I think it'd be good to be (small "c") conservative in our approach.

controlpanel/api/aws.py
xoen: Without going too far, here an example of where some code relies on some of the `Sid` to have certain values (the new migration for example expects this to be `console`):

https://github.com/ministryofjustice/analytics-platform-ops/pull/372/files#diff-1bedc03903ba2fba183c28cf1ea8c8796f8c835f7fa09fbca156d308bff20946R140

ntoll: Great point, but the code you point to is uses the value `console` simply to migrate away from having a SID called `console`. :slightly_smiling_face:

controlpanel/api/aws.py
damacus: both of these have been moved to the LIST_ACTIONS block 👍

damacus:
Thanks for the review @xoen, that helped pull some hidden information out of the PR 👍

PR #856, opened 2020-11-16
Dockerfile: Also pull node:8-alpine/jsdep image from quay.io

xoen:
We missed this `BASE` image in previous commit. 

Again this is to avoid hitting the newly introduced DockerHub pulls rate limit and we're pulling the same image but from our (public) quay.io repository.


andylightfoot (approved):
LGTM

PR #857, opened 2020-11-17
Doc: Note that we should use Helm v2

xoen:
### What
A new joiner was getting the following error while running the tests using Helm 3.x:

```
FAILED tests/api/models/test_tool.py::test_tool_app_version[chart-with-appVersion] - controlpanel.api.helm.HelmError: Error while opening/parsing helm repository cache: 'reposi...
```

So this is to add a note/warning that you should use Helm 2.x when setting your local environment.

### DIFF
--- a/doc/running.md
+++ b/doc/running.md
@@ -280,6 +280,8 @@ kubectl cluster-info
 Install Helm (the K8s package manager) by following
 [these instructions for your OS](https://helm.sh/docs/intro/install/).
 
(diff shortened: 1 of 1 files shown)


ntoll (approved):
Great use of the digits `2` and `3` to disambiguate `helm` versions. :stuck_out_tongue_closed_eyes: :shipit:

xoen:
Thanks @ntoll - I tend to only use digit `0` and `1` but my digit `2` and `3` are not too bad neither 😬

ntoll:
I hear there are even more digits available (coming at some future date). :-P

PR #858, opened 2020-11-26
Improve the usability flow of home-directory reset

ladyrassilon:
## What
This changes the usability of the reset home-directory reset flow by requiring a second explicit confirmation of the reset process.

The only change is additional input tag classes and a confirmation message.

## How to review
1. Go to analytical tools (/tools/) on the control panel
2. Click on reset home directory
3. Click the confirmation tick box
4. Click the reset button
5. You will get a final confirmation message with a choice to continue or cancel.
6. Click cancel will not do anything
7. Repeat the previous steps and ensure that "Ok" triggers the reset.

Satisfies the requirement of https://trello.com/c/TJGXpIXj/859-the-home-reset-button-should-include-a-confirmation-before-actually-resetting

### DIFF
--- a/controlpanel/frontend/jinja2/reset.html
+++ b/controlpanel/frontend/jinja2/reset.html
@@ -61,7 +61,10 @@
               </div>
             </div>
           </fieldset>
-          <input type="submit" class="govuk-button govuk-button--warning" value="Reset"/>
+          <input type="submit"
+                 class="govuk-button govuk-button--warning cpanel-button--destructive js-confirm"
+                 data-confirm-message="I understand that my conda environment will be irretrievably reset"
+                 value="Reset"/>
           </form>
         </div>
     </div>


ntoll (commented):
When I run this code, and I have NOT checked the confirmation box, the alert appears FIRST and then the form validation feedback afterwards. Surely this should be the other way around? Actually, perhaps the users will get "confirmation fatigue" too..? :thinking: 

In any case, this PR proves that @ladyrassilon certainly knows how to get code into the project now...! Hurrah..! :+1:

controlpanel/frontend/jinja2/reset.html
```diff
@@ -61,7 +61,10 @@ <h1 class="govuk-heading-l">Home Directory Reset</h1>
               </div>
             </div>
           </fieldset>
-          <input type="submit" class="govuk-button govuk-button--warning" value="Reset"/>
+          <input type="submit"
+                 class="govuk-button govuk-button--warning cpanel-button--destructive js-confirm"
+                 data-confirm-message="I understand that my conda environment will be irretrievably reset"
```
xoen, line 66: I would probably not mention conda directly (~don't think we do in the rest of the page~ actually we mention as an example of the things that get reset) as we're keeping this "reset your home" vague by design.

Bear in mind the [`user-home-reset` helm chart/job](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/main/charts/reset-user-home/templates/job.yaml#L21-L27) also wipe other things, e.g. some RStudio stuff - and we may also expand that list as well if new and exciting problems arise in the future.

Maybe tweak to something like:

> I understand that my user home will be reset

Don't know if the ticket mentioned any copy/wording or @ntoll has opinions on this.

xoen:
@ntoll 

> Actually, perhaps the users will get "confirmation fatigue" too..? 🤔

Yeah, I was wondering that as well, after all this page is already a confirmation page.

I don't mind either way, this looks good to me 👍

xoen (approved):
*Maybe* the wording could be tweaked to be a bit more vague.

Also good point from Nicholas about "confirmation fatigue" as this confirmation dialog is already in a confirmation page so don't know...

But the code looks good to me 👍

PR #859, opened 2020-12-01
Ensure all JavaScript alerts on destructive buttons are configured correctly with the expected CSS classes.

ntoll:
## What

Some of the confirmation dialog alerts were not firing when they should have been. This was due to the wrong css class and data attribute set in the affected `button` tag. It meant admins could delete a user without a confirmation check (dangerous, due to accidental clicking).

Now the updated buttons work like this:

are_you_sure

## How to review

1. Start the app locally.
2. Log in as an admin user.
3. Visit the user or webapp admin functionality and click on an individual user/app instance.
4. Click on the delete button: an alert will magically appear..!

This PR has no ticket since it's an example of code gardening :rose: (small steps, carefully taken, as we go about our other "scheduled" coding tasks). Thanks to @davidread for pointing out the problem.

### DIFF
--- a/controlpanel/frontend/jinja2/user-detail.html
+++ b/controlpanel/frontend/jinja2/user-detail.html
@@ -99,8 +99,8 @@
 <section class="cpanel-section">
   <form action="{{ url("delete-user", kwargs={ "pk": user.auth0_id }) }}" method="post">
     {{ csrf_input }}
-    <button class="govuk-button cpanel-button--destructive confirm"
-            data-confirmation-prompt="Are you sure you want to delete this user?">
+    <button class="govuk-button cpanel-button--destructive js-confirm"
+            data-confirm-message="Are you sure you want to delete this user?">
       Delete user
     </button>
   </form>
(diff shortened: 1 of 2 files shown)


controlpanel/frontend/jinja2/user-detail.html
```diff
@@ -99,8 +99,8 @@ <h2 class="govuk-heading-m">{{ pronoun }} webapp data sources</h2>
 <section class="cpanel-section">
   <form action="{{ url("delete-user", kwargs={ "pk": user.auth0_id }) }}" method="post">
     {{ csrf_input }}
-    <button class="govuk-button cpanel-button--destructive confirm"
-            data-confirmation-prompt="Are you sure you want to delete this user?">
+    <button class="govuk-button cpanel-button--destructive js-confirm"
```
ladyrassilon, line 102: On the other button we made sure it was input submits rather than button...

Is it better to convert these two?

ladyrassilon (approved):
Otherwise tested and LGTM

controlpanel/frontend/jinja2/user-detail.html
ntoll: Not sure I follow.

davidread (approved):
Excellent

PR #860, opened 2020-12-01
Ensure buttons marked cpanel-button--destructive have confirmation alerts

ntoll:
## What

Following discussion with @davidread on Slack, we decided that the scope of UI elements that needed confirmation alerts should be widened to buttons associated with the `cpanel-button--destructive` class. In addition, those buttons that had mistakenly been mislabelled with the `confirm` CSS class have been updated to `js-confirm`.

In #859 I only updated buttons that had a confirmation message associated with them (indicating such functionality was missing). It turns out that if a button is labelled for confirmation no message is required since it defaults to "Are you sure..?". 

## How to review

1. Do you see an alert when you click on the associated buttons..?

This is an ad-hoc code gardening PR for cleaning up and mowing the proverbial lawn of JavaScript. :sunflower: :seedling:

### DIFF
--- a/controlpanel/frontend/jinja2/datasource-access-update.html
+++ b/controlpanel/frontend/jinja2/datasource-access-update.html
@@ -54,7 +54,7 @@
 <section class="cpanel-section">
   <form method="post" action="{{ revoke_url }}">
     {{ csrf_input }}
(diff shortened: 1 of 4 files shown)


xoen (approved):
LGTM 👍

PR #861, opened 2020-12-03
Add login-fail page and recovery link to control panel.

ntoll:
## What

This PR fixes [this Trello ticket](https://trello.com/c/2X2Bb3UM/864-cp-login-errors-are-displayed-instead-of-redirect-loop) and mitigates the misconfiguration described in [this Trello ticket](https://trello.com/c/0i7JurE1/863-allow-a-user-to-cp-rstudio-after-having-already-logged-into-an-app) (work on which is ongoing with Auth0).

I have made the following changes:

* The `LOGIN_REDIRECT_URL_FAILURE` setting in the configuration of the application is changed from `/` (causing an infinite loop) to a new endpoint `/login-fail/`.
* The new endpoint (see screenie below) includes a link that clears the Auth0 cookie for the user's session and redirects them back to the control panel application in order to log in again from a clean state.
* I've also made a slight modification to `frontend/consumers.py` to check if a `user` is logged in before using the `auth0_id` attribute. This was causing an unhandled exception in the background worker when the (not-logged-in) user was redirected to the `login-fail` page.

fail

It's also important to note that for this change to work, we need to update the Auth0 tenant settings so that the `/oidc/logout/` endpoint is in the "Allowed Logout URLs" setting (see screenie below).

tenant_logout

## How to review

1. We should deploy this branch to `dev`.
2. We should have a working RShiny app running in `dev`.
3. Work through the steps to recreate the "wrong cookie" bug when a user tries to log into the control panel after having been logged into an application.
4. The user should see the expected login challenge rather than an error complaining about cookies and recursive redirects.

This is the API endpoint we're using to achieve the clearing of the Auth0 cookie: https://auth0.com/docs/api/authentication#logout (documentation for which can be [found here](https://auth0.com/docs/logout/log-users-out-of-auth0)). We use [this feature of the API](https://auth0.com/docs/logout/redirect-users-after-logout) to redirect users back to our application.

### DIFF
--- a/controlpanel/frontend/consumers.py
+++ b/controlpanel/frontend/consumers.py
@@ -100,9 +100,11 @@ class SSEConsumer(PatchedAsyncHttpConsumer):
         """
         self.streaming = False
 
-        # leave the group
-        group = sanitize_dns_label(self.scope.get("user").auth0_id)
-        await self.channel_layer.group_discard(group, self.channel_name)
+        # leave the group if user is logged in.
+        user = self.scope.get("user")
+        if user.is_authenticated:
+            group = sanitize_dns_label(user.auth0_id)
+            await self.channel_layer.group_discard(group, self.channel_name)
 
 
 class BackgroundTaskConsumer(SyncConsumer):
--- a/controlpanel/frontend/jinja2/login-fail.html
+++ b/controlpanel/frontend/jinja2/login-fail.html
@@ -1,0 +1,17 @@
+{% extends "base.html" %}
+
+{% set page_title = "Login Failure" %}
+
+{% block content %}
+  <div class="govuk-grid-row" id="reset-container">
+   <div class="govuk-grid-column-two-thirds">
+
+
+    <h2 class="govuk-heading-l">Login Failure</h2>
+
+    <p>Something went wrong when you tried to authenticate.</p>
+
+    <p>Please <a href="https://{{ environment }}-analytics-moj.eu.auth0.com/v2/logout?returnTo=https%3A%2F%2Fcontrolpanel.services.{{ environment }}.mojanalytics.xyz%2Foidc%2Flogout%2F">try again</a>.</p>
+   </div>
+  </div>
+{% endblock %}
--- a/controlpanel/frontend/jinja2/login-fail.html
+++ b/controlpanel/frontend/jinja2/login-fail.html
@@ -11,7 +11,9 @@
 
     <p>Something went wrong when you tried to authenticate.</p>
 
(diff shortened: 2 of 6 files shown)


davidread:
In the case we're talking about, Auth0 gives us an error message "Access denied", because it thinks you are logged in as a different identity, for the app. And there may be other reasons that Auth0 login fails and it would be helpful to know what the problem is. Could you display the error description "Access denied" on that `/login-fail/` web page? Auth0 will redirect back to `https://controlpanel.services.alpha.mojanalytics.xyz/oidc/callback/?error=access_denied&error_description=Access%20denied.&state=sIhnc...` which I think with this PR should redirect to `/login-fail/`. So maybe it's possible to get hold of from that URL.

Secondly, I'm not sure about the UX of having a link with text "try again". It doesn't just try the same thing again, it logs you out before doing that. So this might be confusing. How about offering a simple "Log-out" link/button, suggesting that might help?

ntoll:
@davidread UX-y wise, I've revised the message on the landing page for login failure to, "Please reset your session, and try again." (where "reset your session" is the anchor tag to click to reset the session).

As for the error message: I'm on it and will be adding some code we can exercise in `dev` to hopefully grab the actual error message and reason for the error. We should be able to display something human-centric and useful for debugging purposes. In any case, this is ongoing work/investigation... but I agree with you... if we can use such context, we should.

ntoll:
Hi, so I've spent the morning looking into how we can tell what sort of error message we get from Auth0 in the event of a failed login.

In essence, as @davidread points out, Auth0 responds with context via the query string of the callback URL, with the following keys/values:

```
>>> from urllib.parse import urlparse, parse_qs
>>> my_url = urlparse(request.get_raw_uri())
>>> parse_qs(my_url.query)
{
  'error': ['login_required'],
  'error_description': ['Login required'],
  'state': ['vDbisWUfOOBarKtvxibAZKjtmjUEzLMU']
}
``` 

This is available to us in the `get` method of our `StateMismatchHandler` class (inherited from Mozilla's OIDC view class). Ultimately, [this bit of Mozilla's code](https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/views.py#L59) is run and we're redirected to `settings.LOGIN_REDIRECT_URL_FAILURE` (i.e. the new endpoint created by this PR). Because this is an "automatic" redirect from a third party module, we lose any of the Auth0 error information shown above. By the time our `LoginFail` view gets to handle the request, this context is gone. 

I'm not sure what more we could do (apart from log a failed login attempt, but this would, more than likely, just fill our logs with yet more noise).

I suggest the PR is "good enough" for the time being and, once we have a working RShiny app in `dev`, ready for testing. (NB: I've tested it "locally" but only by forcing a simulated error - I've been unable to test it with a recreation of the original "wrong cookie" problem... which is why we need to check things work / can be recreated in `dev`).

xoen (approved):
LGTM 👍

PR #863, opened 2020-12-10
Update the fallback default Elastic index for user app logs.

ntoll:
## What

Now that Elastic is logging to a new index, the default fallback value should be changed in the Control Panel's settings, to reflect this. In addition, the `ELASTICSEARCH_INDEX_APPLOGS` environment variable should be revised (where???) to reflect this too.

## How to review

1. Is the index in the diff the correct one for Elastic.
2. ???
3. Profit!

This is some quick code gardening to make sure we don't grow weeds in our code.


tomskelley-gjs (approved):
LGTM

PR #865, opened 2021-01-05
Ensure chart name validation and chart context values re-used.

ntoll:
## What

This PR addresses the bug spotted by Aldo, where the helm chart for a new release does not contain the JSON context values.

It introduces the following changes:

* Ensure that only *valid* chart names can be used in the release form.
* Update the UI to remind the user what these valid chart names are.
* When a new release is created, the JSON values from the release for the most recent live version of the referenced helm chart are copied over to the new release.
* A unit test to ensure the JSON values are copied as expected.

Here's a GIF-y of the changes in the UI:

create_release

## How to review

1. Create a new release with a valid chart name, as per the GIF-y above.
2. Open the Django shell: `./manage.py shell`
3. Import the model for releases: `from controlpanel.api.models import Tool`
4. Get the most recent release relating to your chart name (i.e. the one you just created): `release = Tool.objects.filter(chart_name="jupyter-lab").order_by("-created").first()`
5. Ensure you see JSON data in the values field: `release.value`
6. Exit the shell: `exit()`.

See Jira ticket: https://dsdmoj.atlassian.net/browse/ANPL-233

### DIFF
--- a/controlpanel/frontend/forms.py
+++ b/controlpanel/frontend/forms.py
@@ -285,6 +285,26 @@ class ToolReleaseForm(forms.ModelForm):
         else:
             return []
 
+    def clean_chart_name(self):
+        """
+        Ensures that the helm chart name entered by the user is a valid one.
+
+        Why not use a "choice" argument in the Tool model's class? It would
+        result in a new Django migration for the database to enforce this (and
+        we'd have to keep adding new migrations to the database every time we
+        created a new helm chart for a new tool). Furthermore, the HTML select
+        field we'd use in the form isn't supported in the macros we use.
+
+        Hence the path of least resistance with this custom form validation.
+        """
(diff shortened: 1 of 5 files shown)


damacus:
How come we didn't go for a drop down rather than a validation(?) field?

ntoll:
@damacus see the code comment here: https://github.com/ministryofjustice/analytics-platform-control-panel/pull/865/files#diff-cc6c2afafa823a969f5252777c1ec6319f4e0aad505bcda69e4f09cc97168bf8R292 ;-) Honestly, if I could, I would, but our frontend macros are a hold over from a node app and incomplete. TL;DR while we use Django, we don't use some of what Django offers (such as cool form rendering).

damacus (approved):

PR #1215, opened 2023-10-23
Initial implementation of archiving S3 folder contents

michaeljcollinsuk:
Use tasks so that the contents are moved asynchronously. A tasks is fired for every individual object in order to utilise multiple workers to improve performance.
Adds methods to cluster, aws files to handle retrieving and archiving objects.

closes ministryofjustice/data-platform#1887


ymao2 (dismissed):
LGTM :), just some minor comment for removing `print`, instead, using `logging`

ymao2 (approved):
LGTM 👍

PR #1223, opened 2023-11-09
Generate python package upgrade

ymao2:
This PR is for upgrading the python packages raised by the dependantbot,  only one cannot be upgrade , it is channel-redis,  the new version will cause an exception 
```
TypeError: __init__() got an unexpected keyword argument 'timeout'
```
I was thinking it is due to the lower version of `redis` python package, but checked we are using the latest one (only checked my local)


michaeljcollinsuk (approved):
LGTM

PR #1228, opened 2023-12-07
Simplify the register app form

michaeljcollinsuk:
Relates to https://github.com/ministryofjustice/data-platform/issues/2633

## :memo: Summary

Post-migration app migration we want to make registering apps self service, therefore the app registration form has been simplified and opened to all logged in users.

This PR:
- Reduces the app registration form to a repo url and datasource. 
- Improves validation and error handling when registering an app
- Remove unnecessary JS when looking up a repo, and limiting repos to the `ministryofjustice` org
- Only an AWS role is created for the app, other settings and Auth0 clients will be created via the manage app page
- Create cloud platform assume role policy when creating app AWS Role

## :mag: What should the reviewer concentrate on?
- Changes to the app registration flow

## :technologist: How should the reviewer test these changes?
- Run tests with `pytest tests`
- Run app locally and attempt to register a new app. Example repo that will work: https://github.com/ministryofjustice/data-platform-new-app-reg-test
- Try to register an invalid repo - should receive error message

## :books: Documentation status

- [ ] Documentation will be added in the future within user guidance - [see new app registration epic for related issues](https://github.com/ministryofjustice/data-platform/labels/%5B%F0%9F%93%9D%20%5D%20Register%20a%20new%20app%20%28Epic%20%231700%29)

PR #1240, opened 2024-01-31
Bugfix/allow app registration

michaeljcollinsuk:
Follow on changes to https://github.com/ministryofjustice/analytics-platform-control-panel/pull/1228

Fixes some bugs found when testing:

- Display the Register App button to all users
- Update the list of datasources that can be linked to an app, so that only datasources that the user has admin access to are displayed
- Update logic to show the button to create webapp datasources

TODO
In a follow up PR need to add in some extra tests to the new methods on the forms


mitchdawson1982 (dismissed):
LGTM

mitchdawson1982 (approved):
🚀

PR #1249, opened 2024-02-14
Update dependencies and Python version to 3.12

jamesstottmoj:
Merging this PR will update the control panel to run on Python 3.12

## :memo: Summary
This PR resolves 

This PR bumps several dependencies and fixes issues in the tests so that the control panel can run against Python 3.12

The changes in this PR are needed because it updates dependencies and core Python version

- Check side effects, if any

- Run locally. There should be no noticeable difference between main and this branch


requirements.txt
```diff
@@ -1,38 +1,38 @@
...
+django-structlog==7.1.0
 djangorestframework==3.14.0
 djproxy==2.3.6
 elasticsearch-dsl==7.4.1
-gunicorn==20.1.0
+gunicorn==21.2.0
 Jinja2==3.1.3
-kubernetes==25.3.0
-MarkupSafe==2.1.3
+kubernetes==29.0.0
```
michaeljcollinsuk, line 22: I'm a little bit unsure about this jump - there was some discussion with the rest of the team about whether we can upgrade when [dependabot suggested it](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/1218). Probably not something that we would catch running locally

michaeljcollinsuk:
Really good start on this. The [Dockerfile](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/Dockerfile#L12) will need updating as part of this so that it builds an image using python 3.11, which may go some way to resolving the current fail in CI. I wonder what else we may also need to update in the Dockerfile, there may be implications for helm version it installs, something we can discuss that with some of the other guys in the team.

tests/frontend/test_forms.py
```diff
@@ -411,6 +410,10 @@ def test_update_app_with_custom_connection():
         assert "auth0_nomis_auth0_conn_name" in f.errors
 
 
+def test_pass():
```
michaeljcollinsuk, line 413: Can be deleted
```suggestion
```

tests/frontend/test_forms.py
```diff
@@ -411,6 +410,10 @@ def test_update_app_with_custom_connection():
...
+    assert True
+
+
```
michaeljcollinsuk, line 416: Can be deleted
```suggestion
```

jamesstottmoj: Removed

michaeljcollinsuk:
I think the python version mentioned in `running.md` needs to be updated - not sure if the minimum version will now be 3.12, but I presume that 3.8 wont work with this new set of deps?

tests/api/models/test_ip_allowlist.py
```diff
@@ -44,10 +44,10 @@ def test_ip_allowlist_without_app_save_does_not_update_aws_secrets_manager(
 
 # def test_ip_allowlist_delete_updates_aws_secrets_manager(update_aws_secrets_manager):
```
michaeljcollinsuk, line 45: Might as well delete this test since its commented out?

michaeljcollinsuk (changes requested):
Just one minor change needed in `running.md` about the python version to use, but otherwise great work!

tests/api/models/test_ip_allowlist.py
jamesstottmoj: Removed

jamesstottmoj:
Running.md updated

michaeljcollinsuk (dismissed):
🚀

michaeljcollinsuk (approved):
LGTM

michaeljcollinsuk:
Closing as replaced by https://github.com/ministryofjustice/analytics-platform-control-panel/pull/1285

PR #1251, opened 2024-02-16
Revert changes to hide the 'Manage Customers' button

michaeljcollinsuk:
Hiding the "Manage Customers" button broke formatting on the app list page and could lead to confusion. This only became apparent following testing on the dev site. Instead I have improved the messaging displayed to users if they try to manage customers without either creating an Auth0 client or enabling authentication.


jamesstottmoj (approved):
Approved

PR #1257, opened 2024-02-20
Bugfix/namespace length

michaeljcollinsuk:
## :memo: Summary
This PR resolves a bug where our expected namespace is too long to be valid for Cloud Platform

Instead of assuming the namespace to be in the format of `data-platform-app-<repo-name>`, the user will enter their chosen namespace. This is then used when we create the trust policy for the App iam role.

The namespace will be stored on the model. For all existing apps, there is a data migration to add these.

- Older app roles will have the `data-platform-app` prefix, others will not

- Change to the trust policy
- Migrations order

- Will be tested in dev after merge


controlpanel/frontend/forms.py
```diff
@@ -204,6 +206,18 @@ def clean_repo_url(self):
 
         return repo_url
 
+    def clean_namespace(self):
+        """
+        Removes the env suffix is the user included it
```
jamesstottmoj, line 211: Super minor but I presume you wanted to say "if the user included it"?

jamesstottmoj (approved):
Approved

PR #1274, opened 2024-03-18
Update GitHub actions workflow to trigger build/deploy only when releases

michaeljcollinsuk:
## :memo: Summary
This PR resolves a bug where the actions workflow is triggered 3 times when a new release is created

- There is potential that the change does not work, and the workflow does not trigger when the release is created. If this occurs, we can revert the commit

- Actions file chang

- Create a new release, check it triggers the build and test workflow


jamesstottmoj (approved):
LGTM

PR #1281, opened 2024-04-03
Disable justice auth capture flow in dev for non-superusers

michaeljcollinsuk:
Disable the Justice identity via feature flag. This should mean that only superusers will see the page to capture their Justice identity.

Although this is only deployed to dev, there are some normal users currently using dev - and when they see this page it may confuse them, so better to disable for now.


Gary-H9 (approved):
LGTM

PR #1282, opened 2024-04-05
Update justice auth flow to use client_secret and prompt login

michaeljcollinsuk:
Use the client_secret created in Azure app registration, for better security. This is set as an env var, and will be read from a Kubernetes secret, which will be added manually elsewhere. Also update azure client settings to always prompt the user to complete the login process with their justice identity. This will avoid confusion if a user is already logged in to their Justice identity.


jamesstottmoj (approved):
Tested locally by adding new env var and running through login flow. Approved

PR #1283, opened 2024-04-09
ensure redirect to index if no justice id regardless of route hit

jamesstottmoj:
## :memo: Summary
This PR ensures that a user is redirected to the index page if they have not authorised with their justice identity. 

The changes in this PR are needed because if the user accesses the control panel through any other route, e.g control-panel/tools, they will not see the page that allows them to log in with their justice identity

- Check side effects, if any

- Manually test by logging out, access the control panel by any page that isn't the root, then check that you are redirected to the page that asks for your justice credentials


controlpanel/frontend/views/auth.py
```diff
@@ -62,3 +63,17 @@ def update_user(self, token):
         email = token["userinfo"]["email"]
         self.request.user.justice_email = email
         self.request.user.save()
+
+
+class EntraOIDCAuthenticationCallbackView(OIDCAuthenticationCallbackView):
```
michaeljcollinsuk, line 68: This does the job, and works. However the project already overrides the callback view to us this https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/oidc.py#L67 

Then there is a setting that points to that class in the oidc package https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/settings/common.py#L164

So we dont need to override the URL, and also dont want to lose that additional functionality that has already been added. But we can still use your logic for the redirect on that view - if you look in the source code, the `OIDCAuthenticationCallbackView` has a `success_url` method. So I would suggest we add that method to the `StateMismatchHandler` view (could rename the view if you wanted), to check for the email and return the index url if we havent captured it, or call `super()` if we already have it.

michaeljcollinsuk (changes requested):

controlpanel/frontend/views/auth.py
michaeljcollinsuk: Would also be good to add a unit test - should be easier if all the logic is in the `success_url` method

tests/test_oidc.py
```diff
@@ -0,0 +1,21 @@
...
+
+@pytest.mark.parametrize(
+    "email, success_url",
+    [
+        ("", "/"),
+        ("example@justice.gov.uk", "/tools/"),
+    ],
+)
+def test_success_url(users, email, success_url):
+    request = Mock()
```
michaeljcollinsuk, line 14: For future reference, there is a fixture you can use to create a request object https://pytest-django.readthedocs.io/en/latest/helpers.html#rf-requestfactory

michaeljcollinsuk (approved):
Nice work

PR #1291, opened 2024-04-17
Language and UI refactor

jamesstottmoj:
## :memo: Summary
This PR resolves ministryofjustice/data-platform#3380

This PR rewords some text in the control panel to fit a more formal tone
Emojis are removed from the navbar
Support slack links have been changed to the github support request page
A broken Grafana link has been removed

The changes in this PR are needed to improve the readability and user experience of the control panel.

- Feedback on wording changes
- Check side effects, if any 

### DIFF
--- a/controlpanel/frontend/jinja2/base.html
+++ b/controlpanel/frontend/jinja2/base.html
@@ -9,6 +9,7 @@
 {% set home_url = url("index") %}
 {% set signout_url = url("oidc_logout") %}
 
+
 {% block page_title %}
   {{ page_title }} | {{ service_name }}
 {% endblock %}
@@ -83,7 +84,6 @@
         "text": "Home",
         "href": home_url,
         "active": page_name == "home",
-        "admin": request.user.is_superuser,
       },
       {
         "hide": not request.user.is_staff,
@@ -116,7 +116,6 @@
         "text": "Groups",
         "href": url("list-policies"),
         "active": page_name == "groups",
-        "admin": request.user.is_superuser,
       },
       {
         "text": "Training",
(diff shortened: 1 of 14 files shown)


controlpanel/frontend/jinja2/base.html
```diff
@@ -83,7 +84,6 @@
         "text": "Home",
         "href": home_url,
         "active": page_name == "home",
-        "admin": request.user.is_superuser,
```
michaeljcollinsuk, line 86: I think as part of removing this we might as well also remove the associated bit of code in the template that is rendered. Looks to be in `controlpanel/frontend/static/components/navbar/macro.html`:

```
{% if item.admin %}&#128295;{% endif %}
```

Unless you wanted to leave this in for some reason?

controlpanel/frontend/jinja2/reset.html
```diff
@@ -22,12 +22,12 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
         state (for example, a broken conda environment or Jupyter and RStudio won't
         start).</p>
 
-        <p>Reset your home directory to recreate your conda environment and shell
-        settings to a clean state. This feature automates the
+        <p>Reset your home directory to restore your conda environment and shell
+        settings to their original state. This feature automates the
         <a href="https://github.com/ministryofjustice/analytics-platform/wiki/Reset-your-R-Studio-environment">process described here</a>.</p>
 
         <div class="govuk-warning-text">
```
michaeljcollinsuk, line 33: I think we can remove this warning, as its not accurate. As files are stored in EFS then resetting their home directory wont delete anything - the only thing that could be lost is if a file hasnt been saved, so either we remove it entirely or update to something along the lines of "Make sure you have saved any open files before"

michaeljcollinsuk (changes requested):

michaeljcollinsuk (approved):
LGTM

PR #1297, opened 2024-05-20
Feature/lake formation poc

michaeljcollinsuk:
## :memo: Summary
This PR resolves ministryofjustice/analytical-platform#4367 

This PR adds the ability for a user to manage access to tables via Lake Formation permissions. This is using a named resource rather than tags and is explicitly for usage with DPR.

The changes in this PR are needed because it's a stepping stone to manage databases via the control panel in AP

- Check code quality
- Check side effects, if any

- Add an existing database as an env var -`DPR_DATABASE_NAME`
- Sign into the control panel locally as a super user
- See that there should be a new Databases link in the navbar
- Navigate to manage your user and check the database management toggle, then uncheck super user access
- Check if you can still see the Databases link
- Go to a table that is managed in lake formation and see if you can grant and revoke permissions to a user for access to the table
- Go to Athena and check access as well as ability to query data when permissions have been granted


michaeljcollinsuk (commented):
LGTM

jamesstottmoj (approved):
OBO Michael Collins

Approved

PR #1301, opened 2024-06-06
Add management command to cleanup QuickSight users

michaeljcollinsuk:
Management command used to cleanup quicksight users in the data-prod account. Requires a CSV file of users to be removed.

This command has already been used to clean up users, but added as a management command in case it is needed again in future.


jamesstottmoj (approved):
LGTM

PR #1314, opened 2024-06-19
App bedrock access

jamesstottmoj:
## :memo: Summary
This PR resolves ministryofjustice/analytical-platform#4311

This PR allows bedrock policy to be applied to an app role

- Check side effects, if any

- Pull down branch and run locally
- Go to app management page and check that bedrock checkbox exists
- Toggle checkbox and check that bedrock policy is applied/removed appropriately


controlpanel/frontend/views/app.py
```diff
@@ -295,6 +295,21 @@ def form_invalid(self, form):
         raise Exception(form.errors)
 
 
+class EnableBedrockApp(OIDCLoginRequiredMixin, PermissionRequiredMixin, UpdateView):
```
michaeljcollinsuk, line 298: This is largely a copy and paste of the `EnableBedrockUser` view right? So maybe to keep DRY you could create a `BedrockAccessMixin` to have the shared methods, so that we can reduce the views themselves to:
```
class EnableBedrockApp(BedrockAccessMixin, UpdateView):
    model = App


class EnableBedrockApp(BedrockAccessMixin, UpdateView):
    model = User
```

To do this you would need to [add `get_absolute_url` methods](https://docs.djangoproject.com/en/5.0/ref/models/instances/#get-absolute-url) to the `User` and `App` models, and then those methods will be used by the `UpdateView` [when `get_success_url` is called](https://ccbv.co.uk/projects/Django/5.0/django.views.generic.edit/UpdateView/#get_success_url). Let me know if you have any questions or want to pair on this.

controlpanel/api/models/app.py
```diff
@@ -305,6 +306,9 @@ def get_auth0_group_list(self):
     def auth_settings(self):
         return (self.app_conf or {}).get(self.KEY_WORD_FOR_AUTH_SETTINGS, {})
 
+    def get_absolute_url(self):
+        return reverse("manage-app", kwargs={"pk": self.id})
```
michaeljcollinsuk, line 310:
```suggestion
        return reverse("manage-app", kwargs={"pk": self.pk})
```
Django sets a property to return the primary key which is useful in methods like this https://docs.djangoproject.com/en/5.0/ref/models/instances/#the-pk-property

controlpanel/api/models/user.py
```diff
@@ -192,5 +193,8 @@ def bulk_migration_update(cls, usernames, new_state):
                 ],
             )
 
+    def get_absolute_url(self):
+        return reverse("manage-user", kwargs={"pk": self.auth0_id})
```
michaeljcollinsuk, line 197:
```suggestion
        return reverse("manage-user", kwargs={"pk": self.pk})
```

michaeljcollinsuk (changes requested):

michaeljcollinsuk (approved):
🥳

PR #1325, opened 2024-07-31
Update grant/revoke database access

michaeljcollinsuk:
Updates granting/revoking of database tables

## :memo: Summary
This PR relates to https://github.com/ministryofjustice/analytical-platform/issues/4367

After initial testing, it became apparent that when granting permissions on shared tables, the boto calls need to occur on the source table/database. This PR updates the views to get details of the database/tables, and if they point to a 'target', use those details when managing permissions. Part of this means checking what region the shared tables live, and if this is different to the default AWS region, make sure to reinitialise the `glue` client for the correct region.

Also from initial testing we have learned that (in the data-prod account) the resource link that will be used to reference the shared tables will be visible to users via the "standard-database-access" permissions. Therefore there is no need to grant "DESCRIBE" access to the resource link. This reduces the number of boto3 permission updates required, so granting/revoking access now only occurs on the source table.

- Most of the logic has been added to a mixin that is shared between the views that need to lookup tables
- Check general refactoring to make sure that nothing has been missed.
- Further unit tests should be added when possible, but for now I have updated only the existing tests

- When running locally, run against the dev account
- To mimic having standard database access, you will need to copy the permissions from the `standard-database-access` policy in data prod, and assign them to your `dev_user_` role in the dev account
- In `analytical-platform-development` I have setup a database and tables to mimic "shared" tables. These live in eu-west-2 (the same region that DPR data will be shared to). The database is called `michael_dpr_testing` and has 3 tables to test with.
- I have then created a resource link to the database in `eu-west-1`. This is called `michael_dpr_testing_resource_link`. Use this as your env var `DPR_DATABASE_NAME`
- If you then run the server and visit http://localhost:8000/databases/ you should see this database, its tables, and test granting access
- NOTE: You should not be able to grant access to the the table `dpr_test_2` as I have purposely not registered within Lake Formation. I did this by assigning the IAMAllowedAllPrinciples` permission to it, so in Athena you will be able to see the table but not query it
- After granting some permissions to your user, log in to AWS athena as your dev user, (via the link on the My Tools page) and try to query the tables

### DIFF
--- a/controlpanel/frontend/jinja2/table-detail.html
+++ b/controlpanel/frontend/jinja2/table-detail.html
@@ -39,7 +39,7 @@
         <tr class="govuk-table__row">
           <td class="govuk-table__cell">{{ permission.user }}</td>
           <td class="govuk-table__cell align-right no-wrap">
-            <form action="{{ url("revoke-table-permissions", kwargs={ "dbname": dbname, "tablename": table.Name, "user": permission.user }) }}" method="post">
+            <form action="{{ url("revoke-table-permissions", kwargs={ "dbname": dbname, "tablename": tablename, "user": permission.user }) }}" method="post">
               {{ csrf_input }}
               <button class="govuk-button cpanel-button--destructive js-confirm"
                       data-confirm-message="Are you sure you want to revoke access for user?">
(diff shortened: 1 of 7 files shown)


jamesstottmoj (approved):
Pulled down and tested locally. Seems to work ok. Approved

PR #1344, opened 2024-10-03
Added checkbox to give an app textract permissions

jamesstottmoj:
## :memo: Summary
This PR resolves [this](https://github.com/ministryofjustice/analytical-platform/issues/4877) issue

This PR adds a checkbox that will add a textract policy to a Webapp

The changes in this PR are needed as it was asked for by several users

- Check side effects, if any


controlpanel/frontend/mixins.py
```diff
@@ -20,3 +20,19 @@ def form_valid(self, form):
     def get_success_url(self):
         messages.success(self.request, "Successfully updated bedrock status")
         return super().get_success_url()
+
+
+class TextractAccessMixin(OIDCLoginRequiredMixin, PermissionRequiredMixin):
```
michaeljcollinsuk, line 25: I am wondering if it is worth having this as a mixin if it is only used in one place? The bedrock one gets used in two places (apps or users), is there any likelihood of textract being enabled for users? If not, I'd probably just keep all the logic on the actual view class.

Or alternatively, since this essentially does the same thing as the `BedrockAccessMixin` you could refactor to combine the two to a more general mixin that allows a policy to be attached, like a `AttachPolicyMixin` or something like that? You'd need to set the `fields`, `success_message` on the view, and dynamically get the method that gets called on the object. Could then be reused if we need to add any more policies that can be attached.

jamesstottmoj: I've updated the mixin now. See what you think of the implementation

controlpanel/frontend/mixins.py
```diff
@@ -6,33 +6,26 @@
...
-    http_method_names = ["post"]
-    permission_required = "api.add_superuser"
-
-    def form_valid(self, form):
-        self.object.set_textract_access()
-        return super().form_valid(form)
+            getattr(self.object, self.method_name)()
+            return super().form_valid(form)
+        except Exception as e:
+            messages.error(self.request, "An error occurred while updating the policy access level")
```
michaeljcollinsuk: Could be worth sending the error to Sentry using `capture_exception` to help us debug if something does go wrong

jamesstottmoj: I've updated the implementation of this to raise an AttributeError if the method doesn't exist prior to this.

michaeljcollinsuk (approved):
Nice work!

controlpanel/frontend/mixins.py
michaeljcollinsuk: This is better!

PR #1350, opened 2024-10-10
Embedding QS into Control Panel

jamesstottmoj:
## :memo: Summary
This PR resolves ...

This PR adds QS into the control panel, similar to the AP UI.

- Quicksight will now be in the control panel

- Check side effects, if any

- Pull down branch
- ensure you are a superuser locally
- Add the following to your .env file:
   - export QUICKSIGHT_ACCOUNT_ID="381491960855"
   - export QUICKSIGHT_ACCOUNT_REGION="eu-west-2"
   - export QUICKSIGHT_DOMAINS="http://localhost:8000"
   - export QUICKSIGHT_ASSUMED_ROLE="arn:aws:iam::381491960855:role/analytical-platform-control-panel"
- Run control panel locally
- Navigate to quicksight page
- See quicksight in all its glory


controlpanel/frontend/jinja2/quicksight.html
```diff
@@ -0,0 +1,67 @@
...
+    {% else %}
+    <div class="govuk-width-container">
+        <h2 class="govuk-heading-m">Something went wrong, try refreshing the page</h2>
+        <p class="govuk-body">If the problem persists, please contact the AP support team.</p>
+    </div>
+    {% endif %}
+{% endblock content %}
+
+{% block body_end %}
+    <script src="https://unpkg.com/amazon-quicksight-embedding-sdk@2.6.0/dist/quicksight-embedding-js-sdk.min.js"></script>
```
michaeljcollinsuk, line 22: I think this is fine to test with for now - but at some point I think its worth looking at moving the JS out of the HTML and into its own file with the other JS in `static/javascripts/modules`

controlpanel/settings/common.py
```diff
@@ -248,6 +249,19 @@
 # Whitelist values for the HTTP Host header, to prevent certain attacks
 ALLOWED_HOSTS = [host for host in os.environ.get("ALLOWED_HOSTS", "").split() if host]
 
+
+# set this before adding the IP address below
+# TODO We may be able to set this in terraform instead, we should check this
+QUICKSIGHT_DOMAINS = []
```
michaeljcollinsuk, line 255: I think this is going to cause problems in dev/prod, as the Control Panel helm chart sets `ALLOWED_HOSTS` to a [wildcard](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/dcf007bfbc36b26277295a34ceabbbf99137ca34/charts/cpanel/templates/env-configmap.yaml#L12) (which is not good in any case)

So left as it is, I'd expect Quicksight to fail to embed when deployed. Instead, we need to update the ALLOWED_HOSTS to explicitly include the the prod and dev domains: `controlpanel.services.analytical-platform.service.justice.gov.uk` and `controlpanel.services.dev.analytical-platform.service.justice.gov.uk`.

We have a few options for doing this, but I think the best one it to add `ALLOWED_HOSTS` as an env var in the flux repo. Not sure if this would then take precendence over the one set in the helm chart, but we could try it. Or delete the var in the helm chart, and publish a new version.

michaeljcollinsuk (approved):
LGTM

PR #1352, opened 2024-10-16
Check for a 404 error when creating a bucket

michaeljcollinsuk:
## :memo: Summary

This PR resolves [TODO add bug ticket now](https://github.com/ministryofjustice/analytical-platform/issues/5852)

If a 404 is returned, the bucket name is available and can be used. If another error is returned, it may mean that the bucket name is already in use, but we do not have access to it. In this case, we should not attempt to create a bucket with this name. See the note on: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_bucket.html

- Error handling

- See bug ticket


jamesstottmoj (approved):
LGTM

PR #1356, opened 2024-10-21
:wrench: Add `GetObjectTagging` to `READ_ACTIONS` 

Gary-H9:
Updating in relation to [this](https://github.com/ministryofjustice/data-platform-support/issues/904) Issue but this has occurred on numerous occasions in the last year.

Further work to be done to test/apply to all users.


michaeljcollinsuk (approved):
~Saw this resolved the support ticket~ still waiting to see if this resolves the support issue, but approving in advance. If we want this to apply to all users, I can run a shell command to do this. Otherwise, users IAM permissions will update next time they update access to any bucket. Personally I think the latter would be fine for now, as we haven't come across this often

PR #1357, opened 2024-10-21
Resized Quicksight UI height

jamesstottmoj:


julialawrence (approved):
Well...I was worried it would take too much time but... :D Thank you!

PR #1362, opened 2024-10-23
Change log calls to log.warn

jamesstottmoj:
Change to log.warn as log.info isn't printing to the logs


michaeljcollinsuk (changes requested):
I think a better way to fix this is to change the log level to `INFO` here https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/settings/common.py#L556

Then we wont miss any other `info` log outputs elsewhere in the code

michaeljcollinsuk (approved):

PR #1363, opened 2024-10-25
Added button on user page that will re-initialise a users kubernetes namespace

jamesstottmoj:
This PR gives a superuser the ability to re-initialise a user's kubernetes namespace. This is as a result of [this](https://github.com/ministryofjustice/data-platform-support/issues/913) support issue.

- Check side effects, if any

- Pull down this branch
- Delete your own namespace in the development cluster by running `aws-vault exec admin-dev-sso -- kubectl delete namespace <yournamespace>`
- Navigate to your user from the home page
- click the re-initialise user button
- If successful, check your namespace has been rebuilt by running `aws-vault exec admin-dev-sso -- kubectl get namespace <yournamespace>`

### DIFF
--- a/controlpanel/frontend/jinja2/user-detail.html
+++ b/controlpanel/frontend/jinja2/user-detail.html
@@ -13,6 +13,14 @@
 <span class="govuk-caption-xl">User</span>
 <h1 class="govuk-heading-xl">{{ page_title }}</h1>
 
+<h2 class="govuk-heading-m">
+  Date Registered
+</h2>
+
+<p class="govuk-body">
+  {{ user.date_joined.strftime('%d-%m-%Y') }}
+</p>
+
 {% if unused %}
 <div class="govuk-grid-row">
   <div class="govuk-column-two-thirds">
@@ -53,7 +61,7 @@
           "value": "True",
           "text": "Super Admin",
           "checked": user.is_superuser
-        },
+        }
       ]
     }) }}
     <button class="govuk-button govuk-button--secondary">
--- a/controlpanel/frontend/jinja2/user-detail.html
+++ b/controlpanel/frontend/jinja2/user-detail.html
@@ -153,23 +153,21 @@
 </section>
 {% endif %}
 
-{% if request.user.has_perm('api.reset_mfa') %}
+{% if request.user.has_perm('api.add_superuser') %}
 <section class="cpanel-section">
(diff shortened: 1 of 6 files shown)


controlpanel/frontend/views/user.py
```diff
@@ -118,6 +119,23 @@ def post(self, request, *args, **kwargs):
...
+class ReinitialiseUser(OIDCLoginRequiredMixin, PermissionRequiredMixin, View):
+    permission_required = "api.add_superuser"
+    http_method_names = ["post"]
+
+    def post(self, request, *args, **kwargs):
+        user = get_object_or_404(User, pk=kwargs["pk"])
+        cluster_user = ClusterUser(user)
+
+        try:
+            cluster_user._init_user()
```
michaeljcollinsuk, line 131: I'm wondering if it might be better to actually [call the `create` method on the cluster object](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/api/cluster.py#L285).

As that way, if the users role had been removed it would be created, which currently there is no way for us to do via Control Panel. And digging into the code that calls role creation, if it already exists the [exception is caught](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/api/aws.py#L468), so would go on to call the `_init_user()` method.

controlpanel/frontend/jinja2/user-detail.html
```diff
@@ -153,23 +153,21 @@ <h2 class="govuk-heading-m govuk-error-summary-heading">
 </section>
 {% endif %}
 
-{% if request.user.has_perm('api.reset_mfa') %}
+{% if request.user.has_perm('api.add_superuser') %}
```
michaeljcollinsuk, line 156: Since you're already making changes in this area of the code, how about doing a couple of extra minor changes:
1. Display the `date_joined` field on the user detail page - would have been handy to have been able to see that here when we were looking into this support ticket
2. When setting someone as a superuser, mark them as `is_staff` too. Although looking at the view, this one isnt _quite_ as straightforward as I thought it would be, so could leave it as it is for now.

jamesstottmoj: Both have been added. Adding as staff at the same time isn't done in the best way but it's the only way I could get it to work.

michaeljcollinsuk (approved):
LGTM

PR #1370, opened 2024-10-29
Add missing order_by to ensure duplicate values removed from filter

michaeljcollinsuk:
Resolves minor bug found when testing version 2.17.0 in dev environment.

The use of `order_by` ensures duplicate values are removed from the chart filter options


Gary-H9 (approved):
LGTM

PR #1371, opened 2024-11-01
Feature/show app admins and deep link fix

jamesstottmoj:
## :memo: Summary
This PR resolves ministryofjustice/analytical-platform#5942

This PR moves the app log link over to opensearch. It also lists the admins of apps in the superuser all apps view.

The changes in this PR are needed because elasticsearch has been deprecated. Also, the admin list in control panel is a nice to have.

- Check side effects, if any

- Probably need to test in dev region


controlpanel/frontend/jinja2/includes/app-list.html
```diff
@@ -36,6 +41,19 @@
       <td class="govuk-table__cell">
         {{ yes_no(user.is_app_admin(app.id)) }}
       </td>
+      {% if request.user.has_perm('api.add_superuser') -%}
+      <td class="govuk-table__cell">
+        {%- for admin in app.admins %}
```
michaeljcollinsuk, line 46: This is potentially quite an expensive N+1 query to run on the app-list page. So I would recommend looking at using [prefetch_related](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#prefetch-related) to avoid this. To do this, you'll probably need to do some extra work in the view rather than use the `admins` helper property directly.

Also using Django Debug Toolbar (which is installed in the project) is really useful to help debug this sort of thing

michaeljcollinsuk (changes requested):

controlpanel/frontend/jinja2/includes/app-list.html
jamesstottmoj: Surely we're sending over that information already so the queryset already has that in? All I'm taking is what's already in what's being sent over to the template at that point.

michaeljcollinsuk: Querysets don't include all related objects by default, unless you use a `prefetch_related` (or `select_related` for FK's). So the query to get the relatad objects only happen when you try and access them, which is what the new call to `app.admins` will do. But actually someone [already added a `prefetch_related` in to the view](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/controlpanel/frontend/views/app.py#L56) so _might_ be ok. Although I'm doubtful due to the `admins` property using a `filter(is_admin=True)` might actually scupper it.

jamesstottmoj: When I inspect the queryset returned in the view before the template is rendered, the admins object is already there and set. All I've done is take that information and add it to the template. I'd be very surprised if that query is made while the template is being rendered.

michaeljcollinsuk: > When I inspect the queryset returned in the view before the template is rendered, the admins object is already there and set
How do you mean? How are you inspecting them?

michaeljcollinsuk (approved):
LGTM

PR #1378, opened 2024-11-05
Exclude any users that have no auth0-id as we cannot link to that user

jamesstottmoj:
Bug fix for displaying app admins and basic webapp page having an app admin header


michaeljcollinsuk (commented):
LGTM

michaeljcollinsuk (approved):

PR #1380, opened 2024-11-05
Wrap ip allowlist update in transaction lock

michaeljcollinsuk:
Fixes this bug https://github.com/ministryofjustice/analytical-platform/issues/5972

Something of a temporary solution, as it may be better to change logic so that updating app IP allowlists happen as a celery task, so that if it initially fails the call to update is retried. But this can be revisited as part of future work.


julialawrence (approved):
LGTM

PR #1381, opened 2024-11-06
Added button to export admin list as CSV

jamesstottmoj:
## :memo: Summary
This PR is linked to ministryofjustice/analytical-platform#3378

This PR adds a button on the admin webapps page to export a list of apps and their admins as a CSV file

The changes in this PR are needed because this has been requested before

- Check side effects, if any

- Pull down branch
- Create some dummy webapps
- Go to all webapps page
- click button at bottom of page
- check CSV contents as are expected

### DIFF
--- a/controlpanel/frontend/jinja2/webapp-admin-list.html
+++ b/controlpanel/frontend/jinja2/webapp-admin-list.html
@@ -12,12 +12,17 @@
   {% if request.user.has_perm('api.list_app') %}
   {{ app_list(apps, request.user) }}
 
-    {% if request.user.has_perm('api.create_app') %}
-    <p class="govuk-body">
-      <a class="govuk-button" href="{{ url('create-app') }}">
-        Register an app
-      </a>
-    </p>
+    {% if request.user.has_perm('api.is_superuser') %}
+    <h2 class="govuk-heading-l">Export app admins</h2>
+    <p class="govuk-body">Export a list of app admins to a CSV file.</p>
+
+    <form action="{{ url("app-admin-csv") }}" method="post">
+      {{ csrf_input }}
+      <button class="govuk-button js-confirm"
+              data-confirm-message="Are you sure you want to export a list of app admins?">
+              Export Admins
+      </button>
+    </form>
     {% endif %}
   {% endif %}
 {% endblock %}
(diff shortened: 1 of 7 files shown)


controlpanel/frontend/views/app.py
```diff
@@ -74,6 +76,33 @@ def get_queryset(self):
...
 
+class AppAdminCSV(OIDCLoginRequiredMixin, PermissionRequiredMixin, View):
+    permission_required = "api.is_superuser"
+
+    def post(self, request, *args, **kwargs):
+        apps = (
+            App.objects.filter(userapps__is_admin=True)
+            .annotate(
+                username=F("userapps__user__username"),
+                email=Coalesce("userapps__user__justice_email", "userapps__user__email"),
```
michaeljcollinsuk, line 87: I was wondering if it would be better to include both emails, but actually the justice email is probably the one you want to use in practice, as the other could be a non-work email as its based on whats linked to your GitHub. So actually think this is fine

controlpanel/frontend/views/app.py
```diff
@@ -74,6 +76,33 @@ def get_queryset(self):
...
+            )
+            .values("name", "username", "email")
+            .order_by("name")
+        )
+
+        response = HttpResponse(
+            content_type="text/csv",
+            headers={"Content-Disposition": 'attachment; filename="app_admins.csv"'},
```
michaeljcollinsuk, line 95: Might be worth adding a timestamp to the filename?

controlpanel/frontend/jinja2/webapp-admin-list.html
```diff
@@ -12,12 +12,17 @@ <h1 class="govuk-heading-xl">{{ page_title }}</h1>
   {% if request.user.has_perm('api.list_app') %}
   {{ app_list(apps, request.user) }}
 
-    {% if request.user.has_perm('api.create_app') %}
-    <p class="govuk-body">
-      <a class="govuk-button" href="{{ url('create-app') }}">
-        Register an app
-      </a>
-    </p>
+    {% if request.user.has_perm('api.is_superuser') %}
```
michaeljcollinsuk, line 15: I wonder if we should have this at the top of the page, as it could be easy to miss by having it at the bottom as it is such a long list in prod?

michaeljcollinsuk: Another alternative would be to do this in the Django Admin, implemented as an [admin action](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/actions/). The benefit here would be that you can then use some of the built in django admin functionality in order to only export a subset of app admins, rather than all apps.

michaeljcollinsuk (approved):
Nice work!

PR #1388, opened 2024-11-12
Fix parameter list so key and role and displayed in the correct order

michaeljcollinsuk:
Resolves minor bug described in https://github.com/ministryofjustice/analytical-platform/issues/6097


jamesstottmoj (approved):
LGTM

PR #1391, opened 2024-11-18
Changed lifecycle config to use intelligent tiering.

jamesstottmoj:
## :memo: Summary
This PR is associate with ministryofjustice/analytical-platform#6119

This PR changes the bucket lifecycle policy to use intelligent tiering
The changes in this PR are needed because it has been recommended to us and it will help with the new max S3 bucket limit
This also adds a button in the superuser data sources page that will update all existing buckets.

- Check side effects, if any

### DIFF
--- a/controlpanel/frontend/jinja2/datasource-list.html
+++ b/controlpanel/frontend/jinja2/datasource-list.html
@@ -19,6 +19,18 @@
 {% block content %}
 <h1 class="govuk-heading-xl">{{ page_title }}</h1>
 
+{% if all_datasources and request.user.has_perm('api.is_superuser') %}
+<h2 class="govuk-heading-m">Update all bucket lifecycle configurations</h2>
+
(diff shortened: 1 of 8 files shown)


controlpanel/frontend/jinja2/datasource-list.html
```diff
@@ -19,6 +19,18 @@
 {% block content %}
 <h1 class="govuk-heading-xl">{{ page_title }}</h1>
 
+{% if all_datasources and request.user.has_perm('api.is_superuser') %}
```
michaeljcollinsuk, line 22: Is this intended as a one-time action? And then will we remove it afterwards? If thats the case thats fine, in retrospect could have added a [data migration](https://docs.djangoproject.com/en/5.1/topics/migrations/#data-migrations) to loop through the buckets and do the update, to save needing to implement a temporary view etc

michaeljcollinsuk (approved):
LGTM

PR #1392, opened 2024-11-18
🛂 Add statement for cross account assumption

jacobwoffenden:
This pull request:

- Contributes to https://github.com/ministryofjustice/analytical-platform/issues/5743
- Adds a new policy statement template to allow an inputted app role to assume

Signed-off-by: Jacob Woffenden <jacob.woffenden@justice.gov.uk>

### DIFF
--- a/controlpanel/api/templates/assume_roles/cloud_platform_xacct.json
+++ b/controlpanel/api/templates/assume_roles/cloud_platform_xacct.json
@@ -1,0 +1,8 @@
+{
+    "Sid": "AllowCloudPlatformCrossAccountIAM",
+    "Effect": "Allow",
+    "Action": "sts:AssumeRole",
+    "Principal": {
+        "AWS": "{{ app_role }}"
+    }
+}


controlpanel/frontend/views/app.py
```diff
@@ -200,12 +200,28 @@ def get_success_url(self):
 
     def form_valid(self, form):
         try:
+            assume_role = form.cleaned_data.get("allow_cloud_platform_assume_role")
+            role_arn = form.cleaned_data.get("cloud_platform_role_arn")
+            if assume_role and not role_arn:
+                form.add_error("cloud_platform_role_arn", "Role ARN is required")
+                return FormMixin.form_invalid(self, form)
```
michaeljcollinsuk, line 207: Django forms have a `clean` method where you can add validation on fields that depend on each other - so you might want to have a look at moving this there to keep the logic out of the views. Link to docs https://docs.djangoproject.com/en/5.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other

settings.yaml
```diff
@@ -11,6 +11,11 @@ enabled_features:
     _HOST_dev: true
     _HOST_test: true
     _HOST_prod: true
+  cloud_platform_assume_role:
+    _DEFAULT: false
+    _HOST_dev: false
```
michaeljcollinsuk, line 16:
```suggestion
    _HOST_dev: true
```
Should we enable it in dev for testing?

michaeljcollinsuk (commented):
Tested locally, added some suggestions that you might want look into

michaeljcollinsuk (approved):
LGTM

PR #1393, opened 2024-11-22
Potentially fixed issue when updating all buckets.

jamesstottmoj:
Ran into issue in dev trying to update deleted buckets.

The code below should fix this and also log out any other interesting errors.


michaeljcollinsuk (changes requested):

michaeljcollinsuk (approved):
LGTM

PR #1396, opened 2024-11-26
Feature/quicksight embed permission

michaeljcollinsuk:
## :memo: Summary
This PR contributes to https://github.com/ministryofjustice/analytical-platform/issues/6118

It adds an additional checkbox to the manage user page "Quicksight access" section. When checked and submitted, the user will be granted a permission that will allow them to access the Quicksight embed view, and view the link to the page in the navigation bar.

I have used the built in Django permissions to add a new permission for viewing the link to the embedded Quicksight and accessing view. This removes the requirement for a user to be a superuser in order to access it.

Users can still be granted to the "legacy" Quicksight running in the data-prod account as before. As part of the changes, I refactored the view that access was granted to use a form, as this allowed better validation than doing this directly in the view. Additional tests have also been added to check users permissions.

The changes in this PR are needed because currently the only way to grant users access to the embedded quicksight is to make then a superuser - this is not sustainable

- N/A

- Code to add new permissions
- Login in the QuicksightAccessForm

- This is somewhat difficult to test locally, as you need to be a superuser to update a users access, and likely only have one user that you can log in with. 
- A workaround is to:
    - Log in as your superuser
    - Go to the django admin > Users > select your user > Under user permissions, grant the "Can embed quicksight" permission
    - Unset your user as a superuser/staff user
    - Reload to view the control panel as a normal user - you should see the embedded Quicksight link and can access the page
- This will be easier to test in dev where we have more users to test with, and can temporarily remove admin access and grant the embed permission via the "manage user" page in the app
   

- [x] Documentation will be added in the future - documentation on additional steps to take add users to Quicksight are documented in a [google doc](https://docs.google.com/document/d/1sYOAyLVHyiJyVTOHe8AwhPUHCbmkeYChYeBT29mN4fs/edit?tab=t.0#heading=h.y2ecoacdxsjf). We made need to link to this from the Control Panel when this is added to general user guidance.


jamesstottmoj (approved):
Tested locally.

LGTM

PR #1401, opened 2024-12-02
Feature/feedback form

jamesstottmoj:
## :memo: Summary
This PR is linked to ministryofjustice/analytical-platform#6112

This PR adds a form that allows users to give feedback on the analytical platform

- Check side effects, if any

- Pull down branch
- Run sever
- Click feedback link on banner above nav bar
- Add feedback


controlpanel/api/models/feedback.py
```diff
@@ -0,0 +1,25 @@
...
+        (1, "Very dissatisfied"),
+    ]
+
+    satisfaction_rating = models.IntegerField(
+        choices=SATISFACTION_RATINGS,
+        null=False,
+        blank=False,
+    )
+
+    suggestions = models.TextField()
```
michaeljcollinsuk, line 21: I wonder if we want to make this optional? Not sure either way, so maybe check with the team - but if so should just be a case of adding `blank=True`

controlpanel/frontend/views/feedback.py
```diff
@@ -0,0 +1,26 @@
...
+    form_class = FeedbackForm
+    model = Feedback
+    template_name = "feedback-create.html"
+
+    def get_success_url(self):
+        return reverse_lazy("feedback-thanks")
+
+    def form_valid(self, form):
+        form.save()
+        return FormMixin.form_valid(self, form)
```
michaeljcollinsuk, line 22:
```suggestion
```
Dont really need this as you will get this already from the `CreateView` ([this site](https://ccbv.co.uk/projects/Django/5.0/django.views.generic.edit/CreateView/#form_valid) is great for checking CBV's)

michaeljcollinsuk (changes requested):
Looks great! Some very minor comments

Also I dont think it is worth adding a custom page to check the feedback, but a quick win would be to register it with the django admin - that way there is an option to check what is coming in in the brower. See [docs here](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/) but can be as simple as adding the below to the `api/admin.py` file:
```
@admin.register(Feedback)
class FeedbackAdmin(admin.ModelAdmin):
    list_display = ["satisfaction_rating", "suggestions", "date_added"]
```

michaeljcollinsuk (approved):
Nice work!

PR #1404, opened 2024-12-04
Feature/deprecate tool

michaeljcollinsuk:
## :memo: Summary
This PR relates to https://github.com/ministryofjustice/analytical-platform/issues/6111

It allows AP admins to:
- Mark a tool as deprecated
- Optionally add a deprecation message to display to users. Otherwise a default message is used
- When a user selects a deprecated tool, they can still deploy it, but the deprecation message is displayed
- Retire a tool, to make it inaccessible to users. A message is displayed to users telling them they must deploy a different version

Furthermore, it adds `image_tag` as a field to the `Tool` model and makes it a required field when creating a new tool release. This helps simplify the code elsewhere when looking up what version of a tool a user has deployed.

There are a number of other minor related changes including:
- Tighten up the logic that disables the "Open", "Deploy" and "Redeploy" buttons based on the tool the user has selected
- Add a status to tools, and display this in the admin tool release page
- Update tool release filtering to filter by these statuses
- Remove the code which would get details about a release from the helm chart directly. This caused the tool page to be very slow to load, and by making the `image_tag` and `description` fields on a Tool release required fields, is no longer necessary.

- Once merged, it will be difficult to rollback these changes. 
Screenshot 2024-12-13 at 16 17 24
Screenshot 2024-12-13 at 16 17 10
Screenshot 2024-12-13 at 16 16 58
Screenshot 2024-12-13 at 16 15 37

- Any feedback on the code

- Pull down the branch
- Migrate the database
- Try creating some releases, and marking them as deprecated, to the behaviour when a deprecated tool is selected
- Try deploying a deprecated release
- Then make the release retired, go back to the "Your tools" page and check what is displayed
- Switch between tool options to test the scenarios for when the open/deploy/restart buttons are disabled.
- The "Open" and "Restart" buttons should only be enabled when there is an installed tool deployed and it has been selected in the tool dropdown list
- The "Deploy" button should only be enabled when selecting a tool that is not currently installed


jamesstottmoj:
Tested locally. Works well. Needs updating to latest main but otherwise LGTM

jamesstottmoj (approved):
LGTM

jamesstottmoj (approved):

PR #1423, opened 2024-12-27
Convert ToolDeployment to a django model

michaeljcollinsuk:
## :memo: Summary
This PR resolves https://github.com/ministryofjustice/analytical-platform/issues/6266

This PR refactors how deployed tools are tracked and deployed. A database record will be stored for each time a user deploys a new tool. This will make it much easier for us to see how many users are using each version of a tool in the Control Panel.

It also allows us to refactor the frontend for listing and deploying tools, simplifying what details need to used on the tool list page and what to POST back when deploying a tool.

- It will be difficult to rollback these changes because of the extend of changes made. Particularly because much of the information required to deploy a helm chart will now be read from the database.

- Changes in the views
- Changes to convert ToolDeployment to a django model
- Changes to how tool deployments are triggered

- Run the migrations with `python manage.py migrate`
- Rebuild the JS with `make build-js`
- Alternatively rebuild your dev container, which should do run both the above as part of setup
- Run the project locally, ensuring you are running the background worker. Celery is not required but doesnt hurt to have it running either
- You will need some deployable Tool releases that you can use to test with, ideally multiple releases for each tool to test switching between versions. I consider adding some fixtures to create these, but it would expose some secrets that are stored in the `values` field. So instead, I would suggest copying tool releases from prod/dev. If taking from prod, if you want to try accessing the tool you will need to update some of the helm values for the release, such as the auth0 domain and client ID/secret. Let me know if you need a hand with this.
- Go to the Your Tools page, select a tool, and deploy. This should work as before e.g. status updates to deploying, then deployed. She deployed tool should show as "Installed" in the drop down menu when it has successfully deployed
- Switch between different release versions and check everything is working as expected.
- Use the "restart tool" button to restart a tool. Again, this should work as before
- When you have some tools deployed, go to the Tool Release admin page - there is now a column that indicates the number of users that are using each tool. Check the information displayed is correct.


jamesstottmoj (commented):
Manually tested and seems to work ok.

One issue I ran into was the caching of javascript code which meant no buttons would activate on the tools page so that may be something to potentially be aware of (hard refresh gets round this).

Potential idea for making use of this data is to automatically uninstall tool releases that are marked as retired. This will force users to use the control panel to install the latest release of that tool.

controlpanel/api/models/tool.py
```diff
@@ -131,57 +125,78 @@ def status_colour(self):
...
+    is_active = models.BooleanField(default=False)
+
     Error = cluster.ToolDeploymentError
-    MultipleObjectsReturned = django.core.exceptions.MultipleObjectsReturned
 
-    objects = ToolDeploymentManager()
+    objects = ToolDeploymentQuerySet.as_manager()
+
+    class Meta:
+        ordering = ["-created"]
```
jamesstottmoj, line 170: Do we want to follow the table naming scheme for this model?

`db_table = "control_panel_api_tool_deployment"`

jamesstottmoj (dismissed):
LGTM

jamesstottmoj (approved):

PR #1429, opened 2025-01-09
Update create user to expect users from Entra

michaeljcollinsuk:
This PR resolves relates to https://github.com/ministryofjustice/analytical-platform/issues/6237

To enable CICA users to access the CP, we will enable a new connection in Auth0 with EntraID.
This updates the code we use to create a user from the ID token to normalize their name, and store their email as the justice_email, to avoid them having to reauth to capture it.

Tooling and all AP functionality other than QuickSight have been disabled for users that are not created via GitHub.
In future, to enable tooling for users coming via Entra, I think we would need have to update the helm charts so that we do not rely on their username, as those created from Entra are not compatible.

Another side effect is that we have a unique restriction on the `justice_email` field on the user model. This means if you auth via entra and your justice email is already stored against your normal "alpha" user, upon login a 500 error is raised as a user will already exist with that justice email address. To get around this in testing we will need to remove the `justice_email` from your alpha user before authing with entra.

- In Auth0 requires the connection with entra (`justicedigital-panda-awsidentitycenter`) to be set up and enabled for the Control Panel EKS app in Auth0. This is currently set up in dev.
- If you have a `justice_email` set against your current alpha user created by logging in with GitHub, remove it
- Log out and go to the login screen
- Type in your justice email in field below the "or" on the login page
- You should then be logged in and created as a new user and redirected to the "help" page
- A superuser will then need to enable the user as a QuickSight user in order to allow them to access QS

- [x] Documentation will be added in the future because ... (unclear what and how we want to document these changes as they are so specific to CICA currently)

### DIFF
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -249,7 +249,7 @@ class User(EntityResource):
 
     @property
     def iam_role_name(self):
-        return f"{settings.ENV}_user_{self.user.username.lower()}"
+        return f"{settings.ENV}_user_{self.user.slug.lower()}"
 
     def _run_helm_install_command(self, helm_chart_item):
         values = []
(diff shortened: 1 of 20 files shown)


jamesstottmoj (commented):
Tested locally with my justice user.

Changed my existing user's email so user could log into control panel.

Could see limited toolbar options. Switched to superuser and granted new user access to quicksight. 

Switched back to new user and could access quicksight successfully.

jamesstottmoj (approved):

PR #1433, opened 2025-01-15
Fix/quicksight-access-bug

jamesstottmoj:
This PR fixes a bug where clicking the save changes button would remove quicksight access despite none of the tick boxes changing state. Amended tests to capture this fix.


Gary-H9 (approved):
LGTM

PR #1434, opened 2025-01-16
fix/exception

jamesstottmoj:
Fixed exception getting caught too soon. Added logging to user creation and adding to group.


tom-webber (approved):
lgtm

PR #1435, opened 2025-01-16
fix/logging-levels

jamesstottmoj:
Reduced the level of some logging statements as they don't need to be error.


jhpyke (approved):
Lgtm to move to info level

PR #1438, opened 2025-01-21
Ensured policy is only attached/removed for github users

jamesstottmoj:
This PR fixes attempting to attach policies to users who shouldn't have AWS roles attached to them.


michaeljcollinsuk (approved):
LGTM.

PR #1439, opened 2025-01-21
Lower email before checking justice domain

michaeljcollinsuk:
Lowercase the email address before checking if the domain is from an expected Justice domain. Continue to store the email in original case.
Move the code to check the email to a function, to allow easier unit testing.
Update all tests associated with checking the justice domain, to add in cases where email received is in uppercase.


jamesstottmoj (approved):
LGTM

PR #1440, opened 2025-01-23
Remove the broadcast message about NGINX image

michaeljcollinsuk:
Enough time has passed that it seems reasonable to remove this now. Also means we will no longer
have to refer to "CDE NGINX" in new releases, as these will now be the default version of the auth proxy.

image


jhpyke (approved):
A kinder world can exist in our time

PR #1443, opened 2025-01-27
Feature/cloud platform take 2

jamesstottmoj:
## :memo: Summary
This PR is linked to ministryofjustice/analytical-platform#6547

This PR allows existing apps to link a cloud platform role in order to access resources managed in cloud platform. It also bumps some dependencies.

The changes in this PR are needed because the current implementation only works for new apps

- Check side effects, if any

- Create an app through the control panel without applying a cloud platform role
- Go to the app detail page
- Add an AWS arn via the form
- Check that both the trust policy contains both serviceaccount and cloud platform templates
- Check that inline policy has been added to allow the app role to assume the cloud platform role
- Remove ARN and check that inline policy has been removed and  cloud platform section has been removed from trust policy


controlpanel/frontend/views/app.py
```diff
@@ -362,6 +373,46 @@ class EnableTextractApp(PolicyAccessMixin, UpdateView):
...
+        try:
+            with transaction.atomic():
+                app = self.get_object()
+                app.cloud_platform_role_arn = form.cleaned_data.get("cloud_platform_role_arn")
+                app.save()
+                app.update_cloud_platform_access()
+                messages.success(self.request, "Successfully updated cloud platform ARN")
+                return HttpResponseRedirect(self.get_success_url())
+
+        except Exception as e:
```
michaeljcollinsuk, line 402: I know that catching `Exception` is not uncommon already in the codebase, but its not great practice. If you want to catch an exception to handle errors on the frontend more gracefully than just letting the 500 page raise, I would suggest two options:

1. Catch `botocore.exceptions.ClientError` here, log it to sentry, then proceed to redirect the user and display your 'friendly' error message.
2. Catch `botocore.exceptions.ClientError` in `aws.py` (specifically `MalformedPolicyDocumentException` perhaps?) and then raise a custom exception that we catch here so we can log it in sentry, then redirect the user with your 'friendly' error message

controlpanel/frontend/forms.py
```diff
@@ -235,6 +235,30 @@ def clean_namespace(self):
...
+class CloudPlatformArnForm(forms.Form):
+    allow_cloud_platform_assume_role = forms.BooleanField(initial=False, required=False)
+    cloud_platform_role_arn = forms.CharField(
+        required=False, max_length=130, validators=[validators.validate_aws_role_arn]
+    )
+
+    def __init__(self, *args, **kwargs):
+
+        super().__init__(*args, **kwargs)
+
```
michaeljcollinsuk, line 247:
```suggestion
    def __init__(self, *args, **kwargs):

        super().__init__(*args, **kwargs)

```
No need to define this

michaeljcollinsuk (changes requested):
Tested locally following your instructions and worked as expected. Just some feedback on the code to review, mainly around exception handling.

PR #1445, opened 2025-01-29
Attempt to fix bug polling failed subprocess

michaeljcollinsuk:
There is a bug (or maybe it was intended) where the background workers all pick up the same deployment messages. The first one to receive it will get on with deploying, wheres as the others all get an error because the deployment is already in progress. We have previously updated the code to check this error and ignore it if it is a "good" error, but this failed subprocess is then used to check the status of the deploy - which reports a failure.

This PR makes a further change so that when the deploy process fails, we return None. This should then result in the status being checked directly from the deployment, rather than the failed subprocess.

I've also increased logging to try to help with debugging if issues persist. 

I think this should be a temporary measure, I have an idea for a change that should be more robust, by stoping multiple workers trying to deploy the same tool.

### DIFF
--- a/controlpanel/api/models/tool.py
+++ b/controlpanel/api/models/tool.py
@@ -204,11 +204,13 @@ class ToolDeployment(TimeStampedModel):
         """
         if self._subprocess:
             # poll subprocess and maybe parse any output to get status
-            log.info("Polling helm subprocess")
+            log.info(f"Polling status of helm subprocess: {id(self._subprocess)}")
             status = self._poll()
             if status:
                 log.info(status)
                 return status
+
+        log.info("No subprocess to poll, checking deployment status")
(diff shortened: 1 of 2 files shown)


controlpanel/api/models/tool.py
```diff
@@ -204,11 +204,13 @@ def get_status(self, id_token=None, deployment=None):
         """
         if self._subprocess:
             # poll subprocess and maybe parse any output to get status
-            log.info("Polling helm subprocess")
+            log.info(f"Polling status of helm subprocess: {id(self._subprocess)}")
             status = self._poll()
```
jamesstottmoj, line 208: I would also log out the return code in the `_poll` method. I think that's where the tool status is set to failed based on what I could see in the logs.

jamesstottmoj (approved):

PR #1452, opened 2025-02-04
Backport a solution to bug finding correct Tool

michaeljcollinsuk:
After further testing found that users could have
a ToolDeployment created with incorrect tool, as
we were not excluding the tools with high mem/cpu. This fixes that bug.


jamesstottmoj (approved):
LGTM

PR #1461, opened 2025-02-12
Try unpinning botocore

michaeljcollinsuk:
There was a bug in a version of `botocore` that resulted in tests failing. To get around this, we pinned `botocore` to a working version. However this has now been resolved, and since pinning both `boto3` and `botocore` causes issues with dependabot ([like this](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/1458)) this PR unpins `botocore`, and updates `boto3` so that we get latest compatible versions of both.


jamesstottmoj (approved):
Tested locally and works fine

PR #1462, opened 2025-02-12
Feature/uninstall retired releases

michaeljcollinsuk:
## :memo: Summary
This PR relates to https://github.com/ministryofjustice/analytical-platform/issues/6705

The main aim of the PR is to reduce the number of users using old releases. When a tool is marked as retired, a celery task will be triggered to uninstall all active deployments of that tool. This is done using a helm uninstall command. In order to avoid removing releases in the middle of users using them, the task is set to run at 3am the next day (in prod only - in dev it is immediate, to make it easier for us to test).

As part of the changes, I have refactored some of the code related to running helm commands. This is because I encountered some bugs discovered that we were implicitly waiting for helm to run commands in some scenarios - therefore I have made it explicit that when we upgrade/uninstall a helm release, we wait for the return code.

The changes in this PR are needed because we currently have a long list of old and deprecated tools, we want to reduce this right down to just the fewest versions as possible. But currently even after a tool is retired, users can still keep using it if it is deployed by hitting the tool URL directly. By uninstalling the release at the point of retiring the tool, this will no longer be possible, and will mean users have to go to the control panel and install a new release. Further reasons are outlined in this [ticket](https://github.com/ministryofjustice/analytical-platform/issues/6599), which includes the comms that we will send out to our users.

- The tasks to uninstall tool releases
- Celery implementation - this differs to the existing celery tasks. It is less custom and using celery as it is intended.
- Changes to helm to check if any new bugs could be introduced

- Pull the code
- Run the server, celery and background worker together
- Create a tool release if there is not one in your local DB already
- Deploy the tool and wait for it to succeed - use kubectl to confirm that the tool was installed in your namespace
- Go to the manage tool page, and mark the tool as retired
- Check celery worker log output
- Use kubectl to confirm that the tool was uninstalled from your namespace. The URL of your tool should not be accessible either
- Go to the "Your tools" page - a message should be displayed to say that your previous release was uninstalled

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,8 +21,7 @@
     "vscode": {
       "extensions": [
         "EditorConfig.EditorConfig",
-        "GitHub.vscode-github-actions",
-        "GitHub.vscode-codeql"
+        "GitHub.vscode-github-actions"
       ]
     }
   }
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -1052,7 +1052,8 @@ class ToolDeployment:
         try:
             return helm.delete(self.k8s_namespace, self.release_name)
         except helm.HelmError as error:
-            # TODO make this less generic
+            # TODO make this less generic - catch when release not found, so nothing to uninstall
+            # possibly catch when in progress also
             raise ToolDeploymentError(error)
 
     def restart(self, id_token):
--- a/controlpanel/api/cluster.py
+++ b/controlpanel/api/cluster.py
@@ -1051,10 +1051,12 @@ class ToolDeployment:
     def uninstall(self):
         try:
             return helm.delete(self.k8s_namespace, self.release_name)
+        except helm.HelmReleaseNotFound as error:
+            raise error
         except helm.HelmError as error:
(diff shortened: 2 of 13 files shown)


jamesstottmoj (commented):
Tested locally and works. One thing that I did notice that despite the release being retired and removed, it was still marked as active in the database and reflected on the tools page when filtered for retired releases. Incredibly minor and may not be necessary to fix.

michaeljcollinsuk:
> Tested locally and works. One thing that I did notice that despite the release being retired and removed, it was still marked as active in the database and reflected on the tools page when filtered for retired releases. Incredibly minor and may not be necessary to fix.

I did consider this, and there is a commit somewhere where it marked the deployment as "inactive" as part of the retirement process. But if we mark it as inactive, it meant users who dont upgrade themselves, wont see the ["your previous tool was retired, please upgrade" message](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/b9d35206121e331a657c8278feee5e3cd1388e86/controlpanel/frontend/jinja2/tool-list.html#L90). So ultimately decided it would be preferable to leave as active for this case.

jamesstottmoj (approved):
LGTM

PR #1464, opened 2025-02-14
Hide bedrock self serve for non IAM users.

jamesstottmoj:
This PR hides the bedrock option for non IAM users.


michaeljcollinsuk (approved):
LGTM

PR #1470, opened 2025-02-20
Feature/more-self-serve

jamesstottmoj:
## :memo: Summary
This PR links to ministryofjustice/analytical-platform#6855

This PR allows users to unassign themselves from data warehouses. Also added a redirect for the user page that can be added to the user guide.

The changes in this PR are needed to give users more control over actions they can take in control panel as well as reduce the number of support requests we receive.

- Check side effects, if any

- For data warehouse changes:
  - Go to the data warehouse page.
  - Create a data warehouse if you don't have one already
  - Hit the revoke access button
- To test redirect:
  - Go to `/user/` endpoint in control panel
  - Check it has redirected you to your user page 

- [ ] Documentation will be added in the future


controlpanel/frontend/views/datasource.py
```diff
@@ -302,9 +315,26 @@ def get_object(self, queryset=None):
 
     def get_success_url(self):
         messages.success(self.request, "Successfully revoked access")
+
         return reverse_lazy("manage-datasource", kwargs={"pk": self.object.s3bucket.id})
 
 
+class RevokeAccessSelf(RevokeAccess):
+    permission_required = "api.destroy_users3bucket_self"
+
```
michaeljcollinsuk, line 324: Personally I would look up the access object differently:
```
def get_object(self, queryset=None):
    """
    Try to get an UserS3Bucket object for the current user and the given S3 Bucket PK.
    """
    try:
        obj = self.request.user.users3buckets.get(s3bucket_id=self.kwargs["pk"])
    except self.model.DoesNotExist:
        raise Http404() 
    return obj
```

That then removes the need to loop through all S3 buckets [above](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/1470/files#diff-fbc3252a0ecd5c71584e32f6e93931c2121db52e7b971e39cd8c59990f844854R89) to assign the revoke URL, as your revoke URL is much simpler to build as it doesnt require any knowledge of the `UserS3Bucket` object:
```
reverse("revoke-datasource-access-self", kwargs={"pk": datasource.pk})
```
You could call that in the `datasource-list.html` template directly, or preferably I would add it to the `S3Bucket` model and then call that in the template. This way it makes it easier to document and write a unit test, as well as cuts down logic in the template:
```
# on S3Bucket class...

def get_absolute_revoke_self_url(self):
    """
    Build url to the view which revokes the current user's access to the datasource.
    """
    return reverse("revoke-datasource-access-self", kwargs={"pk": self.pk})
```
Another benefit of this approach is that you can also revoke access from the superuser "All datasources" list view - currently I tried doing this but got a 404 even if it was a bucket I had access to.

michaeljcollinsuk (changes requested):
I've tested this locally, the user redirect works great, and revoking access also worked - although not from the superuser "All datasource page", that always returned a 404 even when clicking revoke on a bucket I have access to.

I've proposed some changes to resolve this as well as simplify some of the logic to build the revoke url.

controlpanel/frontend/views/datasource.py
jamesstottmoj: Something to look at on Monday. I wasn't actually aware the buttons had also appeared in the superuser pages. Probably a shared template. To be honest, it makes little sense those buttons being there given that you can manage yourself elsewhere. It may be better having having separate templates between those views.

controlpanel/frontend/views/datasource.py
```diff
@@ -71,7 +72,12 @@ class BucketList(
     template_name = "datasource-list.html"
 
     def get_queryset(self):
-        return S3Bucket.objects.prefetch_related("users3buckets").filter(
+
+        userbuckets = UserS3Bucket.objects.filter(user=self.request.user)
+
+        return S3Bucket.objects.prefetch_related(
```
michaeljcollinsuk, line 78: Do you still need these changes?

michaeljcollinsuk (approved):
LGTM

michaeljcollinsuk (approved):
LGTM

PR #1473, opened 2025-02-27
Pinned numpy as above 2.0 clashes with elasticsearch. Added pipdeptree

jamesstottmoj:
Fixed issue caused by pandas pulling in a version of numpy incompatible with our elasticsearch package.


michaeljcollinsuk (approved):
LGTM

PR #1491, opened 2025-03-12
Use helm timeout rather than subprocess timeout

michaeljcollinsuk:
Updates helm commands to use the `--timeout` flag when necessary, rather than relying on the subprocess timeout.

This way, if the helm command times out, it will surface as a failure, in turn completing the subprocess. Hopefully this means that the helm release should then be marked as failed, rather than left in a pending state as we were seeing in some failed deploys such as https://github.com/ministryofjustice/data-platform-support/issues/1179#issuecomment-2717617421


jamesstottmoj (approved):
Tested locally.

LGTM

PR #1493, opened 2025-03-21
Bugfix/qs exception handling

michaeljcollinsuk:
PR to resolve [bug ticket](https://github.com/ministryofjustice/analytical-platform/issues/7320)

Adds a JusticeDomain model that we can add/update/remove via the admin, to avoid having to redeploy the site if we need to allow different justice email domains to access QS.


jamesstottmoj (approved):
LGTM

PR #1499, opened 2025-03-24
dashboard registration

jamesstottmoj:
## :memo: Summary
This PR partially satisfies ministryofjustice/analytical-platform#7127

This PR adds functionality to the control panel to register existing quicksight dashboards that will be accessible via the dashboard service.

The changes in this PR are needed as this allows users to grant access to and share quicksight dashboards securely.

- Check side effects, if any

- Add new env vars (these are in `control_panel.env.devcontainer` in 1pass)
- Create a dashboard in quicksight
- Go to the dashboards tab in the quicksight embed page
- right click on your dashboard and select copy link
- Go to dashboards page and click register a dashboard
- Add a name and paste the URL into the appropriate box
- Attempt to add admins, customers and domains to the dashboard

- [ ] Documentation will be added in the future because need to fully test to write docs


controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+        return reverse_lazy("manage-dashboard", kwargs={"pk": self.object.pk})
+
+    def form_valid(self, form):
+        # add dashboard, set creator as an admin and viewer
+        with transaction.atomic():
+            user = self.request.user
+            email = user.justice_email.lower()
+            dashboard = form.save()
+            dashboard.created_by = user
```
michaeljcollinsuk, line 73: You've already saved when`created_by` is assigned, so it wont be stored in the DB. To avoid saving twice, you could change to:
```
dashboard = form.save(commit=False)
dashboard.created_by = user
dashboard.save()
```

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+    http_method_names = ["post"]
+    model = Dashboard
+
+    def get_redirect_url(self, *args, **kwargs):
+        return reverse_lazy("manage-dashboard", kwargs={"pk": kwargs["pk"]})
+
+    def post(self, request, *args, **kwargs):
+        self.perform_update(**kwargs)
+        return super().post(request, *args, **kwargs)
+
```
michaeljcollinsuk, line 138: To make it clear that subclasses need to implement `perform_update` I would add:
```
def perform_update(self):
    raise NotImplementedError("Subclasses must define this method")
```

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+    def form_valid(self, form):
+        dashboard = self.get_object()
+        dashboard.delete()
+        messages.success(self.request, f"Successfully deleted {dashboard.name} dashboard")
+        return HttpResponseRedirect(self.get_success_url())
+
+
+@method_decorator(feature_flag_required("register_dashboard"), name="dispatch")
+class UpdateDashboard(
```
michaeljcollinsuk, line 123: This looks like a base view, rather than a view to be used itself, so I'd name it accordingly `UpdateDashboardBaseView`

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+        return reverse_lazy("manage-dashboard", kwargs={"pk": kwargs["pk"]})
+
+    def post(self, request, *args, **kwargs):
+        self.perform_update(**kwargs)
+        return super().post(request, *args, **kwargs)
+
+
+@method_decorator(feature_flag_required("register_dashboard"), name="dispatch")
+class AddDashboardAdmin(UpdateDashboard):
+    model = Dashboard
```
michaeljcollinsuk, line 142:
```suggestion
```
It's on the base view so remove it here? and on the other subclasses

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def form_valid(self, form):
+        self.get_object().add_customers(form.cleaned_data["customer_email"])
+        messages.success(self.request, "Successfully added customers")
+        return HttpResponseRedirect(self.get_success_url())
+
+    def get_form_kwargs(self):
+        kwargs = FormMixin.get_form_kwargs(self)
+        return kwargs
+
```
michaeljcollinsuk, line 208:
```suggestion
```
As per comment above, if you inherit from the correct subclasses there is no need to override this method. 

The reason this was needed before (presume it was copied from the app views?) was that the `UpdateView` implementation of `get_form_kwargs` passes an `instance` kwarg to your form, but your form does not expect this since it is not a ModelForm, so would raise an error.

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+    model = Dashboard
+    permission_required = "api.retrieve_dashboard"
+    template_name = "dashboard-customers-list.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        dashboard = self.get_object()
+
+        customers = dashboard.viewers.all()
+        paginator = Paginator(customers, 25)
```
michaeljcollinsuk, line 178: We could increase this since we read the viewers from the DB - I think the default of 25 for apps came from limitations with the auth0 api. Maybe we bump to 50?

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+    OIDCLoginRequiredMixin,
+    PermissionRequiredMixin,
+    UpdateView,
+):
+    form_class = GrantDomainAccessForm
+    model = Dashboard
+    permission_required = "api.add_dashboard_domain"
+
+    def get_form_kwargs(self):
+        kwargs = FormMixin.get_form_kwargs(self)
```
michaeljcollinsuk, line 274: After changing the classes, you can then use super here `kwargs = super().get_form_kwargs()`

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+        context["page_no"] = page_no = self.kwargs.get("page_no")
+        context["paginator"] = paginator
+        context["elided"] = paginator.get_elided_page_range(page_no)
+        context["customers"] = customers
+        context["remove_customer_form"] = RemoveCustomerByEmailForm()
+        return context
+
+
+@method_decorator(feature_flag_required("register_dashboard"), name="dispatch")
+class AddDashboardCustomers(OIDCLoginRequiredMixin, PermissionRequiredMixin, UpdateView):
```
michaeljcollinsuk, line 190: `UpdateView` isnt the right class to use here, as it your form is not a modelform. So instead go for:
```suggestion
class AddDashboardCustomers(OIDCLoginRequiredMixin, PermissionRequiredMixin, SingleObjectMixin, FormView):
```
This means you still get [`get_object`](https://ccbv.co.uk/projects/Django/5.0/django.views.generic.detail/SingleObjectMixin/#get_object) from the `SingleObjectMixin` class.

And you still get all your form methods from the [`FormView` class](https://ccbv.co.uk/projects/Django/5.0/django.views.generic.edit/FormView/). Also means you can remove the hacky `get_form_kwargs` redefinition.

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+            )
+
+        messages.success(self.request, f"Successfully removed customer {email}")
+
+
+@method_decorator(feature_flag_required("register_dashboard"), name="dispatch")
+class GrantDomainAccess(
+    OIDCLoginRequiredMixin,
+    PermissionRequiredMixin,
+    UpdateView,
```
michaeljcollinsuk, line 267: Same as above - change to use `SingleObjectMixin, FormView` instead of `UpdateView`

controlpanel/frontend/views/dashboard.py
```diff
@@ -0,0 +1,303 @@
...
+    def get_success_url(self):
+        return reverse_lazy("manage-dashboard", kwargs={"pk": self.kwargs["pk"]})
+
+    def form_valid(self, form):
+        domain = form.cleaned_data["datasource"]
+        dashboard = self.get_object()
+        dashboard.whitelist_domains.add(domain)
+        messages.success(self.request, f"Successfully granted access to {domain.name}")
+
+        return FormMixin.form_valid(self, form)
```
michaeljcollinsuk, line 288:
```suggestion
        return super().form_valid(form)
```
Can use super here also, and then the `FormMixin` import can be removed

michaeljcollinsuk (changes requested):
Looks like a lot of comments but its all minor stuff, mostly to try do things a bit more the way django intends. Happy to talk through anything if you want. But otherwise good work, ran it locally and works!

controlpanel/api/models/dashboard.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def add_customers(self, emails):
+        for email in emails:
+            viewer, _ = DashboardViewer.objects.get_or_create(email=email)
+            self.viewers.add(viewer)
+
+    def delete_customers_by_id(self, customer_ids):
+        try:
+            viewers = DashboardViewer.objects.filter(pk__in=customer_ids).all()
+            self.viewers.remove(*viewers)
```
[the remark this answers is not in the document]
jamesstottmoj: It depends how we want this to work. The app model is that the users would still exist in auth0 but wouldn't be linked to a particular app. Keeping dashboard viewers isn't necessarily a bad thing nor is them keeping the role if they've been added. However, if you feel that it's best that we prune them then I'm not against that.

controlpanel/frontend/jinja2/dashboard-customers-list.html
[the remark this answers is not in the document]
jamesstottmoj: The customer list is selectable. So you can click multiple users and then click a button that will remove multiple users. I actually think the delete by email is redundant other than not searching through the paginated users.

controlpanel/api/models/dashboard.py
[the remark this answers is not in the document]
jamesstottmoj: A task has since been added to resolve this

controlpanel/api/tasks/dashboards.py
```diff
@@ -0,0 +1,29 @@
...
+
+# First-party/Local
+from controlpanel.api.models.dashboard_domain import DashboardDomain
+from controlpanel.api.models.dashboard_viewer import DashboardViewer
+
+log = structlog.getLogger(__name__)
+
+
+@shared_task(acks_on_failure_or_timeout=False)
+def prune_dashboard_viewers():
```
michaeljcollinsuk, line 13: I think I understand why this is needed now - it covers the edge case where:

1. User is added to dashboard A, creating a DashboardViewer object
2. DashboardDomain whitelist is added to dashboard B, the same email domain as the user added in step 1
3. User is removed as a viewer for dashboard A
4. This task runs - the DashboardViewer is now orphaned, so is included in your queryset lookup on line 25. 
5. Because the user's email domain is related to another dashboard, it gets caught on line 27 - and so the DashboardViewer object isnt deleted. This ensures they keep the dashboard viewer role, and can continue to log in to the dashboard service.

Is that the right edge case this is covering? If so then I think this will work well! Does mean we still end up with some objects in the DB that look orphaned, but dont think it should cause too many problems.

Has made me realise I may need to give a bit more thought to how we lookup the whitelisted domains in the pre-registration action in Auth0. But can discuss that when we get to it

controlpanel/api/auth0.py
```diff
@@ -629,6 +639,14 @@ def has_existed(self, user_id):
 
         return len(response["users"]) > 0
 
+    def remove_role(self, user_id, role_id):
+        response = self.remove_roles(user_id, [role_id])
+
+        if "error" in response:
+            raise Auth0Error("reset_mfa", response)
```
michaeljcollinsuk, line 646: Should this be something specific to removing the role? Looking at the [signature to `APIException` in the srouce code](https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L108) it expects a message as the first arg, and a code as the second

michaeljcollinsuk, line 646: Should this be something specific to removing the role? Looking at the [signature to `APIException` in the source code](https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L108) it expects a message as the first arg, and a code as the second

controlpanel/api/tasks/dashboards.py
jamesstottmoj: Yes. It covers that exact edge case.

controlpanel/api/auth0.py
jamesstottmoj: Should have changed the `reset_mfa` string but I'm raising that error in the same way that all other methods are raising it.

jamesstottmoj: Changed the way this function handles exceptions.

michaeljcollinsuk (approved):
LGTM! Great work

PR #1504, opened 2025-04-08
Added API endpoints to get dashboard list for a given user and an embed URL

jamesstottmoj:
## :memo: Summary
This PR is linked to ministryofjustice/analytical-platform#7530

This PR adds new API endpoints for the dashboard service to retrieve a list of dashboards for a user and an embed url for a given dashboard.

The changes in this PR are needed to provide functionality for the dashboard service

- Add dashboard(s) to the control panel
- Grant a mix of customers and domain access to your dashboards
- hit endpoints and check results

- [ ] Documentation will be added in the future when entire epic is finished and tested

### DIFF
--- a/controlpanel/api/serializers.py
+++ b/controlpanel/api/serializers.py
@@ -564,7 +564,7 @@ class DashboardSerializer(serializers.ModelSerializer):
 
 class DashboardUrlSerializer(DashboardSerializer):
     class Meta(DashboardSerializer.Meta):
-        fields = ("name",)
+        fields = ("name", "admins")
 
     def to_representation(self, dashboard):
         data = super().to_representation(dashboard)
--- a/controlpanel/api/serializers.py
+++ b/controlpanel/api/serializers.py
@@ -546,10 +546,29 @@ class AppAuthSettingsSerializer(serializers.BaseSerializer):
         auth_settings = app_auth_data["auth_settings"]
         auth_settings_status = app_auth_data["auth0_clients_status"]
         return self._process_auth_settings(auth_settings, auth_settings_status)
+    
+
+class DashboardAdminSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ("name", "email") 
 
 
 class DashboardSerializer(serializers.ModelSerializer):
+    admins = DashboardAdminSerializer(many=True, read_only=True)
 
     class Meta:
         model = Dashboard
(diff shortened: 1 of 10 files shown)


controlpanel/api/serializers.py
```diff
@@ -545,3 +546,29 @@ def to_representation(self, app_auth_data):
...
+    admins = DashboardAdminSerializer(many=True, read_only=True)
+
+    class Meta:
+        model = Dashboard
+        fields = ("name", "quicksight_id", "admins")
+
+
+class DashboardUrlSerializer(DashboardSerializer):
+    class Meta(DashboardSerializer.Meta):
+        fields = ("name",)
```
michaeljcollinsuk, line 567: Do you not want to include the admins here? It might be something we display on the detail page

michaeljcollinsuk (changes requested):

controlpanel/api/serializers.py
jamesstottmoj: The detail page is going to be the embedded dashboard I thought so I wasn't sure if we wanted to display that info on the page. I can add it back if you think it's valuable

controlpanel/api/views/dashboards.py
```diff
@@ -0,0 +1,70 @@
...
+from controlpanel.api.serializers import DashboardSerializer, DashboardUrlSerializer
+from controlpanel.utils import get_domain_from_email
+
+
+class DashboardViewSet(ViewSet):
+    """
+    A ViewSet for managing dashboards.
+    """
+
+    permission_classes = [permissions.IsSuperuser]
```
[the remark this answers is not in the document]
jamesstottmoj: One thing I have done is separate the dashboard domain out into it's own env var; `DASHBOARD_SERVICE_DOMAINS` as I felt it would be better to separate them out. We can keep it in with `QUICKSIGHT_DOMAINS` if you're fine with that though.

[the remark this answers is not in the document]
michaeljcollinsuk: Hadnt noticed that - makes sense to separate it, so leave it that way

controlpanel/api/serializers.py
michaeljcollinsuk: TBH not sure if we will display it or not, but dont think it hurts to include it, and then if we do need to use it we have the option without having to push a change to control panel

michaeljcollinsuk (approved):
Great work!

PR #1509, opened 2025-04-25
Fixed issue related to settings not being set in login-fail page

jamesstottmoj:
Fixes sentry issue caused by settings not being set in a non authenticated page. Force nav to hide to resolve problem


michaeljcollinsuk (approved):
LGTM

PR #1512, opened 2025-05-02
Dashboards pages should be accessible by non iam users.

jamesstottmoj:
## :memo: Summary

This PR allows non IAM users to register dashboards
The changes in this PR are needed because otherwise these users won't be able to share via the dashboard service


michaeljcollinsuk (approved):
LGTM

PR #1516, opened 2025-05-07
Fixed case sensitivity issues when adding customers to dashboards

jamesstottmoj:
This PR fixes some case sensitivity issues when adding customers. It also adds user as a dashboard viewer if added as a dashboard admin in control panel

### DIFF
--- a/controlpanel/api/models/dashboard.py
+++ b/controlpanel/api/models/dashboard.py
@@ -40,7 +40,7 @@ class Dashboard(TimeStampedModel):
 
     def add_customers(self, emails):
         for email in emails:
-            viewer, _ = DashboardViewer.objects.get_or_create(email=email)
+            viewer, _ = DashboardViewer.objects.get_or_create(email=email.lower())
             self.viewers.add(viewer)
 
     def delete_customers_by_id(self, customer_ids):
--- a/controlpanel/frontend/forms.py
+++ b/controlpanel/frontend/forms.py
@@ -520,6 +520,9 @@ class CustomersField(forms.Field):
 class AddCustomersForm(forms.Form):
     customer_email = CustomersField()
 
+    def clean_customer_email(self):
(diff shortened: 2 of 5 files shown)


michaeljcollinsuk:
Changes look good, but when reviewing I noticed that the [test_add_customers](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/main/tests/frontend/views/test_dashboard.py#L238) test doesnt check that DashboardViewers are created. Changing to the below should fix this, and make sure that we store emails lowercase as per these changes:
```
@pytest.mark.parametrize(
    "emails, expected_response, count",
    [
        ("foo@example.com", add_customer_success, 1),
        ("FOO@example.com", add_customer_success, 1),
        ("foo@example.com, bar@example.com", add_customer_success, 2),
        ("FOO@EXAMPLE.COM, Bar@example.com", add_customer_success, 2),
        ("foobar", add_customer_form_error, 0),
        ("foo@example.com, foobar", add_customer_form_error, 0),
        ("", add_customer_form_error, 0),
    ],
    ids=[
        "single-valid-email",
        "single-valid-email-uppercase",
        "multiple-delimited-emails",
        "multiple-delimited-emails-uppercase",
        "invalid-email",
        "mixed-valid-invalid-emails",
        "no-emails",
    ],
)
def test_add_customers(client, dashboard, dashboard_viewer, users, emails, expected_response, count):
    client.force_login(users["superuser"])
    data = {"customer_email": emails}
    response = client.post(
        reverse("add-dashboard-customers", kwargs={"pk": dashboard.id}),
        data,
    )
    assert expected_response(client, response)
    emails = [email.strip().lower() for email in emails.split(",")]
    assert dashboard.viewers.filter(email__in=emails).count() == count
```
Can you add this (assuming you are happy with it)? I dont want to do it myself otherwise I think I dont think I'll be able to approve the PR.

michaeljcollinsuk (approved):
LGTM

PR #1559, opened 2025-07-14
Fix the admin helptext on Dashboard list page

michaeljcollinsuk:
Fix the dashboard admin modal helptext so that it refers to Dashboards, not apps:

image


Gary-H9 (approved):
LGTM

PR #1574, opened 2025-07-28
Remove button to open dev Airflow link

michaeljcollinsuk:
We have shut down Airflow-Dev as part of our migration to new Airflow. This removes now defunct link.

image


jhpyke (approved):
Sayonara, `Airflow-Dev` 🫡  o7

PR #1582, opened 2025-08-11
login-fail-message

jamesstottmoj:
## :memo: Summary
This PR is linked to ministryofjustice/analytical-platform#8350

This PR adds a message to inform users that are logging in using email to use their github identity instead.

The changes in this PR are needed because we have had several instances of people using the email form instead of the github button since changing over to the universal login page.

### DIFF
--- a/controlpanel/frontend/jinja2/login-fail.html
+++ b/controlpanel/frontend/jinja2/login-fail.html
@@ -14,6 +14,8 @@
             </a>
         </p>
 
+        <p>If you are seeing this page after multiple login attempts, you may be using your email to log into the control panel instead of using your GitHub account. Please try logging in using the Continue with GitHub button instead of filling out the email address field.</p>
+
         <p>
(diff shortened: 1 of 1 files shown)


controlpanel/frontend/jinja2/login-fail.html
```diff
@@ -14,6 +14,8 @@ <h2 class="govuk-heading-l">Your session has expired</h2>
             </a>
         </p>
 
+        <p>If you are seeing this page after multiple login attempts, you may be using your email to log into the control panel instead of using your GitHub account. Please try logging in using the Continue with GitHub button instead of filling out the email address field.</p>
```
jacobwoffenden, line 17:
```suggestion
        <p>If you are seeing this page after multiple login attempts, you may be using your email to log into the Control Panel instead of using your GitHub account. Please try logging in using the Continue with GitHub button instead of filling out the email address field.</p>
```

PR #1774, opened 2026-04-10
Switch to use uv for dependency management

michaeljcollinsuk:
I was prompted to make these changes in response to the recent supply chain attacks and after reading this [blog post](https://blog.pypi.org/posts/2026-04-02-incident-report-litellm-telnyx-supply-chain-attack/?utm_campaign=Django%2BNewsletter&utm_medium=email&utm_source=Django_Newsletter_331#protecting-yourself-as-a-developer) about what python developers can do to protect their projects. By switching to uv will be locking all our dependencies, including transient ones. It is also much faster than using `pip` and brings the project in line with our modern projects.

I have removed the old requirements files, along with some other files no longer necessary that would have required updating in order to keep (e.g. docker-compose.dev.yaml, pytest.ini).

### Testing these changes

All build scripts have been updated, so the best way to test is to:
1. Rebuild the dev container
2. In VS Code point your python interpreter to the `.venv` python
3. Run the project as normal and try to login etc


jamesstottmoj (approved):
Tested locally and seems to work.

PR #1914, opened 2026-06-16
Update auth0 connections

jamesstottmoj:
## :memo: Summary
This PR updates the Auth0 SDK to v5. This requires a significant change to our Auth0 code. This is required as connection info will no longer be served via the current endpoints.

Currently, there is a bug in the API that means we cannot get users via their github ID. This means that we need to wait for a fix. In the meantime, there is a workaround to get the user credentials from auth0 but it's not nice. This will be removed once the fix has been issued.

I have not modified some classes that interact with the Auth0 Authorisation extension as they still seem to work in the same way.

Things to test that interact with Auth0

1. Webapp creation/deletion
2. being able to create an Auth0 client
3. Add/remove users from a webapp
4. Add/remove users from a dashboard


michaeljcollinsuk (approved):

michaeljcollinsuk:
Closing as changes were merged in https://github.com/ministryofjustice/analytics-platform-control-panel/pull/1973

